PR #343, opened 2013-09-04
Improve nested swipers support and noSwiping logic. Added new demo for nested horizontal swipers.

SpOOnman:
I've improved Isaac Strack's noSwiping logic and it's now possible to have nested horizontal (I guess vertical too) swipers. There are two main changes:
- when you touch somewhere, the most nested swiper grabs this event, it was ambigous previously,
- noSwipingClass is checked by swiper only on it's own slides, wrapper and container. Previously it checked all nested slides, wrappers and containters which was faulty.

I've attached a new demo that shows this clearly. Play with this, then comment out current 2.1 version JS file and use development JS to see improvements.


nolimits4web:

PR #351, opened 2013-09-12
Index is out of range at assembling loop

slax0rr:
When the option 'loopAdditionalSlides' is set, there is a JavaScript
error that the object (_this.slides) does has no 'innerHTML'. This
happens because the 'loopedSlides' is a sum from 'slidesPerView' and
'loopAdditionalSlides'. This error occurs as well if either are set to
higher than the number of actual slides.
Also changed the assembly a bit, so it is no longer mixed up.


nolimits4web:

PR #543, opened 2013-12-23
Fix/loader loop fix

FrivalszkyP:
I love love love the loader feature (which we discussed is still in beta). I changed the behaviour of the loop logic when the loader is set up with the option loadAllSlides set to false. Since when loadAllSlides are false, the looped slides are supposed to be loaded from the loader (when the active slide is 0, the looped slide before that should be the loader.slides.length-1, and the one after the last slide should be slide 0 and so on). I hope my code makes sense and helps in the development of the loader feature. :)


nolimits4web:

PR #656, opened 2014-02-25
Leave height totally alone when calculateHeight = true

bitflower:
When using the Swiper within responsive context I found it useful NOT to have the plugin touch the height of the wrapper at any time.

Does this interfere with any internal logic of the plugin?


nolimits4web:

PR #710, opened 2014-04-01
Adding RTL language support

jamiemagique:
Hi we are using this script on a site that supports both LTR and RTL languages.

The LTR or RTL gets set in the <html> dir="" attribute.

I spotted that the translate value was getting set as a negative value when it wants to be a positive value for RTL (note that .carousel-slide in the CSS is also set to float: right for RTL and float: left for LTR)

If the HTML dir attribute is set to RTL, I am altering the coords.x value. This value is only set to something other than 0 is the carousel mode is horizontal is far as I can tell.

### DIFF
--- a/src/idangerous.swiper.js
+++ b/src/idangerous.swiper.js
@@ -2544,6 +2544,11 @@ Swiper.prototype = {
             }
             coords[y] = x;
         }
+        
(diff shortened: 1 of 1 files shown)


nolimits4web:
Closing as new Swiper 3 released with different API and full RTL support

Issue #906, opened 2014-08-25
.swipeNext() not working in Chrome with 110% zoom


Flowingweb:
Hi,

I've found a bug when using Chrome and having the pagezoom set to 110%. Other zooms and browsers seems to work. Tested also in Firefox 31 and IE 11. I'm currently running Chrome 36, but this bug existed in earlier versions too.

I've also tested the demo's on your site and the bug can be replicated. So it seems and genuine bug and not an implementation error.

Hope you can fix this in a next release, because otherwise the script is awsome!

Thanks,
Florian

PR #918, opened 2014-09-08
fixed issue #906 where swipeNext breaks when using chrome with 110% page zoom

gnitnuj:


nolimits4web:

Issue #684, opened 2014-03-17
"overflow-y: scroll;" inside vertical swiper


univie:
I have a vertical swiper and I would like to scroll overflowing content (vertically) inside each slide (using overflow-y: scroll), so that swiping to the next slide would only begin after the activeSlide has been scrolled to the bottom. Using the mousewheel is not an option for me. Also I would prefer not to use a scrollContainer inside each slide.

Any simple way to achieve this on mobile?

PR #926, opened 2014-09-11
Added options, methods, events, and API tweaks

sakabako:
This PR adds a number of options and features to Swiper, along with some tweaks to the UX and API. I would have liked to have separated them out, but I don't have them separate, so they're here as one. My apologies.

### New Options:
- `longSwipeMs` - lets you configure the number of milliseconds Swiper uses to determine what a "long" swipe is versus a "short" one. Decrease if swipes are happening unintentionally, increase if it's too hard to swipe.
- `minSwipePx` - configures the number of pixels a user must swipe to cause a page change during a short swipe. Decrease to make swiping easier, increase to make it harder.
- `ignoreTouchClass` - if an element with this class is touched the event will be ignored, this allows elements inside a swiper to use native scrolling. Allows #684

### New Events
These are available as options, just like the other events.
- `onReachEnd( swiper )` - when the carousel has reached the last slide
- `onReachBeginning( swiper )` - when the carousel has reached the first slide
- `onScrollEnd( event )` - fired when a scrollContainer has finished scrolling.

```javascript
onScrollEndEvent = {
    isAtLeft: bool, // only for horizontal swipers
    isAtRight: bool, // only for horizontal swipers
    isAtTop: bool, // only for vertical swipers
    isAtBottom: bool, // only for vertical swipers
    direction: string 'up', 'down', 'left' or 'right'
}
```

### Features:
-  `swiper.lock()`, `swiper.lockNext()`, or `swiper.lockPrev()` - prevent swiping in one direction, the other, or both.
-  `swiper.unlock()`, `swiper.unlockNext()`, or `swiper.unlockPrev()` - the opposite of above.
- `swiper.setMode( 'horizontal or vertical' )`-  changes the orientation of an existing carousel.

### API Changes
- Adding or removing a slide before the current slide will not change the visible slide, instead the current slide will be incremented or decremented.
- If there is 100% resistance and you swipe past the end (or beginning) of a carousel, touch events will bubble up to a parent Swiper.

### UX Changes

- The method of determining swipe speed for scroll containers has been changed to compare the oldest move event up to 150 milliseconds before the latest touch event. This prevents anomalous lightning fast or very slow tosses, and ensures that swipe-stop-release doesn't have any momentum.
- Changes to the transitionTimingFunction so that it matches native Android scrolling very closely.

### Bug Fixes
- End swipe when the browser triggers a touchcancel event. This prevents swiper from getting stuck between slides if the browser decided the swipe should end. It is treated as if the user lifted his/her finger.
- Ensure className of elements is a string before seeing if it should be ignored, as this is an object on SVG elements.


nolimits4web:

PR #968, opened 2014-10-16
Allowed multiple swiper using same pagination element class

Acombes:
Allowed multiple swiper using same pagination element class. Each new one used to override the first one.


nolimits4web:

PR #1055, opened 2015-01-15
Fix Loop-Mode with fractional slidesPerView-setting

Ragnaroek:
If you define slidePerView 1.22 together with loop mode the library throws an undefined object exception.
This should fix the loop mode.


nolimits4web:
Merged, thanks

PR #1057, opened 2015-01-19
Add RTL language support

MotaghiFar:
If your document direction is RTL,
You most add "direction: ltr" to 'swiper-container' for fix working Swiper.


nolimits4web:
Thanks

PR #1064, opened 2015-01-22
Restrict list of exported files

dfcreative:
There's really no need to store all the project files as a package resources.


dfcreative:
@nolimits4web any thoughts?

nolimits4web:
Merged, thanks

PR #1074, opened 2015-01-28
Fix: Remove the Swiper is "undefined" bug

### LINKED ISSUE #1072: Swiper is undefined
I am getting a  ***runtime*** JavaScript error (following a `webpack` compile) like this:

```
Uncaught ReferenceError: Swiper is not defined
```

If I leave the console open while refreshing the browser the error does not occur again, until I close the browser console and refresh that way.

The source line of the error is: https://github.com/nolimits4web/Swiper/blob/master/dist/idangerous.swiper.js#L2893

If I change the code like this:

```js
if (typeof define === 'function' && define.amd) {
  define([], function () {
    console.log(typeof Swiper)
    'use strict';
    return Swiper;
  });
}
```

Then there is no error as well...

I'm not 100% clear on what is going on but I'm pretty sure that a patch to `Swiper` will be needed.

jasonkuhrt:
This will close #1072.

This seems to be sufficient to make a happy runtime resulting from basic `webpack` build.

I don't know why the line
```
_this.loopedSlides = Math.floor(params.slidesPerView) + params.loopAdditionalSlides
```
changed in the build. I didn't do that.


nolimits4web:
Merged, thanks

PR #1094, opened 2015-02-11
Fixing paths to dist files in bower

austinknight:
The current paths don't compile correctly because they point to the wrong place.


nolimits4web:
Right, thanks

PR #1095, opened 2015-02-12
Update bower.json

lxeran:
main shouldn't contain css files, it breaks main-bower-files task for gulp.


nolimits4web:
https://github.com/bower/bower.json-spec#main

PR #1109, opened 2015-02-20
Fix ambiguous $ variable

fabrizim:
Replace references to `$`with `domLib`, otherwise the jquery file will fail if jquery is in noConflict mode.


nolimits4web:
Sure, thanks

PR #1123, opened 2015-02-26
updated component.json since it wasn't pointing to any files

stonebranch:
Fixed issue I ran into issue while using duo. Since component.json pointed to wrong folders component and duo stopped working after the v.3.x.x update.


nolimits4web:
Sure, thanks

PR #1125, opened 2015-02-26
remove trailing comma which may break on IE8

nilni-ck:


nolimits4web:
Merged, but no much sense, as it doesn't compatible even with IE 9)

PR #1143, opened 2015-03-06
3D effects also fail on IE 11, not only IE 10

Rikk:


nolimits4web:
Cooool! Is it cross browser?)

nolimits4web:
Wrong place

PR #1146, opened 2015-03-06
Simplify Preloader CSS

Rikk:


nolimits4web:
I mean it here:) is it cross browser?

Rikk:
There's no mention of incomplete support of the `animation-timing-function` property (the one responsible for `steps(x,end)`).
https://developer.mozilla.org/en-US/docs/Web/CSS/animation-timing-function
http://caniuse.com/#feat=css-animation
https://docs.webplatform.org/wiki/css/properties/animation-timing-function
IE10+ is said to support it fully: https://msdn.microsoft.com/en-us/library/ie/hh772240.aspx

nolimits4web:
Sorry, commits should be made in dist folder, i will move it there

nolimits4web:
Ok, it is good in all browsers, thanks

PR #1156, opened 2015-03-12
fix(hash): moved setHash() call to onTransitionEnd

oori:
1. There's UI lag in the animation while hash changes.   this is a browser issue, and not Swiper. (tested with your demo 25-hash-navigation.html)   
This proposed fix overrides this, by setting the hash **after** transition has ended.

2. Logically, there is no need to set the new hash, until I've stopped swiping. For example: while finger-swiping across multiple slides.


nolimits4web:
Can't be merged automatically, so did this manually. Thanks!

PR #1167, opened 2015-03-18
Keyboard accessibility

### LINKED ISSUE #1130: Keyboard accessibility
Integrating basic keyboard controls would be nice for some accessibility. Swiper still works on desktop and it will be used on desktop, so this makes sense:

- Make navigation controls focusable (arrows and bullets) with Tab, Shift + Tab.
- Provide visual feedback on focusable elements, so user can perceive current position on the page.
- Make navigation controls interactive via keyboard: Enter/Space key.

I can setup PR for this.

roman01la:
This PR closes #1130 

- Focusable navigation arrows (not bullets, I thinks arrows are enough)
- Visual feedback when focused on navigation arrows. Currently uses default platform outline, but can be changed to specific arrow color/opacity level.
- Navigation arrows are interactive via keyboard (Enter key).

Also there's a basic ARIA applied for screen readers:
- Notify about prev/next slide
- Notify about first/last slide

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,5 +1,5 @@
 (function(){
-    'use strict';    
+    'use strict';
     var gulp = require('gulp'),
         gutil = require('gulp-util'),
         connect = require('gulp-connect'),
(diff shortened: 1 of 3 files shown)


nolimits4web:
That looks and "sounds" cool. Thanks! I also did a bit of refactoring to make it more swiper's part and more isolate. Also moved keydown event handler to a11y because with your option using `bind` they become not detachable and produce errors after swiper destroy, so i fixed it

roman01la:
Oh, nice! Thanks!

PR #1175, opened 2015-03-20
Invoke correct method to preload slide image

voxtex:
Bug I found when using `lazyLoadingInPrevNext`.


nolimits4web:
Right, thanks

PR #1176, opened 2015-03-22
Fixed issue with unable to listen to events

miguelb:


nolimits4web:
Wow, what a fail typo from me:) I can't merge it because changes should be comitted to src folder only. So i merged it and updated manually. Thanks

PR #1185, opened 2015-03-23
Fix path to js file in custom plugin demo

rrelmy:
The custom plugin uses a path to the build directory which is not available in the normal download.


nolimits4web:
Thanks

PR #1190, opened 2015-03-24
cleanup autoplay upon destroy, fixes #1189

### LINKED ISSUE #1189: loop is not properly cleared after calling destroy
calling `destroy` on a swiper instance should clear any running timeouts immediately.

mrtnbroder:


nolimits4web:
Almost ok, but it shoul be in init.js, inside of .destroy() method, not here as it not related to events, right?;)

mrtnbroder:
Ok, I will change that as soon as Im back from the gym :) I just skimmed through the code real fast and implemented that, cause I need it fixed right meow :)

mrtnbroder:
closing this one in favor of #1191

PR #1191, opened 2015-03-24
cleanup dynamic styles, classNames and autoPlay upon destroy

mrtnbroder:


nolimits4web:
Accepted, thanks, but cleaning up styles and classes should be optional, will add it

mrtnbroder:
awesome thanks, it's probably missing some cleanup still on buttons or something, you know that better than I do ;)

My pleasure!

nolimits4web:
Yep, and i'll probably merge cleanUpStyles and detachClassNames funcions to one, to save some code lines :) As i don't think that there could be such situation when they needed separately

nolimits4web:
Ok, merged two funcs, also added cleanup for buttons, and better check for pagination and scrollbar elements. And i made it optional, it will not cleanup styles by default as really not always required

mrtnbroder:
:+1: 

planning on updating the version soon? :)

nolimits4web:
Probably on weekend

PR #1199, opened 2015-03-27
get only pagination and prev/next buttons that are related to the curren...

ksere:


ksere:
...in order to use more than one swipers in one page.

btw, Great work with swiper v3 guys.

nolimits4web:
Sorry, can't accept it. With such logic it would be absolutely impossible to move buttons or pagination outside of container, so it just bring limitation. Here is how to use multiple swipers on one page https://github.com/nolimits4web/Swiper/blob/master/demos/24-multiple-swipers.html

But if you want to use them with same classes you can always do it like:
```
$('.swiper-container').each(function(){
    var s = new Swiper(this, {
        pagination: $(this).find('.swiper-pagination')[0],
        scrollbar: $(this).find('.swiper-scrollbar')[0],
        //...
    })
});

PR #1207, opened 2015-03-31
"s is null" when swiper is destroyed before transition end callback fires 

czajkowski:
Here is an example:
https://jsfiddle.net/mwvv9qag/6/

When 'Start' button is pressed, slideNext() method is triggered and then swiper is immediately destroyed.

A 'Transition End' event handler is registered and when it is resolved swiper is already destroyed. Any callback which references the swiper and (the 's' variable) and its methods will cause an error.


nolimits4web:
Could you leave only "Fix" commit that i can merge? Can't accept another ones

PR #1215, opened 2015-04-02
Added allowSlideToPrev and allowSlideToNext

jsundn:
Hi!
 
I noticed that allowSwipeToPrev -and Next only is used in onTouchMove.

My application is an e-reader for newspapers, and I have placed a empty page first so that page nr.1 is placed to the right. But I want to disable swipeToPrev in portrait-mode. because otherwise my users would be able to swipe to a empty page. And allowSwipePrev works great for this. But they can still navigate their with the keyboard arrow keys!

So my idea is so add allowSlideToPrev -and Next properties, and then check for these in the slideTo-function, and this works really good for me.

What do you think?

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -1485,6 +1485,14 @@ s.slideTo = function (slideIndex, speed, runCallbacks, internal) {
 
     var translate = - s.snapGrid[s.snapIndex];
 
+    // Directions locks
+    if (!s.params.allowSlideToNext && translate < s.translate) {
+        return false;
+    }
+    if (!s.params.allowSlideToPrev && translate > s.translate) {
+        return false;
+    }
+
     // Stop autoplay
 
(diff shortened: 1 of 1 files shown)


nolimits4web:
I don't think that we need these additional methods, and can use the same `allowSwipe...`, because i am not sure that there could be such use case when you need only swipes, so it is better to disable both slide and swipes. If we don't need swipes, we can always use `onlyExternal` options. 

So if you leave only this commit https://github.com/jsundn/Swiper/commit/5e1eadbc11f78c1421592eb0897649c452be53b0, but with `allowSwipe...` params, i will merge it

jsundn:
Ok, that sounds good for me =) Fixed in f45c9b7dcfe59c0ca1c6e79519a4280995a38e5c

nolimits4web:
Merged, thanks

PR #1219, opened 2015-04-03
Fixing null exception when destroying Swiper before transition callbacks...

czajkowski:
Here is an example:
https://jsfiddle.net/mwvv9qag/6/

When 'Start' button is pressed, slideNext() method is triggered and then swiper is immediately destroyed.

A 'Transition End' event handler is registered and when it is resolved swiper is already destroyed. Any callback which references the swiper and (the 's' variable) and its methods will cause an error.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -433,6 +433,7 @@ s.pauseAutoplay = function (speed) {
     }
     else {
         s.wrapper.transitionEnd(function () {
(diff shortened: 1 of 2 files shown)


czajkowski:
@nolimits4web 
 It was cleaner to create new pull request.

nolimits4web:
Thanks, merged

PR #1229, opened 2015-04-10
Return false from handleKeyboard if keycode is equal to not allowed direction

jsundn:
I'm swiping between pretty heavy canvas data, and thought about optimizing navigation with keyboard. So I lock swipes in "onTransitionStart " and then on "onTransitionEnd" I unlock swipes again. This way I  (and with my updated code in handleKeyboard) I can disable stacking of slideTo-calls.

Next slideTo if holding down arrow key will be called when current transition is done.


nolimits4web:
Ok, thanks

PR #1232, opened 2015-04-13
enableKeyboardControl() called more than once

felixmaier1989:
I ran into this issue:

Due to conflicts with Twitter Bootstrap's modal, I had to re-initialize keyboard navigation by calling enableKeyboardControl() again.

By pressing left arrow I would then jump to the next next slide! And so on... if you call enableKeyboardControl() again, you would browse slide 3 by 3...

Do you need a jsfiddle to reproduce this behavior?

### DIFF
--- a/src/js/keyboard.js
+++ b/src/js/keyboard.js
@@ -70,6 +70,6 @@ s.disableKeyboardControl = function () {
     $(document).off('keydown', handleKeyboard);
 };
 s.enableKeyboardControl = function () {
-	s.disableKeyboardControl();
-	$(document).on('keydown', handleKeyboard);
+    s.disableKeyboardControl();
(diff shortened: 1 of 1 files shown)


nolimits4web:
Thanks. But i would prefer to avoid such implementation with removing and adding events again in this method. And i don't understand why you need to call it few times, one time should be enough since it uses "live" event. Anyway, you can just call the same in your implementation by calling then:
```
s.disableKeyboardControl();
s.enableKeyboardControl();
```
in your app

felixmaier1989:
After investigation I found out our that our code was a bit dirty, calling enableKeyboardControl() several times in a row. 

Calling disableKeyboardControl() within enableKeyboardControl() would just be valuable in making the API more robust.

PR #1233, opened 2015-04-13
fixed issue with running mocha tests that include this library

ryanzec:
If I include this library with browserify and try to run my unit test, I get this error:

```
    Swiper.prototype = {
    ^
ReferenceError: Swiper is not defined
```

This is because most of the use cases of using the variable `Swiper` assume a browser.  I have slightly modified the code so that the initial creation uses `var Swiper = ...` instead of `window.Swiper = ...`.  At the end of the main file, I then do a `window.Swiper = Swiper;` to make sure it is available globally like it is now.  I also updated the amd script to use `window.Swiper` instead of just `Swiper` (which was just inferring the `window.` anyways).

I tested some of the examples (since there don't seem to be able unit tests) and everything worked fine.  If this could be merged and a new version pushed to npm that would be great (otherwise I will have to push my own version to npm to continue on with my project since I need my tests to pass).

Let me know if there is anything I need to change.

### DIFF
--- a/src/js/amd.js
+++ b/src/js/amd.js
@@ -3,11 +3,11 @@ Swiper AMD Export
 ===========================*/
 if (typeof(module) !== 'undefined')
 {
-    module.exports = Swiper;
+    module.exports = window.Swiper;
 }
 else if (typeof define === 'function' && define.amd) {
     define([], function () {
         'use strict';
-        return Swiper;
+        return window.Swiper;
     });
 }
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -354,7 +354,7 @@ s.loadImage = function (imgElement, src, checkForComplete, callback) {
     }
     if (!imgElement.complete || !checkForComplete) {
         if (src) {
-            image = new Image();
+            image = new window.Image();
             image.onload = onReady;
             image.onerror = onReady;
             image.src = src;
@@ -1241,7 +1241,7 @@ s.onTouchMove = function (e) {
         }
         velocities.push({
             position: s.touches[isH() ? 'currentX' : 'currentY'],
-            time: (new Date()).getTime()
+            time: (new window.Date()).getTime()
         });
     }
     // Update progress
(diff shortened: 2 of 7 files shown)


nolimits4web:
Doesn't it break use cases with Require js or in node.js or in components or browserify?

ryanzec:
I don't see how this would break any of them (I can guarantee it does not break browserify since I am currently using this exactly code with browserify).  All I am doing is instead of directly assigning `Swiper` to the `window` global variable at the top of the file I am assigning it is a scoped variable at the top and then assigning it to the `window` global variable at the bottom of the same scoped function.  Assuming the `window` variable is available (which the current code does) prevents this code from being able to be included outside the context of a browser (like within mocha unit tests).

Bear in mind that I know Swiper itself will not function properly outside the context of the browser however without this change, I can't even load the code for my mocha tests (without removing the inclusion of this npm module within my code which then defeat the purpose of being able to install this package as a npm module).

ryanzec:
Any update on this?

Again I know it works with Browserify/NodeJS and don't see any reason it would not work with other loading systems like RequireJS or Components.  I have never used RequireJS or Components and since there are no unit tests I can't confirm that this change does not break them.  I don't have the time to learn those system just to test this change.

If you don't want to merge these changes in because you don't know if they will break other loading systems and you don't have the time to test them, please close this PR and let me know that.  If that is going to be the case then I will just have to remove the library from my build process and manually include it (not ideal but it would be the only other option for me).

nolimits4web:
Yes, it is because i don't know will it affect other loading systems or not, and also don't have much time now to test it too. I will have a bit more time on next week, so let's keep it open for a while

ryanzec:
Just a little update as I was doing more testing, there where some other locations that was assuming `window` in the global context (like doing `new Date(...)` or `new Image(...)`) and added another commit that adds `window.` to the beginning of those statements.

nolimits4web:
Seems to be everything fine

PR #1247, opened 2015-04-21
Add Sticky option for FreeMode

bluepioupiou:
Allow swiper to stop right on the position of a slider if we swipe to a distant
position


nolimits4web:
Thanks, will check it

nolimits4web:
I merged it manually, as PR shouldn't touch dist/ files. Also made small improvements as we shouldn't rely on `slidesGrid`, we need to use `snapGrid`, as it could not much to `slidesGrid` because we may have different slide grid options, like `slidesPerGroup` and `slidesPerColumn`

Thanks

PR #1282, opened 2015-05-13
removed trailing comma

campbellwmorgan:
Trailing comma on last item in object breaks script for some browsers


nolimits4web:
Thanks, but i am wondering in which actual browser it could break anything?

campbellwmorgan:
It broke IE9 - which just about works with a polyfill! Thanks

nolimits4web:
No, it doesn't broke IE9, it broke  IE 8. Anyway, both of these IEs are not supported :D
Thanks anyway

PR #1301, opened 2015-05-30
Add style directive

dfcreative:
Make able to build swiper style via postcss-import.
https://github.com/postcss/postcss-import#usage


nolimits4web:
Thanks

PR #1314, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3


nolimits4web:
Sorry, but i still will keep it because it is used in Gulp source builder, also it is harmless and may be required in future

kkirsche:
lol Um, their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

nolimits4web:
Maybe, but i need it, and it is harmless and can't break anyhting

PR #1330, opened 2015-06-20
Fix slide on resize

pasyuk:
Prevent slide to previous view when isEnd and you resize window


nolimits4web:
Thanks, i merged it manually, as PR should be made only to the files in `dist/` folder without affecting the current stable release

PR #1342, opened 2015-06-30
Added support for umd

### LINKED ISSUE #1327: UMD
Currently, this library relies on a global jQuery object (`window.jQuery`). Some people using things like browserify do not have that variable, instead relying on `require('jquery')` being called.

Would you accept a PR adding support for [UMD](https://github.com/umdjs/umd)?

callumacrae:
UMD syntax taken from https://github.com/umdjs/umd/blob/master/returnExports.js

This will add support for CommonJS, basically. Currently we're having to use swiper.jquery.js, even though we have jQuery, because jQuery isn't globally defined in our app.

Closes #1327


callumacrae:
This doesn't actually work, gimme a bit…

callumacrae:
Ah, it looks like this won't actually work :(

https://github.com/nolimits4web/Swiper/blob/master/src/js/dom-plugins.js#L23

Because the dependencies are optional, and optional requires aren't possible, it looks like this approach isn't going to work. Unfortunately, your library relies on a global instance of jQuery!

Is there any way around this?

nolimits4web:
Can you give me your dev environment example, or few steps to replicate this issue? And will take a look

callumacrae:
Standard [Browserify](http://browserify.org/) or [jspm](http://jspm.io/) setup: `window.jQuery` is not defined, but `require('jquery')` (or `import $ from 'jquery';` in ES6) both return jQuery.

I'm not sure how it would work with optional dependencies, though, unless there were a separate file that required jQuery—currently swiper.jquery.js requires either Dom7, Zepto, or jQuery, and optional dependencies like that aren't possible with UMD.

callumacrae:
Okay, working now :smiley: 

There are no behavioural changes in swiper.js and swiper.jquery.js. swiper.umd.js is a version that works with a non-globally defined jQuery, but doesn't work with Dom7 or Zepto—a compromise required for UMD support, I guess.

.jshintrc
```diff
@@ -7,7 +7,6 @@
   "eqeqeq"       : true,
   "eqnull"       : true,
   "immed"        : true,
-  "latedef"      : true,
```
callumacrae: I can add this one back, but it will involve making _another_ file, and I don't really want to do that.

nolimits4web:
Ok, thanks. I will have more time on this weekend to investigate and test it well

callumacrae:
:+1:

callumacrae:
Thanks! I can set up an environment for you to test the UMD version in, if you want.

nolimits4web:
This would be great

callumacrae:
https://github.com/callumacrae/swiper-test

nolimits4web:
Ok, had a lot of tests and, indeed, there is no way to do that with optional dependancies, so probably we will need 2 additional files like `swiper.jquery.umd.js` and `swiper.zepto.umd.js` (and their minified versions :smile:) Will go with this approach?

callumacrae:
Sure, I can do that! I'll update the PR tomorrow.

nolimits4web:
Ok, so we just need 2 new files UMD files, one for jQuery and one for Zepto. Standalone version probably should be left intact as it seems to be works well with module includes. And small request, don't make commits to `dist/`files as you will overwrite already published release

callumacrae:
I've renamed the file to swiper.jquery.umd.js. I'm not sure how to go about the zepto one because it doesn't seem to be on npm; I think it would be best to wait until someone actually wants to use the zepto version, and then they can do it :)

callumacrae:
When will you be able to review this? The weekend, again?

nolimits4web:
Nope, tomorrow :)

nolimits4web:
Looks good, thanks

callumacrae:
@nolimits4web Hey! Any eta on a release for this? It's blocking a bugfix on our site. If you're not planning on doing a release for a while, we can use it off github, but we'd prefer not to.

Thanks :)

nolimits4web:
I will release today, a bit later

Отправлено с iPhone

nolimits4web:
Released

callumacrae:
Thank you!

PR #1346, opened 2015-07-02
Make the controllers respect the grids of each other

nerdmed:
Hey @nolimits4web,

first of thanks for maintaining this great library. I am using it in a current project for mobile.

I am using the controller feature to have some kind of 2 way selection on mobile.
The problem I found is that the controlling feature does not really respect the grid of the controlled swiper, in some special cases (centered slides, and different sizes in one slides). This is because the current implementation maps the controllers by a simple multiplicator thus by percentage.
This is a problem if you have some kind of title scroller where you scroll variable length text and wand both to be centered no matter which one you scroll.

The problem will get very clear if you change your current example 23 and replace the galleryThumbs images with text, and fixed width (like in this PR). This is how the scroller will behave if you do it with the current implementation: [Problem Image](http://i.imgur.com/vbKoht7.jpg)

I think the only way to solve this is by mapping the SnapGripds to each other and calculate the position for the controlled Swiper a little different. I have cerated a proof of concept  with an Linear Spline function i have optimized for performance in this PR. I have added some comments to the code. Would be great to get some feedback from you. I am willing to  clean this up and make it ready for a next version?

What do you think, any advice?


nolimits4web:
Thanks, looks interesting, will take a look closer at the weekend

nerdmed:
@nolimits4web great! I am using it right now in my project. Waiting for your advice how to proceed with it :+1:

nolimits4web:
Ok, i merged it manually, and make such method optional (enabled by default). But there is a small issue, when we have enabled loop, or use different slidesPerView without centeredSlides, then we have different amount of values in `snapGrid` array. As a workaround i used `slidesGrid` for this case, not the best option, but works better at least. Maybe is it possible to tweak somehow `LinearSpline` function?

PR #1364, opened 2015-07-16
Only get direct children for loopedslides

moorscode:
We have a situation where we want to have a slider inside a 'looped' slider.
This gives the problem that the events on the nested slider is getting the containing sliders events.
The slider should only look at the direct children to determine what to loop on.


nolimits4web:
Sorry, i can't merge it. First of all it breaks test scripts. This line doesn't have much sense:
```
var slides = s.wrapper.children('> .' + s.params.slideClass);
```
as:
* s.wrapper.`children` - the `children` method already says to look only in children
* `'> .' + s.params.slideClass` - selector starting from `>` is not a valid selector for `querySelector` and it will work only for jQuery

PR #1371, opened 2015-07-21
fix Swiper doesn't back to the first slide when `translate === 0`.

chinghanho:
I made a screen record for this issue I occured: https://youtu.be/6r_j_UyVc_w

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -1628,10 +1628,7 @@ s.slideTo = function (slideIndex, speed, runCallbacks, internal) {
     s.previousIndex = s.activeIndex || 0;
     s.activeIndex = slideIndex;
 
-    if (translate === s.translate) {
-        s.updateClasses();
-        return false;
-    }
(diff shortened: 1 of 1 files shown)


nolimits4web:
I don't really understand what happened there. Could you please give me the full code of this demo so i can test it?

sparkrevolutions:
I think I have the same issue? Certainly described the same way anyway!

I have a swiper container of 5 ember components set in a loop with centered true and slidesPerView set to 1.

When I attempt to swipe between the last item and the first item (completing one iteration of the loop), the next item (first item) does not render correctly.

To fix it, I have to manually set the translate3d value to -768px,0,0 or -3840px,0,0 on the swiper-wrapper using jquery or slightly move the slide again for it to render.

This means until the user has completed the loop, the page they are swiping in is not rendered correctly.

Any ideas?

nolimits4web:
Closing as incompatible with current version of Swiper

PR #1378, opened 2015-07-25
Update swiper.jquery.js

pzotov:
It fixes issue with building pagination when s.params.slidesPerview is 'auto' and slider should be looped.


nolimits4web:
I can't merge it as commits should be made ONLY in files in `src` folder. I will add it manually

PR #1380, opened 2015-07-28
+ Added support for iOS in-app UIWebView edge-swipe detection (for swipe-to-go-back).

live-cn:


nolimits4web:
Thanks, i will check it. But anyway i can't merge it as all commits should be made ONLY to `src` folder. You just modified the current release

nolimits4web:
Thanks, merged it manually

PR #1416, opened 2015-08-30
fix wrapper translate values round

Ashot-KR:
`roundLengths` property rounds only slides widths ignoring wrapper translate property. It may cause blurry effect. Fix that in tis commit.


nolimits4web:
Thanks, but i can't accept it as the commit should be made only to the files in `src` folder. If you can change it, i will accept it

Ashot-KR:
But all changes is in `src` folder. My previous pull request was wrong, so i create new one.
Just noticed, that there is too much changes in one file. Looks like my editor decided to reformat code(remove unnecessary tabs and spaces) without my agreement. I will create a new pull request tomorrow without such stupid changes.

PR #1419, opened 2015-09-01
fix roundLengths for wrapper translate

Ashot-KR:
Another try.
Fix `roundLengths` for wrapper translate.


nolimits4web:
Thanks, will check it in the weekend

nolimits4web:
Looks good, thanks

PR #1424, opened 2015-09-03
Remove trailing commas

robertberg:


nolimits4web:
Thanks, but does it actually improve anything? :)

PR #1452, opened 2015-10-02
Pagination markup logic change

vxc-:
Pagination wrapper markup has to be now a children of the main container if more than one pagination container found.

If more than one pagination container elements are found in the whole DOM, we force them to be a children of the main swiper container in order to avoid possible miss-behaviour of different Swiper instances with same pagination container classes. In this case, the flexibility of placing the pagination markup wherever you want across the DOM is lost, but gains behaviour consistency in the case of multiple swiper instances.

It's rather an unnecessary feature if you always set different pagination-container classes while instantiating different swipers, but feels quite more comfortable to just instantiate them without setting a specific class for each one's pagination container and respect a specific html markup hierarchy. If you dont, the auto-generated html bullets of the last instantiated swiper override all the previously generated ones, so does the click event listeners on the bullets, so, if you click on a bullet that is not a bullet of the last instantiated Swiper, triggers movement on itself and the next Swipers.

The change it's quite transparent to the library users since both @ documentation and @ the demos the pagination container html markup is always placed inside the container, people might not even know that you can place it outside.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -289,6 +289,9 @@ s.wrapper = s.container.children('.' + s.params.wrapperClass);
 // Pagination
 if (s.params.pagination) {
     s.paginationContainer = $(s.params.pagination);
+    if (s.paginationContainer.size() > 1){
+        s.paginationContainer = s.container.find(s.params.pagination);
+    }
     if (s.params.paginationClickable) {
         s.paginationContainer.addClass('swiper-pagination-clickable');
     }
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -289,7 +289,7 @@ s.wrapper = s.container.children('.' + s.params.wrapperClass);
 // Pagination
 if (s.params.pagination) {
     s.paginationContainer = $(s.params.pagination);
-    if (s.paginationContainer.size() > 1){
+    if (s.paginationContainer.length > 1){
         s.paginationContainer = s.container.find(s.params.pagination);
     }
     if (s.params.paginationClickable) {


nolimits4web:
Thanks. But the problem here is that i have few requests for Swiper to be able to handle multiple paginations, for example, on one top, and one and bottom, don't actually know why it is required, but anyway, how to be in this case? 

P.S. There is no `.size` method in Dom library

vxc-:
This case refers to different Swiper Instances on the same DOM, miss-behaving if the pagination container of those Swipers has the same class. It DOES NOT refer to multiple pagination controls of a single Swiper, that's another functionality.

I know `.size` is not present in the DOM library. It's a jQuery method (and since Swiper is a jquery-compatible lib I guess it's usage is more than justified), being applied to a jQuery result set in this case.

You can check the pagination missbehave in this jsFiddle: https://jsfiddle.net/d9rvhe8g/

Cheers!

nolimits4web:
It is not missbehave, it is because it is passed in a wrong way. The problem in your fix is that it totally will break such case https://jsfiddle.net/hbdxfqmy/

> I know .size is not present in the DOM library. It's a jQuery method (and since Swiper is a jquery-compatible lib I guess it's usage is more than justified), being applied to a jQuery result set in this case.

So why have you used it? With this, standalone version will become broken, because there is no such method there. You should use something they both have, otherwise you have to add such method to Dom7 as well. But you can use here `.length` instead of `.size()` which is supported by both

vxc-:
> The problem in your fix is that it totally will break such case https://jsfiddle.net/hbdxfqmy/

Yes, as I said:

> If more than one pagination container elements are found in the whole DOM, we force them to be a **children of the main swiper container in order to avoid possible miss-behaviour of different Swiper instances with same pagination container classes**. In this case, the flexibility of placing the pagination markup wherever you want across the DOM is lost, but gains behaviour consistency in the case of multiple swiper instances.

I guess is just a matter of priorities, whether you prefer Swiper to be consistent on a single instance behaviour with no markup placement restrictions or to become slightly more strict with the markup if more instances with same class names are present, but at least you get them to behave in the right way.

It's up to you Vladimir :P

nolimits4web:

PR #1473, opened 2015-10-22
add support for meteor

Chun-Yang:
TODO (@nolimits4web ):
1. create a meteor user account (you can use nolimits4web, if you want to use other name, please change 'name' attribute in package.js)
2. publish the package https://atmospherejs.com/i/publishing
3. when you update your package to a new version, please change the version in package.js and republish.

Thank you for your awesome package! It would be awesomer if you can support meteor users!

### DIFF
--- a/package.js
+++ b/package.js
@@ -1,0 +1,24 @@
+var version = '3.1.7';
+
+Package.describe({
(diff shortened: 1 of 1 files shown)


nolimits4web:
Great, thanks

PR #1480, opened 2015-10-27
fix bug with placeholder editing in IE

atolmachev:
Reproduced https://jsfiddle.net/gxo4nydd/4/ on IE 10,11 on Windows 7
Just click on input several times rapidly until you see cursor and placeholder at the same time. Now you can edit the placeholder. Input value stays empty. 

PS What use of blurring :inputs?
placeholder

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -1197,7 +1197,7 @@ s.onTouchStart = function (e) {
         var preventDefault = true;
         if ($(e.target).is(formElements)) preventDefault = false;
         if (document.activeElement && $(document.activeElement).is(formElements)) {
-            document.activeElement.blur();
+            if (!s.browser.ie) document.activeElement.blur();
         }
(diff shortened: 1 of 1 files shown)


nolimits4web:
I guess this is more IE issue rather than Swiper's. This `blur` is required for mobile/touch devices to actually blur inputs when you click outside of it. With your fix, can you still blur input after focusing it if you click on other place of slide?

atolmachev:
Ok, is the goal to avoid keeping focus on a non-active slide? 
Blurring input on touchend / touchcancel will be much safer then.

nolimits4web:
The goal is to blur input when you click outside of input. By default it won't blur because of `e.preventDefault` in `onTouchStart`, maybe touchend/cancel can be better, will check it

nolimits4web:

PR #1487, opened 2015-11-04
added new callback functions: onSlideNext and onSlidePrew

yapro:
Example usage:
```
    var mySwiper = new Swiper('.swiper-container', {
        keyboardControl: true,
        onSlideNext: function(s){
            console.log("right : " + mySwiper.activeIndex);
        },
        onSlidePrew: function(s){
            console.log("left : " + mySwiper.activeIndex);
        }
    });
```

### DIFF
--- a/dist/js/swiper.jquery.js
+++ b/dist/js/swiper.jquery.js
@@ -726,9 +726,6 @@
             if (s.rtl) offsetCenter = translate;
         
             // Visible Slides
-            var containerBox = s.container[0].getBoundingClientRect();
-            var sideBefore = isH() ? 'left' : 'top';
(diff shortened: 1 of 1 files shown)


nolimits4web:
Hi, thanks. But i can't accept this PR for few reasons:
- all PRs should be done to `src/` files only
- why have you removed this line `var clientLeft = s.container[0].clientLeft;` in slideNext/Prev methods? It should be there
- also, it is wrong place for these callbacks. They are attached to transition so they should be called in that place
- it is required to use `emit` method for callbacks, not calling them directly

Let me know if you are going to change it, otherwise, i can add it manually

Issue #1131, opened 2015-03-04
Angularjs porting?


tipodisimmia:

beepiams:
I have started on extending this from a former swiper/angular post on stack overflow: http://stackoverflow.com/questions/18323621/javascript-swiper-library-in-an-angular-directive

This version I'm using with Foundation For Apps (F4A):

It loads the swiper, but I have an issue that unless the window is reSized, the Swiper becomes 'stuck', and won't allow for initial interactivity. 

my swiperangular.js helper file:
```javascript
var mySwiper;
var myApp = angular.module("swiperAngular", []);
myApp.run(["$rootScope", "$window", function($rootScope, $window) {
  $rootScope.doReady = function () {
    console.log("doReady");
    mySwiper.reInit();
  }
}]);

myApp.directive("swiperDirective", ["$rootScope", function($rootScope) {
  return {
    restrict: "A",
    controller: function() {
      this.ready = function() {
        console.log("ready");
        $rootScope.doReady();
      }
    },
    link: function(scope, element, attrs) {
      mySwiper = new Swiper(".swiper-container", {
        loop:true,  
        autoplay: 500
      });
    }
  }
}]);

myApp.directive("swiperSlide", [function() {
  return {
    restrict: "A",
    require: "^swiperDirective",
    // template: "<div>hello {{data[$index].copy}}</div>",
    link: function(scope, element, attrs, ctrl) {
      console.log("a");
      if(scope.$last) {
            //console.log("i'm last");
            ctrl.ready();
        }
    }
  }
}]);

```

wodka:
a better way would be to have it compatible with the current layout (no changes to the html structure needed).

all options are supported -> if something is configured the wrong way it will only show an error in the console (and not break the whole angular app)

```javascript
(function (angular) {
    var app = angular.module('mm.swiper', []);

    app.directive('swiperContainer', [
        '$log',
        function($log) {
            return {
                restrict: 'C',
                link: function(scope, element, attributes) {
                    if (typeof attributes.breakpoints === 'string') {
                        attributes.breakpoints = JSON.parse(attributes.breakpoints);
                    }
                    if (typeof attributes.fade === 'string') {
                        attributes.fade = JSON.parse(attributes.fade);
                    }
                    if (typeof attributes.cube === 'string') {
                        attributes.breakpoints = JSON.parse(attributes.cube);
                    }
                    if (typeof attributes.coverflow === 'string') {
[...]

PR #1516, opened 2015-11-26
angularjs port

wodka:
basically implementation of https://github.com/nolimits4web/Swiper/issues/1131

all options are supported -> if something is configured the wrong way it will only show an error in the console (and not break the whole angular app)

```javascript
(function (angular) {
    var app = angular.module('mm.swiper', []);

    app.directive('swiperContainer', [
        '$log',
        function($log) {
            return {
                restrict: 'C',
                link: function(scope, element, attributes) {
                    if (typeof attributes.breakpoints === 'string') {
                        attributes.breakpoints = JSON.parse(attributes.breakpoints);
                    }
                    if (typeof attributes.fade === 'string') {
                        attributes.fade = JSON.parse(attributes.fade);
                    }
                    if (typeof attributes.cube === 'string') {
                        attributes.breakpoints = JSON.parse(attributes.cube);
                    }
                    if (typeof attributes.coverflow === 'string') {
                        attributes.breakpoints = JSON.parse(attributes.coverflow);
                    }

                    try {
                        var instance = element.swiper(attributes);

                        scope.updateSlides = function () {
                            instance.update();
                        };
                    } catch (e) {
                        $log.error('failed to init swiper, invalid options?', attributes, e);
                    }

                }
            };
        }
    ]);

    app.directive('swiperSlide', [
        "$log",
        function($log) {
            return {
                restrict: 'C',
                link: function(scope, element) {
                    if (scope.$last) {
                        if (!scope.$parent.updateSlides) {
                            $log.error('used swiper-slide class outside of swiper-container', element);
                        }
                        scope.$parent.updateSlides();
                    }
                }
            };
        }
    ]);
})(angular);
```

then without doing anything different:
```html
<div
        class="swiper-container"
        data-autoplay="0"
        data-loop="1"
        data-slides-per-view="3"
        data-breakpoints='{"300": {"slidesPerView":1},"500": {"slidesPerView":2}}'
>
    <div class="swiper-wrapper">
        <div class="swiper-slide" ng-repeat="data in someList">
            <div class="paddings-container">
                <ng-include src="'slide.html'"></ng-include>
            </div>
        </div>
    </div>
    <div class="pagination"></div>
</div>
```

### DIFF
--- a/src/js/angular.js
+++ b/src/js/angular.js
@@ -1,0 +1,56 @@
+if (typeof(angular) !== 'undefined') {
+    (function (angular) {
+        var app = angular.module('nolimits4web.swiper', []);
+      
+        app.directive('swiperContainer', [
+            '$log',
+            function($log) {
+                return {
+                    restrict: 'C',
+                    link: function(scope, element, attributes) {
+                        if (typeof attributes.breakpoints === 'string') {
+                            attributes.breakpoints = JSON.parse(attributes.breakpoints);
+                        }
+                        if (typeof attributes.fade === 'string') {
+                            attributes.fade = JSON.parse(attributes.fade);
+                        }
+                        if (typeof attributes.cube === 'string') {
+                            attributes.breakpoints = JSON.parse(attributes.cube);
+                        }
+                        if (typeof attributes.coverflow === 'string') {
+                            attributes.breakpoints = JSON.parse(attributes.coverflow);
+                        }
+      
+                        try {
+                            var instance = element.swiper(attributes);
+      
+                            scope.updateSlides = function () {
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thanks, will tweak it a bit

nolimits4web:
Sorry, reviewed again, and decided that i cannot merge it. First of all, i'm not familiar with Angular so i can not maintain it. Also it looks incomplete, how to use callbacks, how to access Swiper instance, etc. This probably should be in separate repo

wodka:
no problem - so there we have it: https://github.com/wodka/AngularSwiper

PR #1535, opened 2015-12-12
Fix activeclass

wardpeet:
Found bug when slidesPerView is a float you get an issue with the last slide never will be active.
I added a demo to the  folder 35-slidesperview-float.html


nolimits4web:
Hi, sorry, but it is not the solution. `snapGrid` and `slidesGrid` are for different purpose, it is wrong to get activeIndex from `snapGrid`, because for example with usage of slidesPerGroup or slidesPerColumn it has nothing common with slides

wardpeet:
Ok thanks for the answer, wasn't sure if it was the solution. I'll try to create another bugfix :)

PR #1538, opened 2015-12-15
Error fix for params[param].nodeType when params[param] is null 

hanjukim:
Error fix for params[param].nodeType when params[param] is null because null is an type of object.


nolimits4web:
Hi, thanks. Can you remove PR with changed to `dist/` folder and i will merge it? Otherwise let me know and i can merge it manually

hanjukim:
I made a new PR. Closing.

PR #1540, opened 2015-12-16
Removing unused variable autoplayStopOnLast which prevented stop looping

### LINKED ISSUE #1539: When autoplay is set and loop is false it is still looping
I want to let swiper automatically start, but not looping. I have the following configuration:

    var swiper = new Swiper("#mod-swiper", {
		direction: "horizontal",
		calculateHeight: true,
		speed: 1000,
		paginationClickable: true,
		nextButton: ".swiper-button-next",
		prevButton: ".swiper-button-prev",
		spaceBetween: 10,
		slidesPerView: "auto",
		loop: false,
		autoplay: 10,
		pagination: "#mod-swiper-pagination-container",
		centeredSlides: true,
		effect: "slide",
	});

After the page has loaded, swiper starts automatically to iterate trough the slides. After the last slide it starts again by the first one. But it should not as loop is set to false. Is there a configuration setting missing or is it a bug?

laoneo:
Couldn't find the variable autoplayStopOnLast, means it will always be false. Then it will never stop auto play and will always endless loop swiper.

This PR fixes #1539.


nolimits4web:
`means it will always be false` - it is only by default. If you set it to `true` in parameters then it will be `true`. It is pretty useful thing, no need to remove it. I'll better add it to documentation instead of removing and then adding it by someone request again :smile:

laoneo:
I agree that it would be helpful, but it is not initialized and also not documented. Would save me a couple of hours because it looked like relic from an older release.

laoneo:
Did you test that setting? For me it made no difference of I set it to true or false. I'm using the following code:

	var dpSlider' . $module->id . ' = new Swiper("#mod-dpslider-' . $module->id . '", {
		direction: "horizontal",
		calculateHeight: true,
		speed: 1000,
		paginationClickable: true,
		nextButton: ".swiper-button-next",
		prevButton: ".swiper-button-prev",
		spaceBetween: 10,
		slidesPerView: 1,
		loop: true,
		autoplay: 2,
		autoplayStopOnLast: ' . (!$params->get('loop', 1) ? 'true' : 'false') . ',
		centeredSlides: true
	});

It doesn't make any difference if the parameter loop is true or false. It always restarts the swiper to the beginning and runs in an endless loop.

nolimits4web:
Yes, it works as expected with `loop` disabled. With `loop` there is no much sense in it

laoneo:
For sure there is. If you want to start the slideshow automatically till the last slide and then stop. BUT you still want to offer that the user can hit the > button to move to the first slide.

PR #1567, opened 2016-01-08
Error fix for params[param].nodeType when params[param] is null 

hanjukim:
Error fix for params[param].nodeType when params[param] is null because null is an type of object.


nolimits4web:
Thanks

Issue #1574, opened 2016-01-13
number of bullets mistake


DorinLevi:
hello,
When I am using 2 rows of slides- and 4 in each group, 6 in each view.
I found a mistake in the calculation since when I have 14 slides, I should have 3 bullets and it appears 4.
Your calculation is: 
s.params.loop ? Math.ceil((s.slides.length - s.loopedSlides * 2) / s.params.slidesPerGroup) : s.snapGrid.length;
and in my case it  gets 4 as the s.snapGrid.length
my suggestion is:
parseInt(s.slides.length / s.params.slidesPerGroup * s.params.slidesPerColumn);

and in the same subject, to adjust the "next" button:
if (s.isEnd) - doesnt do the trick, it should be replaces with:
if ($(s.bullets[s.bullets.length - 1]).hasClass('swiper-pagination-bullet-active'))

can you please review it and give my your opinion if there is any chance your code will get an update.
Thank you.

PR #1580, opened 2016-01-17
re calculating number of bullets  for pagination due to bug when ther…

mssatt:
re calculating number of bullets  for pagination due to bug when there are 2 collums or more.

resolved issue: #1574


nolimits4web:
I will check it, but
1) PR should be done only to files in `src/` folder
2) Proposed fix is a bit wrong, state of Prev/Next buttons shouldn't rely on pagination

nolimits4web:
Fixed with c8798c218b4d13643ee1bf1c4bd53026bb2e097a

PR #1583, opened 2016-01-18
Update lazy-load.js

promoinvideo:
Added quote inside background-image URL to let url with special chararacters like "()" works inside browser.


nolimits4web:
Merged, thanks

PR #1594, opened 2016-01-30
Lazyload on resize

Kilbourne:
Allows to lazyload not loaded images when  the number of visible slide changes by resize


nolimits4web:
Thanks, but i think this fix should be in another place - in resize handler, i just pushed commit to fix this

PR #1600, opened 2016-02-04
Fix paginationClickable property in demo files to be boolean

WarriorRocker:
Noticed while creating type definitions for Swiper 3.3.0 that in 16-effect-fade.html and 25-hash-navigation.html the paginationClickable property was being passed a selector. In the documentation and after examining Swiper's internal use of this property it is treated only as a boolean.

Typescript defitions for Swiper 3.3.0
https://github.com/WarriorRocker/DefinitelyTyped/blob/swiper/swiper/swiper.d.ts

Let me know what you think.

### DIFF
--- a/demos/16-effect-fade.html
+++ b/demos/16-effect-fade.html
@@ -56,7 +56,7 @@
     <script>
     var swiper = new Swiper('.swiper-container', {
         pagination: '.swiper-pagination',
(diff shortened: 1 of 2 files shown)


nolimits4web:
Merged, thanks, my mistake

PR #1602, opened 2016-02-05
Fix resistance to enable/disable resistant bounds

antoniocosta:
Currently, setting resistance to false does NOT disable resistant bounds. This fixes that.


nolimits4web:
Can't replicate this issue, parameter works fine, setting `resistance: false` totally disables it, effect the same as setting `resistanceRatio: 1`

antoniocosta:
I don't understand what you call resistance then. Setting `resistance: false` produces no effect here. I thought it was supposed to be the same effect as `resistanceRatio: 0`

nolimits4web:
setting resistance to false means NO any resistance, so swiper can be moved freely

Issue #1625, opened 2016-02-21
onSlideNext and onSlidePrevEnd not being fired


natete:
Hey,

I'm trying to bind some functions to these events but they are never fired. I've been digging a little in the code and I think I've spotted the problem:

These events are emitted if activeIndex is < or > than previousIndex but both variables have always the same value because of the assignment in line 1859 of file core.js (s.previousIndex = s.activeIndex || 0;)

nolimits4web:
Go to http://idangero.us/swiper/demos/01-default.html
Add in console:
```
swiper.on('onSlideNextStart', function(){console.log('slide next')});
swiper.on('onSlidePrevEnd', function(){console.log('slide prev end')});
```
you may see in console everything works

natete:
Hello!

I've been digging in this exemple and I've find out that these callbacks are only being called when using slide effect. Maybe I'm wrong and these callbacks are only intended to work this way but I think it would be nice to have these callbacks also working with the rest of effects.

Here's a codepen showing the problem I'm trying to point out.

http://codepen.io/natete/pen/ONPYXE?editors=1010

nolimits4web:
And i still see callbacks with Fade effect http://codepen.io/anon/pen/aNdzbq?editors=1000 Where do you test it?

natete:
I'm using Chrome/Ubuntu.

I've been trying again and I've found out that it fails only under certain circumstances. I you drag the slide until the end of the transition and then release this callbacks aren't fired. If you drag and then release with a short swipe it works. I've added some more info in my codepen to show the problem. And also a gif because my explanation might be incomplete.

http://codepen.io/natete/pen/ONPYXE?editors=1000

![swiper](https://cloud.githubusercontent.com/assets/4098303/13611375/815815d2-e562-11e5-8c83-d85e74015b6c.gif)

nolimits4web:
In cube effect the `watchSlidesProgress` is enabled automatically, and in this case active index updated dynamically during touch interaction. Use `onTransitionStart/End` callbacks instead then

nolimits4web:
Issue is closed because of outdated/irrelevant/not actual/needed more information/inactivity.

If this issue is still actual and reproducible for latest version of Swiper, please create new issue and fill the issue template correctly:
* Clearly describe the issue including steps to reproduce when it is a bug.
* Make sure you fill in the earliest version that you know has the issue.
* Provide live link or JSFiddle/Codepen or website with issue

PR #1627, opened 2016-02-21
Fixes previousIndex assignment

natete:
The assignment made before causes previousIndex always being the same as activeIndex. Now it's initialized to 0 or keeps its value as it should instead of taking activeIndex value.

Related to #1625


nolimits4web:
This fix is wrong, as you may see on next line `activeIndex` will get the new value

nolimits4web:

PR #1652, opened 2016-03-22
added option to use history.replaceState() to not break browser history

hochitom:


nolimits4web:
Thanks, i merged this manually as PR shouldn't touch `dist/` files and renamed parameter to `hashnavReplaceState`

PR #1665, opened 2016-03-30
added touchend to next/prev/dots click events

beard86:
I've been using Swiper on a desktop sized touch screen and needed to modify the code in order for the next & previous buttons to function.


nolimits4web:
Sorry, but this will cause wrong behavior on mobile devices, if you start swiping from somewhere and stop on (for example) Next button, it will cause slide to next slide, which can be not what user was expecting

Issue #1673, opened 2016-04-07
Request: disableTouchControl() and enableTouchControl() methods


azaslavsky:
As mentioned in [this issue](https://github.com/nolimits4web/Swiper/issues/1235), I think it would a pretty useful feature to disable user swiping controls, without "locking the swiper entirely."  Currently, `.lockSwipes()` seems to work by completely preventing all slide changes, even those triggered from the API using the `.slideTo()` method.  What would be useful is a method that would prevent the user from sliding, but still enable the developer to programmatically move slides in the swiper.  I feel like an API that mimics `.disableKeyboardControl()` and `enableKeyboardControl()` would fit in nicely with the available methods.

azaslavsky:
So I figured out that modifying the param `onlyExternal` like so:

`s.params.onlyExternal = true; //or false, if you so desire`

Since this is so simple, I think I'll make a PR implementing these methods for consistency, if you don't mind.

nolimits4web:
Thanks, closed by #1674

PR #1674, opened 2016-04-07
Add enableTouchControl() and disableTouchControl()

azaslavsky:
Implements the feature request described in #1673. Simply a couple of wrappers for toggling s.params.onlyExternal after the Swiper has been instantiated.


nolimits4web:
Looks ok, merged, thanks

PR #1681, opened 2016-04-14
Handle AutoHeight while scrolling every 150ms in freemode

mrnix:
Throttle wrapper for scroll event. It may be useful in cases of different slides heights


nolimits4web:
Sorry, can you describe a bit what does it do? Maybe with example showing difference with and without this commit?

nolimits4web:

Issue #1696, opened 2016-04-26
Need a way to get the accurate activeIndex in loop mode


phongmedia:
I'm using Swiper for a user to set their 'current active image', and so hooking into the `onSlideChangeEnd()` callback.

As you know, when in loop mode, the `activeIndex` property is offset by the duplicated slides, although what I need is essentially the `data-swiper-slide-index` attribute of the currently selected slide. I can't find a way to get the currently selected slide or it's index.

I looked repeatedly for this value in the API, although I can't find anything, it seems like it should be there somewhere. The only other way I can think to get this value, would be to query the DOM for the slide with the active class on it, and grab it's `data-swiper-slide-index` value, although that's a pretty messy way to get a simple value.

navyxie:
in loop mode active index value will be always shifted on a number of looped/duplicated slides. you can get attribute 'data-swiper-slide-index' whit function like:

navyxie:
in loop mode active index value will be always shifted on a number of looped/duplicated slides. you can get attribute 'data-swiper-slide-index' whit function like:

```js
function getActiveIndex(swipe){
    var activeIndex = swipe.activeIndex;
    var slidesLen = swipe.slides.length;
    if(swipe.params.loop){
        switch(swipe.activeIndex){
            case 0:
                activeIndex = slidesLen-3;
                break;
            case slidesLen-1:
                activeIndex = 0;
                break;
            default:
                --activeIndex;
        }
    }
    return  activeIndex;
}
```

usage:

```js
var mySwiper = new Swiper('.swiper-container', {
    direction: 'vertical',
    loop:true,
    onSlideChangeEnd:function(swipe){
        console.log(getActiveIndex(swipe))
    }
}); 
```

@phongmedia

phongmedia:
@navyxie Thanks for the function, I'll give that a spin. It would be great if something like this was already provided by the API, with something like `swiper.getActiveIndex()`

navyxie:
@phongmedia    Sorry , Above way is not better , and will take a bug  when you set param loopAdditionalSlides or loopedSlides:

```js
var mySwiper = new Swiper('.swiper-container', {
    direction: 'vertical',
    loop:true,
    loopAdditionalSlides:1,
    loopedSlides:2,
    onSlideChangeEnd:function(swipe){
        console.log(geSlideDataIndex(swipe))
    }
}); 
```

[...]

PR #1697, opened 2016-04-27
add property realIndex

navyxie:
add property realIndex:
```js
s.updateRealIndex = function(){
    s.realIndex = s.slides.eq([s.activeIndex]).attr('data-swiper-slide-index') || s.activeIndex;
};
```

from issue : [Need a way to get the accurate activeIndex in loop mode](https://github.com/nolimits4web/Swiper/issues/1696)

from stackoverflow : [iDangero.us Swiper slide count when loop is true](http://stackoverflow.com/questions/29348926/idangero-us-swiper-slide-count-when-loop-is-true/36880901#36880901)


nolimits4web:
Looks good, thanks. But PR shoud be done only to `src/js/core.js` file without files in dist/ folder. Also:
```
s.slides.eq([s.activeIndex])
```
should be just
```
s.slides.eq(s.activeIndex)
```
Can you do these changes and i will merge it? Otherwise i'll merge it later manually

navyxie:
@nolimits4web  It's ok ,thanks.

nolimits4web:
Merged, thanks

pvaladez:
Hey not sure if this is by design, but the realIndex property is a string - could it instead be an integer?

PR #1700, opened 2016-04-29
new lazyLoadingClass parameter

promoinvideo:
Added new param lazyLoadingClass usefull when you have nested swiper using lazy load and if you don't want that the main swiper preloads all slide images of nested swipers.

### DIFF
--- a/swiper.jquery.js
+++ b/swiper.jquery.js
@@ -1,0 +1,3779 @@
+/**
+ * Swiper 3.3.1
+ * Most modern mobile touch slider and framework with hardware accelerated transitions
+ * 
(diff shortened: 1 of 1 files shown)


nolimits4web:
Can you make commit only to related JS files in `src/` folder?

nolimits4web:
Merged manually, thanks

promoinvideo:
Sorry Vladimir, I was busy with my daytime work this week. 
If you would like to view some real works I'have made using Swiper you could check http://videolibro.esseoquattro.it/ or http://www.promoinvideo.com. Thanks for your help and for you excellent job.
-Alberto Cobre

PR #1762, opened 2016-06-28
Accurate center button prev/next

gerardcuadras:
Center vertically `.swiper-button-next` and `.swiper-button-next` in all conditions.


nolimits4web:
This can be a solution but can't merge it as it may break existing swiper integration styles with update

PR #1797, opened 2016-07-27
Cross browser mousewheel

icek:
This is rewritten version of mousewheel.js based on solution from [facebook fixed-data-table mormalizeWheel.js](https://github.com/facebook/fixed-data-table/blob/master/src/vendor_upstream/dom/normalizeWheel.js).
This should fix most of mouse wheel related issues.


nolimits4web:
Thanks, had to merge it manually as the PR shouldn't touch `dist/` files

icek:
Sorry about that, next time I will remember. Thank you for merging despite my mistake.

PR #1805, opened 2016-08-02
Add possibility to change all class names

mrksbnch:
Currently, it's not possible to change all class names in the settings. I also fixed some hardcoded class names (e.g. `slideClass` or `slideDuplicateClass`, see `src/js/keyboard.js`) that could already be changed in the settings.


nolimits4web:
Merged, thanks

PR #1832, opened 2016-08-30
Fixing autoHeight if slidesPerView is more than 1

raahede:
Currently Swiper only measures the height of the active slide when setting container height. This fix also looks at other slides in view, and uses the tallest slide's height.

**NOTE:** This fix doesn't take into account sliders with `slidesPerView: 'auto'`, it only ignores it. A more elaborate fix is needed to support that.


nolimits4web:
Looks good, thanks, medged

PR #1833, opened 2016-08-30
Fixing IE9 issue.

radojesrb:
The following use-case was causing the regression error bug:
- Using 'dist/js/swiper.jquery.min.js' on IE9
- Setting 'loop' property on instantiation of Swiper to 'true'
- Using arrow as navigation (swiping was working fine)

If this use-case is happening (requirement on my project) 's.animating' within the 'core.js' will be re-assigned  to 'true' on each click on the arrow/s. Resulting in only one slide change (initial click).

Solution:
- added the lte (lower then or equal) IE9 conditional check in 'src/js/swiper-proto.js' > property 'browser' > property 'lteIE9' (detecting browser by feature instead of version or UA etc.)
- wrapping line 1907 in 'src/js/core.js' method 'slideTo' into the IF statement (if (!s.browser.lteIE9) s.animating = true;). Idea is not to do the re-assignment on 's.animating' if code is being executed on IE9 or lower version of IE.

Took me some time to get into your nice code and debug our archenemy aka IE...

### DIFF
--- a/demos/38-jquery-ie9-loop.html
+++ b/demos/38-jquery-ie9-loop.html
@@ -1,0 +1,88 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8">
+    <title>Swiper demo</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1, maximum-scale=1">
+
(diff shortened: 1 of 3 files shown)


nolimits4web:
Looks good, thanks

PR #1853, opened 2016-09-13
Added history api support with fallback to hashnav.

jferrie-urbandaddy:
Parameter: history
Type: string
Description: Set to string to match backend routing. Set data-history on slide.
For example:

var swiper = new Swiper('.swiper-container', {
    history: 'slug'
});

<div class="swiper-slide" data-history="1">Slide 1</div>

String will be used in the url: domain.com/[slug]/[data-history].

If hashnav and history are both set by hashnav will be disabled. If history is
not supported it will fallback to hashnav.

Changes: Deprecated hashnavReplaceState changed to replaceState for use in
hashnav and history.


nolimits4web:
Looks good, thanks, merged

nolimits4web:
Had to merge it manually as `dist` files shouldn't be touched during PR

PR #1879, opened 2016-10-12
update navigation.less to calc 50%

theskillwithin:

Always follow the [contribution guidelines](https://github.com/nolimits4web/Swiper/blob/master/.github/CONTRIBUTING.md) when submitting a pull request. 

using transform: translateY(-50%); so that the top: 50% is measured from the center of the element and not the top.


nolimits4web:
Sorry, this PR will give wrong result as it still keeps negative margin-top. Also it may break existing implementations with Swiper update

Issue #1893, opened 2016-10-19
Incorrect offset for lazy loading with >1 slidesPerGroup


matt-oconnell:
This is a (multiple allowed):
* [x] bug

* Swiper Version: 3.4.0
* Platform/Target and Browser Versions: All 

### What you did
Configuring with *more than 1 `slidesPerGroup` and `lazyloading/lazyLoadingInPrevNext: true`

Relevant config settings: 
```
{
  slidesPerGroup: 3,
  slidesPerView: 3,
  slidesPerColumn: 2,
  lazyLoading: true,
  lazyLoadingInPrevNext: true
}
```

### Expected Behavior

Swiper is configured with a 2 x 3 grid (6 elements.) Lazy load happens when the next 6 elements are transitioned to.

### Actual Behavior

Lazy load images are offset incorrectly. 
![screen shot 2016-10-19 at 2 39 10 pm](https://cloud.githubusercontent.com/assets/8006041/19532478/e4e8c292-9609-11e6-991a-115a3c7b5612.png)

matt-oconnell:
see here: https://jsbin.com/vujekucexi/edit?html,output

PR #1894, opened 2016-10-19
Fix activeSlide bug on lazy load

matt-oconnell:
This resolves: https://github.com/nolimits4web/Swiper/issues/1893

Originally, the offsets were incorrect because `s.activeIndex` was not taking into account multiple slides per column.


matt-oconnell:
accidentally added build files

PR #1897, opened 2016-10-20
Lazy load

jrefano:
This resolves: https://github.com/nolimits4web/Swiper/issues/1893

Originally, the offsets were incorrect because `s.activeIndex` was not taking into account multiple slides per column.

See example here: https://jsbin.com/vujekucexi/edit?html,output

cc @matt-oconnell


nolimits4web:

PR #1908, opened 2016-10-25
Fixes an error on swiper.destroy() when scrollbar is enabled

sconix:
This fixes a typo in the disableDraggable function which causes swiper.destroy() to error out when the scrollbar is enabled.


fredrik-macrobond:
Hope they will fix this, Sconix. Glad you found it though!

nolimits4web:
Merged, thanks!

sconix:
Would it be possible to get new release with now when this fix is merged? I would like to release a new version of the Angular 2 Swiper wrapper library and this is blocking it since with AOT support I can't patch the Swiper anymore in the library package.

PR #1912, opened 2016-10-31
Fix formatting on v2 docs

kevinhikaruevans:
Just fixed the numbering and formatting on the v2 docs.

The numbered list wasn't aligning correctly and it just bugged me a little.


nolimits4web:
Thanks!

PR #1915, opened 2016-11-01
Fix : activeIndex not changed when more than one slidePerView at the end of the swiper

thollander:
When you have more than one slidePerView in the swiper and when you are at the end of it, if you click on one of the last visible slides, you don't need to slide to see them. However, it is good to change the activeIndex (to have the activeClass for example) on the clicked slide.

The initial slideIndex were changed when we check if we need to slide to see the clicked slide, therefore, the activeIndex were always the first visible slide.


nolimits4web:

PR #1938, opened 2016-11-24
Bug fix: hashnav attachEvents did not always launch at init.

kosminen:
Previously, attachEvents did not fire when there was no hash, since init function returned too early. Thus hashnavWatchState did not work properly; Swiper did not slide anywhere when document.location changed.

All I did was prevent the if statement from returning early, and moved the existing code inside the if statement block.


nolimits4web:
Thanks!

PR #1991, opened 2017-01-25
Emit keyPress event at the end of keyboard control

unicod3:
I needed to bind custom functions for custom key codes so i just emit a new event which is onKeyPress


nolimits4web:
Merged, thanks

Issue #2046, opened 2017-03-15
[Bug] Slide to last item (unwillingly) when carousel is hidden


nova-tan:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: V3.4.2
* Platform/Target and Browser Versions: I have replicated this bug on Chrome and Safari and I think it'll happen to any browser, but it's mostly visible on iOS Safari because a scroll triggers `resize` event.

### What you did
Simple initialization with slidesPerView set to `auto`. `new Swiper('.swiper', { slidesPerView: 'auto' })`

### Expected Behavior
When the parent dom of `'.swiper'` is the to `display: none` and I scroll the page up and down a bit in iOS Safari. When the parent dom is set back to `display: block` and I scroll the page up and down, I should see the active slide remains the same.

### Actual Behavior
When the parent dom of `'.swiper'` is the to `display: none` and I scroll the page in iOS Safari a bit. When the parent dom is set back to `display: block` and I scroll the page up and down, I see the active slide becomes the last item of carousel.

What happened:
In iOS safari, page scroll will trigger the resize event which causes Swiper to update progress. Because dom is invisible and the width of each item becomes 0. The difference of maxTranslate()
and minTranslate() becomes 0 and Swiper thinks the carousel has reached to its end as well as beginning. See [this line of code](https://github.com/nolimits4web/Swiper/blob/94d064f5f52b50f0a31a09ae47e85b760963953c/dist/js/swiper.js#L994). When the dom is set to visible again, another scroll triggers a resize event again. Because `isEnd` is set to true, Swiper will slide the carousel to the last item programmatically. See [here](https://github.com/nolimits4web/Swiper/blob/94d064f5f52b50f0a31a09ae47e85b760963953c/dist/js/swiper.js#L1309)

This behavior can be produced in browser other than iOS safari as long as a resize event is triggered when dom is hidden.

PR #2047, opened 2017-03-15
do not resize when slides are hidden

nova-tan:
This is to fix this issue: https://github.com/nolimits4web/Swiper/issues/2046


nolimits4web:
Merged, thanks

PR #2058, opened 2017-03-28
Better cleanup when instance is deleted

mems:
Clean up jQuery data and object reference attached to the DOM element, remove reference to container and wrapper.


nolimits4web:
Thanks

PR #2089, opened 2017-04-26
Use translate3d only if really needed

L1Q:
Translate3d caused me troubles couple times in the past with different problems.
Today I found another problem with it: it broke visibility of an element positioned and transition-translated over the slider.

Please comment if you need any more detailes and I'll provide you with more info and examples.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -1112,7 +1112,7 @@ s.updateClasses = function () {
                 scaleY = scale;
                 scaleX = 1;
             }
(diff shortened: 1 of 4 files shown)


nickdevereaux:
FYI as far as I know `translate3d` uses the GPU where possible unlike `translate`, so I'm not sure if this is beneficial to everyone. Might be worth making this an option to be able to do both?

nolimits4web:
Closing as incompatible with current version of Swiper

PR #2123, opened 2017-06-06
Swiper with server side render

alom2:
I realy like your library, but i was having a problem with a aplication witch uses React with server side render, and to solve this problem i needed to add a verification every time you used window, because in serve side render we dont have window.

I have create this pull request to show you what i did and someone can have this same problem.


jesseweed:
+1 would love to see this get merged!

surepe:
Hey Andre!

Thanks for this fix, but I have found another issue. 
Section's height is wrong untill the browser window is resized. Not sure how those 6399px is calculated, but after resizing it becomes normal 970px (for 1920x1080 chrome window)

alom2:
Hello @surepe,
When i used this with Angular 4, i din't have any problem, but when i was using ReactJs I had the same problem, dont know the best way to solve this problem, Swipper doesn't fit the window and i haven't much time to solve so i used a setTimeout this will make Swipper to fit the window and apply all the options.

PsychicCat:
please merge this!!!!

haffla:
I am using Swiper successfully in a server side rendered setup (Vue). But I don't run the Swiper code on the server. Something like

``` javascript
if (typeof window === 'object') {
  // init swiper
}
```

nolimits4web:
Closing as incompatible with current version of Swiper and probably not needed anymore

PR #2143, opened 2017-06-30
Use clickable class instead of modifier with postfix

Manduro:
A specific constant exists for the `pagination-clickable` class, but isn't currently used. Now it does!


nolimits4web:
Thanks!

Issue #1800, opened 2016-07-29
Problem with loop and autoplay


markbuettner:
We're using Swiper in one of our shops. We have one slider with 17 items in it. Swiper is initialized with the following code:

     var car_swiper = new Swiper('#home_slider_1', {
    	nextButton: '#car-slider .svg-button-next',
        prevButton: '#car-slider .svg-button-prev',
    	loop: true,
    	slidesPerView: 5,
     	slidesPerGroup:5,
    	autoplay: 4000,
        paginationClickable: true
    });

The problem is, as soon as the Swiper shows the last two items and the first three, this slide is stuck in an endless loop. 
It seems that this is a problem with the combination of loop and autoplay, using console.log I saw that the slideTo(slideIndex, speed, runCallbacks, internal) function is called three times with 3, 17 and 20 as slideIndex.

The live example can be found at [https://www.lott.de/de/](https://www.lott.de/de/) on the bottom of the page. The affected slider is the vehicle manufacturer slider.

nolimits4web:
The issue here is that with such configuration you must have number of slides even to 5, so as a workaround i recommend you to add 3 empty slides to have 20 in total

PR #2150, opened 2017-07-09
Add blank slides to fit slides per group

xu3u4:
I encounter the same problem as #1800 
When using infinite loop with slidesPerGroup, some slides will repeat if the number of slides doesn't fit the group.

For example, 
If there are 10 slides with following settings:
```js
slidesPerView: 4,
slidesPerGroup: 4,
loop: true,
```
it becomes:
page 1: 1, 2, 3, 4
page 2: 5, 6, 7, 8
page 3: 9, 10, 1, 2
page 1: 1, 2, 3, 4

1 and 2 are repeated in page 3.
Thus, this PR adds the feature to fit the last page with blank slide(s) when the new parameter `fitSlideGroupWithBlank` is set to true.
Then page 3 will become: 9, 10, (empty), (empty)

### DIFF
--- a/demos/40-infinite-loop-with-multiple-slides.html
+++ b/demos/40-infinite-loop-with-multiple-slides.html
@@ -1,0 +1,90 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8">
+    <title>Swiper demo</title>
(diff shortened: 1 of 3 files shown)


nolimits4web:
Looks good, thanks!

PR #2154, opened 2017-07-13
Export defaults in Swiper object

cesarizu:
This PR exports the defaults object as Swiper.defaults so others can use it. This is specially useful for wrappers like ember-cli-swiper that need to manage the parameters that the components can use.

See https://github.com/Suven/ember-cli-swiper/pull/32


nolimits4web:
Thanks

PR #2164, opened 2017-07-25
add CSS Modules support

emolchanov:
This PR resolves problem with using Swiper and CSS Modules in project.

If I use swiper with css modules - I can't override auto-prefixed classes and write styles for them.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -153,7 +153,6 @@ var defaults = {
     // Passive Listeners
     passiveListeners: true,
     // NS
-    containerModifierClass: 'swiper-container-', // NEW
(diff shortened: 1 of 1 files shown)


nolimits4web:
Well, got your point, but this PR brings breaking changes, and such parameters doesn't look pretty usable. I think i'll implement in version 4.x which i am working on

emolchanov:
@nolimits4web so, version 4.0 has been released and what about CSS Modules?

nolimits4web:
No CSS modules for now. If it can be added without changing parameters to something so scary :) i can accept it. But anyway with new Swiper modular structure, it can be done by creating a plugin for Swiper

PR #2188, opened 2017-08-21
Prevent adding extra blank items

xu3u4:
It is related to https://github.com/nolimits4web/Swiper/pull/2150
Sorry, I forgot to update the situation that if slides.length can be divided by slidesPerGroup.
In this case, there will be one more blank group in the end.


nolimits4web:
Thanks

Issue #2003, opened 2017-02-03
Slider with slidesPerView: 'auto' when pressing previous button on the last slide(slider must have more than 2 slides) it moves back 2 slides instead of 1.


kursez:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 3.4.1
* Platform/Target and Browser Versions: All browsers.

### Environment and action
In a slider with the option slidesPerView: 'auto', that has more than two slides, press previous button.

### Expected Behavior
Expected to go back one slide.

### Actual Behavior
Goes back 2 slides.

kencanak:
Hey there, encountering the same issue. is there any fix for this?

levani:
I have the same issue. Prev button jumps to start when slidesPerView is auto.

Here is a jsfiddle demo: https://jsfiddle.net/3ua6ktjd/

Next button goes on slide at a time but prev button goes at the beginning. This looks like a bug.

rmalveis:
+1
I have the same problem.. no Solution found, yet. 
I could track that when we reach the last slide, swiper does not update "slide-prev" and "slide-active".

wszgxa:
+1
same problem. just jump 2 slides..

nolimits4web:
Issue is closed because of outdated/irrelevant/not actual/needed more information/inactivity.

If this issue is still actual and reproducible for latest version of Swiper, please create new issue and fill the issue template correctly:
* Clearly describe the issue including steps to reproduce when it is a bug.
* Make sure you fill in the earliest version that you know has the issue.
* Provide live link or JSFiddle/Codepen or website with issue

PR #2198, opened 2017-09-03
fix prevButton will jump 2 slide

wszgxa:
#2003 
this PR is to solve the problem when you set a `slidesPerView: 'auto'` and press previous button on the last slide, it moves back 2 slides instead of 1.


nolimits4web:
Closing as incompatible with current version of Swiper. If you think the issue is still replicable in v4, open PR for it with corresponding changes

PR #2216, opened 2017-09-26
Bugfix: flexbox support check

Garbanas:
Since "support" is no longer a property of the Swiper core class, a check for "support.flexbox" throws an error for an invalid object access.

### DIFF
--- a/src/components/core/update/updateSlides.js
+++ b/src/components/core/update/updateSlides.js
@@ -141,7 +141,7 @@ export default function () {
     rtl && wrongRTL && (params.effect === 'slide' || params.effect === 'coverflow')) {
     $wrapperEl.css({ width: `${swiper.virtualSize + params.spaceBetween}px` });
   }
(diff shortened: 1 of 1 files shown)


Tronix117:
This should absolutely be merged asap !

nolimits4web:
> Since "support" is no longer a property of the Swiper core class

Why? It is still available as `swiper.support.flexbox`

Tronix117:
On v3 yes, on v4 it seems there is an issue because `swipes.support` is undefined on IE10. I didn’t investigate further than this PR though, so maybe the issue is somewhere else.

Garbanas:
@nolimits4web 
To clarify: I'm using the Swiper node module in version 4.0.0-beta.4.
`swiper.support.flexbox` is the only occurrence of `swiper.support` in the code - everything else references the new Support module.

nolimits4web:
Weid, but i am ok with using `Support`. Thanks!

PR #2231, opened 2017-10-08
Remove obsolete prefixed event names

blikblum:
This PR removes obsolete transitionend prefixed events as discussed at https://github.com/framework7io/Framework7/issues/1288 and at https://github.com/framework7io/Framework7/pull/1849


nolimits4web:
Merged, thanks!

PR #2247, opened 2017-10-11
Add jsDelivr hits badge

LukasDrgon:
Hey, I noticed that Swiper gets 8M hits per month on jsDelivr, so I thought you might like this badge.


nolimits4web:
Thanks

PR #2248, opened 2017-10-12
Set default loopedSlides to avoid NaN

### LINKED ISSUE #2243: loop error if in breakepoints
in page onload size in breakepoints has error
Uncaught TypeError: Cannot read property 'classList' of undefined
but onload size not in breakepoints if resize window works well
here is my demo https://jsfiddle.net/8vf1aph6/1/

mitchheddles:
Sets a default loopedSlides count to avoid NaN. The problem resulted in `swiper.slideTo(NaN, 0, false)`. It only occurred when using using both loop and breakpoints, and initialising on a viewport within a breakpoint.

Also Fixes #2243

### DIFF
--- a/src/components/core/breakpoints/setBreakpoint.js
+++ b/src/components/core/breakpoints/setBreakpoint.js
@@ -21,7 +21,7 @@ function getBreakpoint(breakpoints) {
 
 export default function () {
   const swiper = this;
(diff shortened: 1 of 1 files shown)


nolimits4web:
Thanks

PR #2250, opened 2017-10-12
Fixed a problem with loading plugins if Object.prototype has custom functions.

MarcAMo:
Fixed an error in src/js/plugins.js, that occurs while loading the plugins when functions have been added to Object.prototype at runtime.

This was due to any custom function in Object.prototype being loaded as a plugin and failing in that.
Added a check if the plugin is an own property of the plugins object.


nolimits4web:
Can you please revert changes in `dist/` files?

MarcAMo:
Of course. 
Sorry if that causes inconvenience. I'm new to this.

nolimits4web:
Thanks

PR #2285, opened 2017-10-27
Fix prevented click event

### LINKED ISSUE #2051: Swiper JS breaks the SELECT MULTIPLE TAG
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: Swiper-3.4.2
* Platform/Target and Browser Versions: MAC Chrome 57

### What you did
It seems that after adding the swiper.min.js to my HTML page, the following selection box stopped working: 
```
<select multiple>
  <option value="volvo">Volvo</option>
  <option value="saab">Saab</option>
  <option value="opel">Opel</option>
  <option value="audi">Audi</option>
</select>
```
It is supposed to be a multiple selection box. However, it seems that I cannot select or highlight any options, at all. I already tried several ways to fixed this bug (all failed): 
I tried to add the following CSS: 
```
select[multiple]:focus option:checked {
  background: red linear-gradient(0deg, red 0%, red 100%) !important;
}
```
But it seems that swiper completely overwrote my css file. 

I also tried to add the following code to js: 
```
   jQuery('select').on('mousedown touchstart MSPointerDown', function(e){
           e.stopPropagation();
   }); 
```
Still no luck. 

Then I tried to add the swiper-no-swiping class to the select multiple tag. Still no luck.

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](http://stackoverflow.com/questions/tagged/swiper)
for that.

Before you open an issue, please check if a similar issue already exists or has been closed before.

arc279:
Fixes #2051

### DIFF
--- a/src/components/core/core-class.js
+++ b/src/components/core/core-class.js
@@ -185,7 +185,7 @@ class Swiper extends SwiperClass {
         startTranslate: undefined,
         allowThresholdMove: undefined,
         // Form elements to match
-        formElements: 'input, select, textarea, button, video',
+        formElements: 'input, select, option, textarea, button, video',
         // Last click time
         lastClickTime: Utils.now(),
         clickTimeout: undefined,


nolimits4web:
Thanks!

PR #2415, opened 2018-01-08
check for window.PointerEvent, fixes #2410

### LINKED ISSUE #2410: Feature detections broken since navigator.pointerEnabled and friends are removed.
Edge removed the non-standard `navigator.pointerEnabled` and MS prefixed forms. This means the feature detects used in Swiper fail for it now. Would doing something like `"onpointerdown" in window` be a good compromise?

DanielRuf:


src/utils/browser.js
```diff
@@ -17,9 +17,9 @@ const Browser = (function Browser() {
   return {
     isSafari: isSafari(),
     isUiWebView: /(iPhone|iPod|iPad).*AppleWebKit(?!.*Safari)/i.test(window.navigator.userAgent),
-    ie: window.navigator.pointerEnabled || window.navigator.msPointerEnabled,
+    ie: window.navigator.pointerEnabled || window.PointerEvent || window.navigator.msPointerEnabled,
```
jdalton: What's the `ie` check for? If it's just for pointer events maybe the detect should be renamed as other browsers support it too.

src/components/core/core-class.js
```diff
@@ -167,7 +167,7 @@ class Swiper extends SwiperClass {
       touchEvents: (function touchEvents() {
         const touch = ['touchstart', 'touchmove', 'touchend'];
         let desktop = ['mousedown', 'mousemove', 'mouseup'];
-        if (window.navigator.pointerEnabled) {
+        if (window.navigator.pointerEnabled || window.PointerEvent) {
```
jdalton: I notice multiple checks of the same properties. Could this be distilled into a single property to check?
This would make changes in the future easier too.

DanielRuf: What about older implementions and old browsers?

src/utils/browser.js
DanielRuf: I guee it is a check for IE / Edge to see if touch is supported. There are also other checks. But you are right, this is partly browser sniffing.

src/components/core/core-class.js
jdalton: It's not removing support for older browsers it's moving all of the duplicated checks to a property.
So instead of `if (a || b || c || d || e)` in multiple places it becomes `if (check.foo)` in multiple places.

DanielRuf: If you mean my change: I just added it where it is needed.
I agree that we should moe it to one global check.

DanielRuf: @nolimits4web what do you think?

jdalton: > I agree that we should moe it to one global check.

Ya, maybe better for a follow-up PR separate from this.

DanielRuf: Right, might be candidate for some refactoring in a separate PR.

nolimits4web:
Thanks guys, i'm refactoring now this thing with multiple checks

PR #2424, opened 2018-01-12
[Fix] Fix issue when append slides to rtl mode swiper goes to first s…

ahmad-sa3d:
Dear nolimits4web 👍  thank you for your awesome library,

i used your library in RTL mode and i encountered an issue when dynamically append slides to swiper as follows:

if swiper active slide Index is not first slide then i append slides to swiper and run update() method
swiper goes to first slide automatically and expected is that swiper should stay at current activeIndex, also this issue does not exists in LTR.

i fixed this issue on another branch and run 'npm run test' to make sure linter is OK, i also edited 'playground' example by adding a new slider to represent that issue is fixed

Thanks Bro, and i hope that this fix will be useful.


nolimits4web:
Looks good, but can you please refactor PR or send new one with changed only in `/src/` folder files?

ahmad-sa3d:
Ok, i will make new pull request to new branch with changes in src folder only

ahmad-sa3d:
Oh, Thanks, i fetch your new updates and i found that you have applied this fix, so no need to make new pull, thanks man.

PR #2429, opened 2018-01-15
Fix detection statement whether a swiper is in the viewport - fixes #2380

### LINKED ISSUE #2380: Keyboard - navigation through slides not working
This is a (multiple allowed):
* [x] bug

* Swiper Version: 4.0.7
* Platform/Target and Browser Versions: Chrome 62 on Mac OS X 10
* Live Link or JSFiddle/Codepen or website with isssue: http://idangero.us/swiper/demos/

### What you did
Navigation through slides with keyboard is not working when the slider is not on the top of the page. You can see it also on your demos page: http://idangero.us/swiper/demos/
Or you can visit http://idangero.us/swiper/demos/260-keyboard-control.html directly and set `body{padding-top: 101vh;}`

### Expected Behavior
If the slider is in the viewport, it should slide with keyboard arrows.

### Actual Behavior
It doesn't slide.

I guess the problem is in the `handle` method of the `Keyboard` class, where you check if swiper is inside of visible area of window. I think you are using the `offset()` function for the `swiper.$el` wrongly, because it returns relative offset to actual viewport so you don't need to recalculate it once again in `Keyboard.handle()` method.

gogy27:
This pull request should fix the problem with keyboard navigation. The problem was with detection of the swiper inside of the viewport.
The previous code did an unnecessary calculations, but the actual offset state is properly calculated within `offset` method of the `Dom7` library so there is no need to do it once again.

### DIFF
--- a/src/components/keyboard/keyboard.js
+++ b/src/components/keyboard/keyboard.js
@@ -28,10 +28,6 @@ const Keyboard = {
       if (swiper.$el.parents(`.${swiper.params.slideClass}`).length > 0 && swiper.$el.parents(`.${swiper.params.slideActiveClass}`).length === 0) {
         return undefined;
       }
-      const windowScroll = {
-        left: window.pageXOffset,
-        top: window.pageYOffset,
-      };
       const windowWidth = window.innerWidth;
       const windowHeight = window.innerHeight;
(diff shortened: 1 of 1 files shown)


nolimits4web:
Thanks

PR #2447, opened 2018-01-22
dynamic bullets: multiple main bullets

tzvikaf:
Hi,
right now the dynamic bullets in the pagination are limited to one active and 4 more, 2 for each side.
i added code to have more main bullets instead of just one, 
it feels more smooth when there is room for the active bullet to move in.

i'm using your code in a mobile-web project at my company
and i would really like if the code would become part of the package instead of me using locally the modified version.

i'll be happy to hear any comments,
have a nice day,
tzvika

### DIFF
--- a/src/components/pagination/pagination.js
+++ b/src/components/pagination/pagination.js
@@ -194,7 +194,7 @@ const Pagination = {
       $el.addClass(`${params.modifierClass}${params.type}-dynamic`);
       if (typeof params.dynamicBullets !== 'object') {
         params.dynamicBullets = {
-          numOfMainBullets: 1
+          numOfMainBullets: 1,
         };
       }
       params.dynamicBullets.indexOnMainBullets = 0;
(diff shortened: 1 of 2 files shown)


tzvikaf:
use example:
var swiper = new Swiper('.swiper-container', {
      pagination: {
        el: '.swiper-pagination',
        dynamicBullets: {
          numOfMainBullets: 4
        }
      },
    });

nolimits4web:
Looks good, thanks! I did some small renaming and fixed lastIndex calculation, was wrong a bit

tzvikaf:
thanks :)

tzvikaf:
what about documentation?

nolimits4web:
Will be update with new version release. If you want to add it then PR is welcome to https://github.com/nolimits4web/swiper-website

tzvikaf:
cool, i'll prepare something till the weekend

PR #2464, opened 2018-02-02
fix: Add custom class to scrollbar draggable + Fix scrollbar event

ianvieira:
Almost all the classes from Swiper can be customised, but the scrollbar draggable is not possible to change, as it's hard coded. I created a new option called "dragClass" and changed the element creation.

Also, the scrollbar is not working on desktop at the master version, it's trying to use mobile events when in fact should be mouse events (makes no sense need to enable the simulateTouch to have a working scrollbar, as we have the draggable option at scrollbar component).


nolimits4web:
Thanks, but few points:
* no changes in `/dist/` files
* it must pass test
* events shouldn't be changed, as it will more like break Windows touch devices. It is better to change `params.simulateTouch` condition to: 
```
params.simulateTouch || swiper.params.scrollbar.draggable

ianvieira:
@nolimits4web This fix for the events is not working, the browser is always entering in the this if:
`if (!Support.touch && (Support.pointerEvents || Support.prefixedPointerEvents)) {`
because the Support.pointerEvents is true, even on Mac where we have no touch events. I'll try to fix it with another logic.

ianvieira:
@nolimits4web I abstracted the touchEvents name generator function to Utils, this way, we can have different events for the carrousel and the scrollbar, as the conflict was in the asserting of which event to use.

PR #2480, opened 2018-02-10
dynamicBullets: fix for when user jumps to a more distant slide + fix for wrong offse…

tzvikaf:
…t when there are less bullets than the dynamicMainBullets

there is a bug when user moves to a slide that isn't next to the current slide,
because the active bullets will move only one step instead of the true distance between the slides

also if there are less bullets than the number configured for the dynamicMainBullets, the calculation of the offset will be wrong because it will be as if there are more bullets and so the bullets will look skewed


nolimits4web:
Thanks

tzvikaf:
sure :), it's my responsibility

PR #2490, opened 2018-02-15
enable a11y by default

DanielRuf:
a11y should be enabled by default like in most frameworks, it should be optout.

### DIFF
--- a/src/components/a11y/a11y.js
+++ b/src/components/a11y/a11y.js
@@ -153,7 +153,7 @@ export default {
   name: 'a11y',
   params: {
     a11y: {
-      enabled: false,
+      enabled: true,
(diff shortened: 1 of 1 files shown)


nolimits4web:
Well, maybe this is a good to enable by default, but it will be a breaking change, as i already see how many issue it will raise with update with a11y enabled

callumacrae:
It's better to have some users with screen readers have localisation issues than to have all users with screen readers hear nothing at all, in my opinion.

There are some things that could be improved without adding any text though, such as turning the navigation circles into links or buttons.

DanielRuf:
> It's better to have some users with screen readers have localisation issues than to have all users with screen readers hear nothing at all, in my opinion.

Definitely.

PR #2503, opened 2018-02-23
Fixes a markdown issue with v3 API docs

christiannaths:
Due to extra newlines, the table format of the v3 API document was formatted incorrectly. This change fixes that.


nolimits4web:
Thanks

Issue #2513, opened 2018-03-04
Vertical Swiper blocks natural flow of the page scroll on mousewheel.


shashank1010:
This is a (multiple allowed):
[x] bug
[ ] enhancement
[x] feature-discussion (RFC)

Swiper Version: 4.1.6
Platform/Target and Browser Versions: CHROME for MacOS, Must be similar for other Platforms as well, haven't tested.
JSFiddle https://jsfiddle.net/rqvuf8ex/10/

### What you did
I have a long page, with the Swiper Slider in one of the sections in the middle. I want the slider to be a full width, full page setup with vertical scrolling.

### Expected Behavior
Note: I have releaseOnEdges set to true. As I start a scroll/ mousewheel event on a section of the page where swiper is not enabled, the page should seamlessly flow over through the swiper to the other sections of the page. Swiper should handle mousewheel event, only if the event happened on the container element, which is by default .swiper-container. Swiper shouldn't slide to the second slide as soon as the cursor hits a Swiper element while blocking natural page flow.

### Actual Behavior
It blocks the flow of the document irrespective of where the scroll event was initialized. Also, the next slide event is immediately triggered.

shashank1010:
Closed with #2514

PR #2514, opened 2018-03-04
Handle mouseenter and mouseleave events

shashank1010:
Fix Issue #2513 I raised earlier today. 

Added mouseenter and mouseleave handlers to set a flag which conditionally lets the mousewheel handler complete. If the scroll/ mousewheel event is triggered while the mouse is on an element that is not a Swiper element, the event will not propagate.

### DIFF
--- a/src/components/mousewheel/mousewheel.js
+++ b/src/components/mousewheel/mousewheel.js
@@ -99,19 +99,17 @@ const Mousewheel = {
   },
   handleMouseEnter() {
     const swiper = this;
-    swiper.mouseEntered = true
+    swiper.mouseEntered = true;
   },
   handleMouseLeave() {
     const swiper = this;
-    swiper.mouseEntered = false
+    swiper.mouseEntered = false;
   },
   handle(event) {
     let e = event;
     const swiper = this;
 
-    if(!swiper.mouseEntered) {
-      return;
-    }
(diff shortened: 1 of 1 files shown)


nolimits4web:
Didn't get the point of the fix and issue because mousewheel event is already assigned to swiper container and not handled if scroll happens outside. Can you explain?

shashank1010:
Hi, I have been using Chrome for MacOS. I have mentioned the details in the issue I had reported.

I just tried on Safari, it seems to work just fine without the fix. I suppose it's a very specific use case to Chrome.
I have a vertical page, with a full page vertical swiper, with releaseOnEdges enabled.
The swiper is not on the first fold of the page. As I scroll through the page, following things happen.

- On Safari, swiper works perfectly fine.
- On Chrome for Mac, swiper prevents the page from scrolling and an immediate next slide event is triggered.
- On Chrome for Windows, swiper pauses the page scroll for a fraction of a second, but the next slide event is triggered.

My fix, I now realise doesn't account for the releaseOnEdges scenario. This can be handled in my next commit. However, with that fix in place, swiper works without any janky scroll on Chrome on Mac and also on Windows.

To test, here is the link to the JSFiddle I had created. https://jsfiddle.net/rqvuf8ex/10/

nolimits4web:
Thanks

PR #2523, opened 2018-03-12
Snap to nearest snap point in slide reset

### LINKED ISSUE #2485: Snapping in freeModeSticky with slide groups prefers previous slide
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: 3.4.2 and 4.1.5
* Platform/Target and Browser Versions: All
* Live Link or JSFiddle/Codepen or website with isssue: https://codepen.io/lohfu/pen/oEWOrq.

### What you did
See the linked codepen

### Expected Behavior
When using slide groups, I would expect Swiper to snap to the next group if i have swiped more than half of the current group (even when not passing minimum velocity).

If using the mousewheel, i expect to stay at the end if i have scrolled to the end.

### Actual Behavior
When swiping or scrolling you have to scroll over 80% of a slide group to go to the next one. Ie, in the codepen i linked i have to swiper past slide 4 to go the group that starts with slide 6.

If using the scroll wheel, it is not possible to get the swiper to stay in the last swipe group. If i scroll to the end, the swiper will snap back to the last full group when scrolling stops.

Is this the intended behaviour or am I missing an option to change the behaviour?

lohfu:
Currently the swiper will "reset" to the current snap point, even if the next snap point is actually closer. This pull request checks if the current or next snap point is closer to the current translation, and afterwards snaps to the one that is closer.

Fix #2485

### DIFF
--- a/src/components/core/slide/slideReset.js
+++ b/src/components/core/slide/slideReset.js
@@ -1,5 +1,19 @@
 /* eslint no-unused-vars: "off" */
 export default function (speed = this.params.speed, runCallbacks = true, internal) {
   const swiper = this;
-  return swiper.slideTo(swiper.activeIndex, speed, runCallbacks, internal);
+  let index = swiper.activeIndex;
+  const snapIndex = Math.floor(index / swiper.params.slidesPerGroup);
+
+  if (snapIndex < swiper.snapGrid.length - 1) {
+    const translate = swiper.rtl ? swiper.translate : -swiper.translate;
+
+    const currentSnap = swiper.snapGrid[snapIndex];
+    const nextSnap = swiper.snapGrid[snapIndex + 1];
+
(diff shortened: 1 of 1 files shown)


nolimits4web:
But slideReset means reset to the currently active slide, so it can be breaking change. I would prefer to introduce new method like `slideResetToClosest` or something like this, that can be used in free mode reset

nolimits4web:
Ok, merged manually into new `slideToClosest` method. Thanks!

lohfu:
@nolimits4web fair enough, but the slideReset method seems to be primarily used for freeModeSticky. The only other time it is used is when releasing a the scrollBar, where imo it also makes sense to snap to the nearest snap point.

lohfu:
@nolimits4web ok, nice. but will it be called inside the drag end handlers or do i need to call the new method manually somehow?

Issue #2539, opened 2018-03-19
VirtualTranslate and autoplay still not working together


alien-house:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 4.2.0

* Platform/Target and Browser Versions: Mac google chrome 65.0.3325.162

### What you did

Setup Swiper with loop, autoplay, and virtualTranslate(true) together.
And I added my own transition animation using CSS by myself.

```
var swiper = new Swiper({
            el: '.swiper-container',
            loop: true,
            autoplay: {
                delay:5000,
                disableOnInteraction:false
            },
            virtualTranslate:true
        });
```

There is the same issue [here](https://github.com/nolimits4web/swiper/issues/2400) and already closed but it seems not fixed yet.

### Expected Behavior

After each delay time, slideshow goes around. 
And never stop.

### Actual Behavior

After one slide ended, it's stopped.

nolimits4web:
You need to set `waitsForTransition: false` autoplay option

PR #2540, opened 2018-03-19
[issue #2539]fix VirtualTranslate and autoplay together

alien-house:
I just add a issue that is "VirtualTranslate and autoplay still not working together".

[issue #2539](https://github.com/nolimits4web/swiper/issues/2539)
- Setup Swiper with loop, autoplay, and virtualTranslate(true) together. But it doesn't work. 

In terms of fixing this issue, I changed some below:

/src/components/autoplay/autoplay.js
In "pause" functions, I separated these execute if virtualTranslate is on.
```
if (!swiper.params.virtualTranslate) {
    swiper.$wrapperEl.transitionEnd(() => {
        if (!swiper || swiper.destroyed) return;
        swiper.autoplay.paused = false;
        if (!swiper.autoplay.running) {
        swiper.autoplay.stop();
        } else {
        swiper.autoplay.run();
        }
    });
} else {
    if (!swiper.autoplay.running) {
        swiper.autoplay.stop();
    } else {
        swiper.autoplay.run();
    }
}
```

/src/components/core/slide/slideNext.js
/src/components/core/slide/slidePrev.js
In next and prev slide functions,
if virtualTranslate is on, the paused value is gonna be forced assigned false.
```
if (!swiper.params.virtualTranslate) {
    if (animating) { return false; }
} else {
    swiper.autoplay.paused = false;
}
```


nolimits4web:
There is an autoplay option `waitsForTransition` that should be set to `false` in order to work with virtualTranslate.

Issue #2210, opened 2017-09-21
Freemode + Mousewheel - breaks loop functionality and smooth scrolling


nailsiddeeq:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 3.4.2
* Platform/Target and Browser Versions:  Windows

`var mySwiper3 = new Swiper ('#column-1', {
		    	// Optional parameters
			    direction: 'vertical',
			    loop: true,
			    speed:500,
			    spaceBetween: 20,
        		   centeredSlides: true,
        		   mousewheelControl: true,
        		   slidesPerView: 'auto',
        		   freeMode: true,
        		   freeModeMomentum: true,
        		   mousewheelSensitivity: 0.02,
        		   simulateTouch: true
		  	}); 

		  	var mySwiper4 = new Swiper ('#column-2', {
		    	// Optional parameters
			    direction: 'vertical',
			    loop: true,
			    speed:1000,
			    spaceBetween: 20,
			    centeredSlides: true,
        		    slidesPerView: 'auto',
        		   freeMode: true,
        		   freeModeMomentum: true,
        		   mousewheelSensitivity: 0.02,
        		   simulateTouch: true
		  	}); 

		  	mySwiper3.params.control = mySwiper4;
		  	mySwiper4.params.control = mySwiper3;`

[see here](https://jsfiddle.net/7t4s1hz4/l)

It works perfectly when I drag with mouse or swipe; however, using the mouse-wheel with free-mode breaks the loop functionality entirely and the scrolling is very choppy. I've tested this in FF and Chrome on Windows 10 with the same results.

I've tried playing with mousewheelSensitivity - smaller values make the slides move in smaller increments, as such its less choppy but progression is too slow; however, that seems to fix the loop functionality. If it were possible to somehow accumulate some inertia after each mouse-wheel trigger that would pretty much do the trick I think.

What can I do to achieve the same experience as swiping and dragging with mousewheel as well?

Thanks!

StickNitro:
I got the same issue, is there any update on this?

My issue is with Mouse Wheel and Touch Swipe, my config is below:

swiperConfig: SwiperConfigInterface = {
        direction: "vertical",
        slidesPerView: 5,
        centeredSlides: true,
        loop: true,
    keyboardControl: true,
    mousewheelControl: true,
    mousewheelForceToAxis: true,
    mousewheelReleaseOnEdges: false,
[...]

PR #2543, opened 2018-03-21
Fix freeMode + mousewheel + loop

elaman:
With `freeMode` and `loop` are enabled, `mousewheel` component doesn't respect those parameters. This will fix it, by using the `slideTo` method, to move back/forward from **duplicate** slide to **real** one.

Note: Changes are made to the latest commit. `npm run build:prod` had been run after changing `mousewheel` component file.

Pull request must fix this issue #2210.


nolimits4web:
Hi, thanks. But changes to `dist/` files shouldn't be submitted. Also PR must pass lint tests https://travis-ci.org/nolimits4web/swiper/builds/356416021?utm_source=github_status&utm_medium=notification

nolimits4web:
Ok, better now, thanks. But from what i see and quick tests it doesn't fix the problem. Evenmore with `slidesPerView: 1` mousewheel doesn't work at all.

nolimits4web:
There is a more simple fix for this by calling just `fixLoop()` for the free mode

elaman:
Thanks for `fixLoop()`

PR #2546, opened 2018-03-25
New Architecture document

PythonLinks:
The swiper documentation is very good at defining attributes and functions.  But it does not tell much about the architecture.  One is not able to reason about Swiper.  So I decided to write this architecture document.


nolimits4web:
Thanks!

PR #2548, opened 2018-03-26
checkOverflow: click problem && add lock events

fenyagg:
some click events did not work with checkOverflow=true


nolimits4web:
Looks good, thanks

PR #2583, opened 2018-04-12
Prevent apply grab cursor when swiper is locked

danijelGombac:
Prevent apply grab cursor when swiper is locked.


nolimits4web:
Thanks

PR #2601, opened 2018-04-25
Swiper.destroy - fix for multi invocation of detachEvents() function

### LINKED ISSUE #2600: BUG: Swiper.destroy (Path included)
I found a bug with can affect multiple version:

```
Uncaught TypeError: Cannot read property 'touchEventsTarget' of undefined
    at Swiper.detachEvents (swiper.js:3095)
    at Swiper.destroy (swiper.js:3814)
    at deleteSwiper (make-swiper.js:25)
    at switchClasses (make-swiper.js:36)
    at make-swiper.js:44
```

> make-swiper.js
```
require(['jquery', 'swiper', 'domReady!'], function($, Swiper) {
    let resizeTimer = setTimeout('',500);
    let swiper;

    function isMobile() {
        if ( window.innerWidth > 1023) {
            return false;
        } else {
            return true;
        }
    }

    function makeSwiper() {
        $("[data-make-swiper-container]").addClass('swiper-container');
        $("[data-make-swiper-wrapper]").addClass('swiper-wrapper');
        $("[data-make-swiper-slide]").addClass('swiper-slide');

        swiper = new Swiper( $("[data-make-swiper-container]") ,{
            speed: 400,
            spaceBetween: 3,
            slidesPerView: 'auto',
            watchSlidesVisibility: true
        });
    }

    function deleteSwiper() {
        if ( swiper !== undefined && swiper !== '' ) {
            swiper.destroy(true, true);
            swiper = undefined; // THIS WAS NEDEED
        }
        $("[data-make-swiper-container]").removeClass('swiper-container');
        $("[data-make-swiper-wrapper]").removeClass('swiper-wrapper');
        $("[data-make-swiper-slide]").removeClass('swiper-slide');
    }

    function switchClasses() {
        if ( isMobile() ) {
            makeSwiper();
        } else {
            deleteSwiper();
        }
    }

    $(document).ready(function () {
        switchClasses();

        $(window).on('resize', function () {
            clearTimeout(resizeTimer);
            resizeTimer = setTimeout(function() {
                switchClasses();
            }, 500);
        });
    });
});

```

File with need path:
> swiper/src/components/core/events/index.js

Temporary path: Added "swiper = undefined;" after destroing of method. 

What's wrong: It show bug instead of quiet ommit additional step - additional check require to work correcly
Solution: 

In file *swiper/src/components/core/events/index.js* on line *65* add:

```
if(typeof params === 'undefined'){
 return true
};
```

Fixes all problems with destroy method in async mode and timeout.

Similar: https://github.com/nolimits4web/Swiper/issues/1367
----
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: (from lock file)
swiper@^4.1.6:
  version "4.1.6"
  resolved "https://registry.yarnpkg.com/swiper/-/swiper-4.1.6.tgz#8e426a741f211a38a3713c8fa0d5c0152c8a6a87"
  dependencies:
    dom7 "^2.0.2"
    ssr-window "^1.0.0"

* Platform/Target and Browser Versions: ALL

Triloworld:
As described in #2600
Closes #2600


nolimits4web:
Ok, but i think the logic is a bit hacky. How did you come up that you have undefined swiper.params? Don't you try to destroy already destroyed swiper?

nolimits4web:
So maybe more correct to check `if (swiper.destroyed) { return null }` instead?

Triloworld:
Null value in return is added. 
You have right :) 

swiper.destroyed isn't set before because it need to run just one time. We don't have something to test instance of swiper to be correctly defined and filled with data. 

I found this bug because "setTimeout" create new thread every time. This new threads (because I create many - not so optimize code, by the way) was overlapping and there one thread go faster and run same function with same instance. We need one variable with will check just that, then it won't be so hacky.

PR #2633, opened 2018-05-17
Fix #2632

### LINKED ISSUE #2632: iOSEdgeSwipeDetection do not work
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: 29acf89e7ed41414f1a00785615e82a388b97d53
* Platform/Target and Browser Versions: IOS
* Live Link or JSFiddle/Codepen or website with isssue: it dosen't work

### What you did
I think it's a wrong conditions in this line. It's always false , i think.
https://github.com/nolimits4web/swiper/blob/62a2e0722190ce69f6a0c7c67097bac156663217/src/components/core/events/onTouchStart.js#L33-L41

### Expected Behavior
Right

### Actual Behavior
Wrong

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](http://stackoverflow.com/questions/tagged/swiper)
for that.

Before you open an issue, please check if a similar issue already exists or has been closed before.

langjun:
judge the right or left edge

This is a right conditions,  i think.

### DIFF
--- a/src/components/core/events/onTouchStart.js
+++ b/src/components/core/events/onTouchStart.js
@@ -34,8 +34,8 @@ export default function (event) {
     Device.ios &&
     !Device.cordova &&
     params.iOSEdgeSwipeDetection &&
-    (startX <= params.iOSEdgeSwipeThreshold) &&
-    (startX >= window.screen.width - params.iOSEdgeSwipeThreshold)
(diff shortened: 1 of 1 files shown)


nolimits4web:
Thanks, merged 👍

Issue #2556, opened 2018-03-29
Add addSlide(index, '<slide>') support


kochizufan:
This is a (multiple allowed):
* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: I'm using 4.2.0.
* Platform/Target and Browser Versions: all.
* Live Link or JSFiddle/Codepen or website with isssue: none.

### What you did
There are appendSlide(slide), prependSlide(slide) and removeSlide(index) methods for dynamic slide functions.

### Expected Behavior
I want to have addSlide(index, slide) methods in addition.

### Actual Behavior
No such method now.

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](http://stackoverflow.com/questions/tagged/swiper)
for that.

Before you open an issue, please check if a similar issue already exists or has been closed before.

nolimits4web:
Fixed by #2648

PR #2648, opened 2018-05-30
Add index

kochizufan:
This PR implements issue #2556 (addSlide API).
And also, I found bug of removeSlide method around handling activeIndex with loop:true option.
The bug was also fixed by this PR.

Test is here: https://jsfiddle.net/kochizufan/aou1ajsq/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ custom
 *.log
 
 .versions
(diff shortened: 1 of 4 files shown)


nolimits4web:
Looks good, thanks! 👍

Issue #2647, opened 2018-05-30
Moving behavior works wrong and some slide not shown with loop:true


kochizufan:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 4.3.0.
* Platform/Target and Browser Versions: I tested with Chrome 66 on Windows
* Live Link or JSFiddle/Codepen or website with isssue: https://jsfiddle.net/kochizufan/keLvga7u/

### What you did
Please see attached jsFiddle.
I use swiper with loop:true, then some behavior works wrong.
In 10 slides and 2.5 slidesPerView case,
* Left arrow button stops between Slide 9 and Slide 8
* 2 swipes are need to move from Slide 9 to Slide 8
* Slide 8 is often not shown in left of Slide 9

### Expected Behavior
Moving continuously and showing every slide correctly.

### Actual Behavior
As written above.

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](http://stackoverflow.com/questions/tagged/swiper)
for that.

Before you open an issue, please check if a similar issue already exists or has been closed before.

nolimits4web:
Fixed by #2651

PR #2651, opened 2018-05-31
Fix loop behavior (Issue #2647)

kochizufan:
Fixed loop behavior, which reported in issue #2647.

In 10 slides and 2.5 slidesPerView case,

* Left arrow button stops between Slide 9 and Slide 8 => Fixed
* 2 swipes are need to move from Slide 9 to Slide 8 => Fixed
* Slide 8 is often not shown in left of Slide 9 => Fixed

Test is at https://jsfiddle.net/kochizufan/su1ka5ov/


nolimits4web:
Awesome, thanks 👍

PR #2659, opened 2018-06-04
:bug: Fix Bug: swiper.stopAutoplay() => swiper.autoplay.stop()

Brightcells:
``swiper.stopAutoplay()`` should be replaced by ``swiper.autoplay.stop()``


nolimits4web:
Thanks!

PR #2689, opened 2018-06-24
Fix lock-threads config

dessant:
Spotted with Sentry.

```js
Error: ValidationError: "daysUntilStale" is not allowed. "daysUntilClose" is not allowed
```


nolimits4web:
Thanks

PR #2728, opened 2018-07-27
Fix typo in passive listeners.

nyon:
This pull request fixes a typo in `scrollbar.js` which prevents using the param `passiveListeners` in this specific module.


nyon:
The travis build fails because there is an unused variable in a file I never touched. However, should I fix that problem in this branch?

nolimits4web:
All good, thanks!

Issue #2730, opened 2018-07-28
Performance Issues using observer option


rayvincent-bsd:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 4.3.3.
* Platform/Target and Browser Versions: All
* Live Link or JSFiddle/Codepen or website with isssue:
  * https://codepen.io/rayvincent-bsd/pen/mjqOVZ
    * With codepen editor
  * https://s.codepen.io/rayvincent-bsd/debug/mjqOVZ/wQrPoNZOqgNM
    * Without codepen editor overhead

### What you did
Enabled `observer` option and added 500 slides dynamically

### Expected Behavior
Upon the next mutation event, the swiper should process the update efficiently without performance issues.

### Actual Behavior
Swiper will execute 'onResize' event handler over 500 times, once for each mutation that occurred in the mutation event.  This is causing drastic performance issues.

With the given codepen link, you can see that the it takes about 15 seconds with the attached Chrome profile. Tested in Chrome v67.0.3396.99 (Mac OSX)

- [codepen.profile.gz](https://github.com/nolimits4web/swiper/files/2238018/codepen.profile.gz)

nolimits4web:
Fixed by #2731

PR #2731, opened 2018-07-28
#2730: Ensure observerUpdate event is only dispatched once on a MutationEvent

rayvincent-bsd:
I found that the `observerUpdate` event is being dispatched for every mutation in a MutationEvent.

This causes serious performance issues depending on the number of mutations that are present in the MutationEvent.  I saw that all the methods that react to the `observerUpdate` will safely accommodate this change.  Calling `observerUpdate` multiple times will have the same end result as dispatching it once within a callstack.

And to make things run just a little smoother, I decided to add the ability to run this within an animation frame to help keep the ui from locking up.


nolimits4web:
Great, thanks! 👍

PR #2746, opened 2018-08-09
Fix touch swiping in UWP app on surface tablets

ph1p:
I want to use **swiper** in a Universal Windows Platform App (UWP) on a surface tablet.
The swiper does not work with touch events on this device, because UWP Apps on Windows 10 are using the Edge Browser.

It seems like the Edge Browser needs the `WP8 Touch Events Fix` in ` /src/components/core/classes/addClasses.js` too.
I've added the `isEdge` property to not confuse anyone.
Because it's not the Internet Explorer, but it behaves like that.

### DIFF
--- a/src/components/core/classes/addClasses.js
+++ b/src/components/core/classes/addClasses.js
@@ -33,7 +33,7 @@ export default function () {
     suffixes.push('ios');
   }
   // WP8 Touch Events Fix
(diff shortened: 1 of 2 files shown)


nolimits4web:
Looks good! But can you please remove commits from `/dist/` files? Only `/src/` files should be affected

ph1p:
Cool :) I've updated my commit.

nolimits4web:
Thanks 👍

Issue #2755, opened 2018-08-17
Virtual slides ignore initialSlide in the beginning


websirnik:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 4.3.5
* Platform/Target and Browser Versions: CHROME
* Live Link or JSFiddle/Codepen or website with isssue: https://jsfiddle.net/bqv2137n/11/

### What you did
When using `initialSlide` setting together with `virtual`, there is extra computing happening. Despite the value of `initialSlide`, Swiper currently always calls `renderSlide` with `0`&`1` indexes before rendering `initialSlide` slide. This is costly, when my `createSlide()` function is heavy. 
```
var mySwiper = new Swiper('.swiper-container', {
    initialSlide: 4,
    virtual: {
        slides: ['Slide 1', 'Slide 2', 'Slide 3', 'Slide 4', 'Slide 5', 'Slide 6'],
        renderSlide: function(slide, index){
            console.log(slide);
            return createSlide(slide);
        }
    }
});
```

### Expected Behavior
Console Output: 
```
Slide 4
Slide 5
Slide 6
```

### Actual Behavior
Console Output: 
```
Slide 0
Slide 1
Slide 4
Slide 5
Slide 6
```

nolimits4web:
Fixed by #2756

PR #2756, opened 2018-08-17
Remove swiper.virtual.update() from beforeInit callback

websirnik:
This is a fix for: https://github.com/nolimits4web/swiper/issues/2755

Before, on `beforeInit` event, `swiper.virtual.update()` being executed when value for `swiper.activeIndex` was not set. That led to Swiper rendering `0`&`1` slides, even if the `initialSlide` was set to something else.

`virtual.update()` is not nessesary in the `beforeInit` callback, since its being called in the `setTranslate` callback and by that time `swiper.activeIndex` was initialised with the right  `initialSlide` value.


nolimits4web:
Thanks! 👍

PR #2758, opened 2018-08-22
Fixed handler unsubscription error on multiple create/destroy events …

glebmachine:
In case of empty listeners `self.eventsListeners[event]` this code brokes `self.eventsListeners[event].forEach()`


nolimits4web:
Thanks! 👍

PR #2774, opened 2018-08-30
normally calculate break points;

PushkarevMax:
If during the development of the project the principle mobile first is used, it is not possible to normally calculate break points. 
I suggest calculating points to do the opposite for convenience.

For example:
```
breakpoints: {
            // when window width is >= 320px
            320: {
                slidesPerView: 2
            },
            // when window width is >= 768px
            768: {
                slidesPerView: 3
            },
            // when window width is >= 1170px
            1170: {
                slidesPerView: 4
            }
        }
```

### DIFF
--- a/src/components/core/breakpoints/getBreakpoint.js
+++ b/src/components/core/breakpoints/getBreakpoint.js
@@ -11,7 +11,7 @@ export default function (breakpoints) {
   points.sort((a, b) => parseInt(a, 10) - parseInt(b, 10));
   for (let i = 0; i < points.length; i += 1) {
     const point = points[i];
-    if (point >= window.innerWidth && !breakpoint) {
+    if (point <= window.innerWidth) {
       breakpoint = point;
     }
   }


nolimits4web:
Sorry, it can be merged and changed this way, because it will be a breaking change

PushkarevMax:
What if i add some flag for example 'mobileFirst' and if it true, then i can chenge calculating?

nolimits4web:
@PushkarevMax but what it adds? It doesn't really change or add anything, it is just another way of doing same

PushkarevMax:
by default this flag will be false and accordingly nothing will break in previous versions. and the one who wants to use breakpoints calculation if he uses the mobile first design principle, he will simply include this flag and can normally calculate breakpoints.
If i am not right, i apologize, but it would be convenient calculate breakpoints in rhis way. 
In any case, your product is great! :)

nolimits4web:
Ok, let's add something like `breakpointsInverse: false` default parameter and use your logic

PushkarevMax:
Thank you)

nolimits4web:
Will you send a fix with additional parameter or you want me to add it?

PushkarevMax:
I send

nolimits4web:
Merged manually, thanks

Issue #2798, opened 2018-09-14
Can't change the slide by dragging on a link(image)


phx-vic:
Swipper 4.4.0. MacOS. 
When change the slide by dragging the mouse on a link or image, the native drag is turned on.

https://codepen.io/phx-vic/pen/RYBpOx

nolimits4web:
Can't see such issue, on horizontal swipes/drag everything is correct

nolimits4web:
Fixed by #2799

nolimits4web:
Fixed in 4.4.1

PR #2799, opened 2018-09-14
Hot fix for touchStart prevent default issue

exodusanto:
Hot fix #2794 

I saw `swiper.allowTouchMove` and I thought that `swiper` has inside all params but you extends swiper instance and add `allowTouchMove` property on another file 😖.

I fix it with the correct `params` object


exodusanto:
Related #2798

nolimits4web:
Ah, that is why :) Thanks!

PR #2842, opened 2018-10-19
Scrollbar events based only on Support.touch

### LINKED ISSUE #2840: Setting simulateTouch to false breaks scroll bar in Safari
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: 4.4.1
* Platform/Target and Browser Versions: macOS, Safari
* Live Link or JSFiddle/Codepen or website with isssue:  https://jsfiddle.net/pcwudrmc/61182/

### What you did
Enable a scroll bar with the default settings.  Set simulateTouch to false.  This stops the use of the scroll bar in Safari only.  Still works in Chrome.

```
var mySwiper = new Swiper ('.swiper-container', {
  direction: 'horizontal',
  loop: true,
  scrollbar: {
    el: '.swiper-scrollbar',
    hide: false,
    draggable: true,
    snapOnRelease: true
  },
  simulateTouch: false  // Breaks scroll bar in safari
});
```

### Expected Behavior
I would expect the scroll bar to still work.

### Actual Behavior
Scroll bar won't drag

hanoii:
Fixes #2840

Further thinking on what's on the issue, I think the logic deciding what events to attach (at least for the scrollbar) can be simplified in just weather touch is enabled or not. The touchEvents() function takes care of which desktop events to use.

This may be applied to the main part where a similar decision making is made in this regards: 

https://github.com/nolimits4web/swiper/blob/aef21b8e6add21c86cc8681be83ac4647dc23c7a/src/components/core/events/index.js#L30-L45

https://github.com/nolimits4web/swiper/blob/aef21b8e6add21c86cc8681be83ac4647dc23c7a/src/components/core/events/index.js#L71-L86

I can submit a different PR for that if OK

### DIFF
--- a/src/components/scrollbar/scrollbar.js
+++ b/src/components/scrollbar/scrollbar.js
@@ -2,7 +2,6 @@ import { document } from 'ssr-window';
 import $ from '../../utils/dom';
 import Utils from '../../utils/utils';
 import Support from '../../utils/support';
-import Device from '../../utils/device';
 
 const Scrollbar = {
   setTranslate() {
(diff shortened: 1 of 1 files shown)


nolimits4web:
Looks good, thanks! I think we can reuse same logic in main events, but anyway we need `simulateTouch` related logic there

PR #2874, opened 2018-11-13
Fix incorrect calculation of the number of full columns.

nukefusion:
In grid mode, the number of full columns is incorrectly calculated in updateSlides.js causing data-row-swiper to be incorrectly calculated and subsequently incorrect margins to be applied to the slides.

For example, a 8 x 3 grid (slidersPerView=3 and slidesPerColumn=8) yields 24 slides. If there are 17 slides in the collection then there are 2 full columns.

Existing logic calculated this as 3 - ((8 * 3) - 17) = -4. Clearly there cannot be -4 complete columns.

Correct logic should yield the integer component of (17 / 8) which is 2.


nolimits4web:
Don't remember already why i chose so weird formula for this, but this one makes sense. Will hope it won't break anything

PR #2888, opened 2018-11-20
fix wrong choose for module

Qymh:
webpack default options for `resolve.mainFields` is module.In everyday habits we ignore `node_modules` in babel plugins.so if we ues es6 it will have some mistakes in the browser which is lower version


nolimits4web:
Everything is correct, it should bundle module by default

PR #2939, opened 2019-01-08
Improve the Slideable Navigation Drawer Demo

kireerik:
resolves issues with the implementation after Swiper `4.x` update


nolimits4web:
Thanks!

Issue #308, opened 2013-08-13
Change mode (horizontal/vertical) on the fly


Xochitl:
Hi,

I am trying to use Swiper and I would like to modify the mode (horizontal/vertical) when device orientation changes.
I tried mySwiper.params.mode = 'vertical' / 'horizontal', and it seems that the value is actually modified, but the swiper keeps swiping in the same direction.

Is there a way to do this ?

PR #2964, opened 2019-02-01
fix bug horizontal to vertical and vertical to horizontal direction

Enigama:
Added additional checks in the onResize and updateSize functions to ensure correct operation of the plug-in when changing the width of the document and changing the direction.

https://github.com/nolimits4web/swiper/issues/308


nolimits4web:
Thanks! But can you please revert changes in anything except in `src/` folder files?

Enigama:
Ok, i'm try)

Issue #2159, opened 2017-07-21
Swiper SASS styling


Giovanni-Mattucci:
This is a:
* [x] enhancement

It would be great if Swiper also had sass in addition to less styling. If (like me) you develop your projects using sass, being able to import and modify sass variables before being processed would be a nice addition :)

camdagr8:
I used this tool when I needed to set it up as a sass file. It's not perfect, but it at least gave me a starting point to then variablize my styles. 

http://css2sass.herokuapp.com/

landsman:
yes please! :)

asso1985:
It would be great

mescie:
+1 would be great to have SASS

James-Bovis:
+1 I would love to see a SASS port too. I'm facing the same problem as @goranNetgen ]

paolocavanna:
+1 it'd be absoluteley a useful feature to be added

dangelion:
+1 please

goranNetgen:
Guess this won't be a problem anymore soon since LibSass decided to undeprecate CSS imports (yay) https://github.com/sass/libsass/pull/2684
It would still be a nice feature.

dangelion:
Actually on [Hugo](https://gohugo.io/hugo-pipes/scss-sass/) I get error

`error: failed to transform resource: SCSS processing failed: file "stdin", line 57, col 1: File to import not found or unreadable: ../../../../node_modules/swiper/dist/css/swiper.`

because of importing css from an scss is not good

PR #2973, opened 2019-02-09
Add Sass Styling

polarfawx:
I made a Sass port of the original Less styles as requested in issue #2159.  Could you include the `.scss` files into the source tree, or it would be better if I create a separate package for these styles?


nolimits4web:
Thanks! 👍

Issue #2975, opened 2019-02-12
Add support for appending & prepending array of slides to Virtual API (RFC)


garrettmaring:
**Abstract**

The Virtual API should support the same appending & prepending operations as it's non-virtual counterpart.

* [ ]  `prependSlide` should support taking an array or a single slide
* [ ]  `appendSlide` should support taking an array or a single slide

The virtual cache should update appropriately as well.

PR #2976, opened 2019-02-12
Add support for appending & prepending array of...

garrettmaring:
...slides to Virtual API.

Fixes [#2975](https://github.com/nolimits4web/swiper/issues/2975)

### DIFF
--- a/demos/410-virtual-slides.html
+++ b/demos/410-virtual-slides.html
@@ -69,15 +69,19 @@
     <div class="swiper-button-prev"></div>
   </div>
   <p class="append-buttons">
+    <a href="#" class="prepend-2-slides">Prepend 2 Slides</a>
     <a href="#" class="slide-1">Slide 1</a>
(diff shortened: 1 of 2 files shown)


garrettmaring:
@nolimits4web I attempted to parody the code in `core/manipulation/appendSlide.js` and `core/manipulation/prependSlide.js`.

When looking at those functions versus their Virtual counterparts, I noticed that the `runCallbacks` parameter is false in the core append/prepend yet true (default) in the Virtual append/prepend. I'm curious to know why that is the case or if it was an accidental oversight.

nolimits4web:
Thanks, merged!

> When looking at those functions versus their Virtual counterparts, I noticed that the `runCallbacks` parameter is false in the core append/prepend yet true (default) in the Virtual append/prepend. I'm curious to know why that is the case or if it was an accidental oversight.

Can't remember for sure right the reason now, but it was for sure required. I remember there were some use cases when Virtual slides won't work correctly if no callbacks passed there

PR #2982, opened 2019-02-16
func onResize, func updateSlides fix for direction

### LINKED ISSUE #308: Change mode (horizontal/vertical) on the fly

Enigama:
direction horizontal to vertical, vertical to horizontal when resize window
Fixes #308


nolimits4web:
Hi, thanks for your PR. But again, it affect files not in `src/` folder :) So i merged it manually with some restructuring. I am afraid it doesn't cover many use cases, but for basics it is fine.

GlebGulin:
Sorry, but I have question about mobile adaptation fow swiper. So, in version 3.4.1 when width of windows is 485px, we have two columns. If these colums with for example Testimonials, then we have so long text. It's not beautifully. When width is 320px, we have one columns. I look at code, but algorithm so dificult. How can I do two columns for width 485px. Thank you for answer.

Issue #2990, opened 2019-02-21
Errors on use thumbnail (change, click)


MartDSam:
I have 2 sliders, 1 for the thumbnails and the other for the larger images.

When I click on the thumbnail to change the larger image some errors are happening. Errors do not occur at the same time, sometimes one, sometimes another.

1 - I click on the thumbnail and this error happens:
Cannot read property 'indexOf' of undefined

In addition, the thumbnail is "glued" to the mouse pointer, with and if you are dragging with your finger on the phone.

2 - I click on a thumbnail, but it does not become active. The active class continues in the previously clicked thumbnail.

3 - Sometimes the swiper-slide-thumb-active class changes to the active slide, but the swiper-slide-active class remains on the previous slide.

4 - The clicks are not fluent, I often click on the thumbnail and nothing happens, it seems that it does not identify the click. Sometimes I click more than once or twice for him to identify.

5 - It often looks like the thumbnail only changes if I double-click.

Here demo: [Code pen demo](https://codepen.io/Haisan/pen/wNbNQm)

efor74:
Решил попробовать перейти на этот слайдер, раньше использовал Slick, но на первом же проекте возникла та же проблема :(
Тормознул пока этот процесс.

nolimits4web:
Fixed by #3020

PR #3020, opened 2019-03-28
Bugfix thumbnails

michaelWoe:
Fixes the bug 'Cannot read property `indexOf` of undefined' described in step 1 of #2990 that sometimes occurs on use of thumbnails


nolimits4web:
Thanks!

PR #3094, opened 2019-06-06
Don't override Dom7 methods if already declared

axten:
I'm using dom7 modular in a project together with swiper and patched some methods to be more jquery compatible (e.g. added second argument for toggleClass).

But now my methods gets randomly overidden (depending on webpacks processing order).


nolimits4web:
Looks good, thanks 👍 +1

PR #3096, opened 2019-06-08
add support lazy load for picture with source

exdeniz:
Add support lazyload for picture element with source tag.


nolimits4web:
Please, only changes to `src` files are allowed. Remove any changes to `dist` files and i will merge it

bpeab:
Is there any release date for that feature ?

exdeniz:
> Please, only changes to `src` files are allowed. Remove any changes to `dist` files and i will merge it

done

bpeab:
@exdeniz It seems that your solution also download the fallback image (the source inside the `` element) but from my understanding, it shouldn't be the case. Only the defined viewport source should be downloaded, thus only one image should be downloaded. Wdyt?

bpeab:
@exdeniz Also, it doesn't work at all with IE 11 even with Picturefull acting as a polyfill when `<picture>` isn't natively supported.

exdeniz:
> @exdeniz It seems that your solution also download the fallback image (the source inside the `` element) but from my understanding, it shouldn't be the case. Only the defined viewport source should be downloaded, thus only one image should be downloaded. Wdyt?

If use ```<source srcset="..."/>``` and `````` browser load a picture in ```<source ...>``` anyway. Need use ```<source data-srcset="..."/>``` for correct load picture. This patch fixes this. For more detail see https://github.com/verlok/lazyload

exdeniz:
> @exdeniz Also, it doesn't work at all with IE 11 even with Picturefull acting as a polyfill when `<picture>` isn't natively supported.

If you use `<picture>`, then you know about polyfill.

bpeab:
Have you testing on IE11 yourself ? It just doesn't work with Picturefill. It always base itself on the fallback image. I have lazySizes enabled on all other images and it works perfectly with Picturefill but with Swiper and the changes you've made, it calculates the height based on the fallback image.

exdeniz:
> Have you testing on IE11 yourself ? It just doesn't work with Picturefill. It always base itself on the fallback image. I have lazySizes enabled on all other images and it works perfectly with Picturefill but with Swiper and the changes you've made, it calculates the height based on the fallback image.

I have no way to test and fix this on IE11

bpeab:
Ok. I am investiguating since I really need this feature. `source` seem to be completly ignored once lazy-loaded. Can't really find why. I have some base64 png as `srcset` before making the switch with the image and the height seems right (it takes the good viewport one) but once lazy-loaded, it doesn't work at all... I'll keep searching...

exdeniz:
> Ok. I am investiguating since I really need this feature. `source` seem to be completly ignored once lazy-loaded. Can't really find why. I have some base64 png as `srcset` before making the switch with the image and the height seems right (it takes the good viewport one) but once lazy-loaded, it doesn't work at all... I'll keep searching...

Stop using IE11 =)

bpeab:
I found a work-around. It'll be in a PR later today.

> Stop using IE11 =)

That would be nice but unfortunatly not negociable 😄

bpeab:
I submitted a PR (#3151) with the necessary changes to support the `picture` element lazy-loading and the ability to use Picturefill if it is available. 

@exdeniz @nolimits4web Let me know if it's okay or if something is missing.

knynkwl:
Any idea if `<source>` will be supported?

PR #3101, opened 2019-06-16
perf(animation): improve animation perfomance

nosovk:
We can significantly improve slider performance by letting chrome little hint about our goals - making animations.
https://developer.mozilla.org/ru/docs/Web/CSS/will-change

In another PR https://github.com/zefoy/ngx-swiper-wrapper/pull/221 was suggested that here will be better place for that improvement.

### DIFF
--- a/src/components/core/core.less
+++ b/src/components/core/core.less
@@ -27,6 +27,7 @@
 }
 .swiper-container-android .swiper-slide, .swiper-wrapper {
   transform:translate3d(0px,0,0);
+  will-change: transform;
 }
 .swiper-container-multirow > .swiper-wrapper {
   flex-wrap: wrap;
--- a/src/components/core/core.scss
+++ b/src/components/core/core.scss
@@ -28,6 +28,7 @@
 .swiper-container-android .swiper-slide,
 .swiper-wrapper {
   transform:translate3d(0px,0,0);
+  will-change: transform;
 }
 .swiper-container-multirow > .swiper-wrapper {
   flex-wrap: wrap;


nolimits4web:
@nosovk would be good to see live example or JSFiddle comparing swiper without it and with `will-change` where it will clearly illustrate performance improvements. From my experience, it can (and probably will) make it even worse. Also check the notes and MDN page:

> Don't apply will-change to too many elements. The browser already tries as hard as it can to optimize everything. Some of the stronger optimizations that are likely to be tied to will-change end up using a lot of a machine’s resources, and when overused like this can cause the page to slow down or consume a lot of resources.

> Use sparingly. The normal behavior for optimizations that the browser make is to remove the optimizations as soon as it can and revert back to normal. But adding will-change directly in a stylesheet implies that the targeted elements are always a few moments away from changing and the browser will keep the optimizations for much longer time than it would have otherwise. So it is a good practice to switch will-change on and off on using script code before and after the change occurs.

> Don't apply will-change to elements to perform premature optimization. If your page is performing well, don't add the will-change property to elements just to wring out a little more speed. will-change is intended to be used as something of a last resort, in order to try to deal with existing performance problems. It should not be used to anticipate performance problems. Excessive use of will-change will result in excessive memory use and will cause more complex rendering to occur as the browser attempts to prepare for the possible change. This will lead to worse performance.

> Give it sufficient time to work. This property is intended as a method for authors to let the user-agent know about properties that are likely to change ahead of time. Then the browser can choose to apply any ahead-of-time optimizations required for the property change before the property change actually happens. So it is important to give the browser some time to actually do the optimizations. Find some way to predict at least slightly ahead of time that something will change, and set will-change then.

> Be aware, that will-change may actually influence the visual appearance of elements, when used with property values, that create a stacking context (e.g. will-change: opacity), as the stacking context is created up front.

AaronGaim:
I have a [heavy page](https://irongrif.com/workouts), where switching slides is quite common, after adding 'will-change' its improved the performance on mobile greatly. Now if I have 20 exercises + it's quite smooth, and even 40 exercise + have decent performance. so in my case, it was worth it.

nolimits4web:

nosovk:
@nolimits4web just not ready to argue. Because of significant improvement in perfomance we just  added those changes directly to projects where we use swiper. Because of css nature we can do it without modifications into the library.

About demo:
I created branch with reverted commit about will-change optimization [link](https://btw-0-swiper.stage.7bitcasino.com/en/games/allGames)
there is swiper slider in header without `will-change` on it.
On [master branch](https://master.stage.7bitcasino.com/en/games/allGames) will-animate fix is precent.
Comparing them on ios you may see that ios devices have sufficient problems in sliding big images. 

 > Don't apply will-change to elements to perform premature optimization 

If we use css transform in infinite loop (like typical slider behaviour) that wouldnt be premature optimizatiom

Issue #2978, opened 2019-02-14
IE11: Wrong (virtual) slides on init


vertic4l:
Swiper does have a bug on init within IE11. I'm trying to use the virtual slides feature but they are initially wrong.

This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 4.4.6
* Platform/Target and Browser Versions: Windows IE11.

### What you did
Setup a React Component which uses Swiper

### Expected Behavior
Swiper initializes on IE11 like it does on Chrome. Slides are correct, initialSlide gets noticed and correct slide gets displayed.

### Actual Behavior
Swiper initializes on IE11 but shown (virtual) slide is wrong. It's still the first slide, not the 8th in my case. Works in Chrome, FF, etc. like a charm.
`this.swiper.realIndex` seems to be correct because both Arrows are getting displayed, but shown slide is still the first, not the 8th. And it doesn't matter if i set it as `initialSlide` or use `slideTo()`.

Interestingly the data of `handleRenderExternal` differs from IE11 to Chrome.
In IE11 `data.from`and `data.to` is set to 0. In Chrome it's correct... `data.from`= 7, `data.to` = 9

Without `Virtual` everything works as expected. So there must be an issue within `Virtual` module.

### Side note
Transpiled with webpack + babel-loader:

```
{
    test: /\.m?js$/,
    include: [
        path.resolve(projectRoot, "node_modules/dom7"),
        path.resolve(projectRoot, "node_modules/ssr-window"),
        path.resolve(projectRoot, "node_modules/swiper")
    ],
    use: {
        loader: "babel-loader",
        options: {
            presets: ["@babel/preset-env"]
        }
    }
},
```
^ This should be placed in the documentation too by the way!

### Actual code
```
import * as React from "react";
import { findDOMNode } from "react-dom";
import autobind from "autobind-decorator";
import { Swiper, Virtual, SwiperOptions, SelectableElement } from "swiper/dist/js/swiper.esm.js";
import Icons from "Application/view/generated/Icons";

Swiper.use([Virtual]);

interface IReactSwiperProps extends SwiperOptions {
    initialSlide?: number;
    slides: Array<JSX.Element>;
    onSlideChange?: (slide: number) => void;
    withNavigation: boolean;
}

interface IReactSwiperState {
    slides: Array<React.ReactNode>;
    offset: number | null;
    isInitialized: boolean;
}

[...]

PR #3110, opened 2019-06-21
Add "browser" field to the package.json

yuezk:
When bundling the Swiper module using Webpack, the bundled js will cause a Syntax error in IE 11.

The reason is that Webpack will use the "module" field (since the default value of `mainFields` defined in Webpack is `["browser", "module", "main"]` ) as the entry point of Swiper.

As the class syntax inside the "module" entry point was not transpiled and then was bundled into the target js files, which will cause the Syntax error in IE 11, because IE 11 doesn't support the class syntax.

The relevant issue is #2978

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "4.5.0",
   "description": "Most modern mobile touch slider and framework with hardware accelerated transitions",
   "main": "dist/js/swiper.js",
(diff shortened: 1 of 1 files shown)


nolimits4web:
No, sorry, it will break many things. Enable transpiling or import UMD module directly

Issue #2013, opened 2017-02-10
Img overlap


nam-co:
This is a (multiple allowed):
* [x] bug

Hello, hope somebody can help me, I using the fade effect but Im getting this overlap of images, how can I fix this? I tried this css but it didn't work, appreciate any help

![img_9602](https://cloud.githubusercontent.com/assets/3951529/22844922/5f2d9692-efae-11e6-978f-a3d915c30c13.JPG)

.swiper-slide {
	min-height: 100%;
	background-color: white !important;
}

JacobDB:
I'm guessing that photo is supposed to rotated 90 degrees, right? If so, it's not working because each slide is only the height of it's contents. To adjust for this, add this to your CSS:

    .swiper-slide {
        flex-grow: 1;
    }

That should allow you to set a background color on `.swiper-slide` as you're trying to do.

nolimits4web:
Issue is closed because of outdated/irrelevant/not actual/needed more information/inactivity.

If this issue is still actual and reproducible for latest version of Swiper, please create new issue and fill the issue template correctly:
* Clearly describe the issue including steps to reproduce when it is a bug.
* Make sure you fill in the earliest version that you know has the issue.
* Provide live link or JSFiddle/Codepen or website with issue

ockle:
@nam-co I know this is a year old, but I had the same issue and found this bug, so in case you or anyone else who finds this wants to make this work as expected, you need to set:
```js
fadeEffect: {
  crossFade: true
}
```
(see http://idangero.us/swiper/api/#fade-effect)

PR #3113, opened 2019-06-23
Update 220-effect-fade.html

lexxyungcarter:
Added the missing piece in the documentation.
Thanks to this reply: 
https://github.com/nolimits4web/swiper/issues/2013#issuecomment-367494109


nolimits4web:
Sorry, but it is described in docs, not needed here

PR #3137, opened 2019-07-06
Make possible of work shift + mouse wheel (horizontal scroll)

crutch12:
When create swiper with horizontal scroll only (forceToAxis)
```
mousewheel: {
  forceToAxis: true,
  invert: true,
},
```
it doesn't scroll if try to scroll with `Shift + Scroll` combination

Checked in all browsers (except safari) + mobile + win touchpad, works fine


nolimits4web:
Thanks, merged

PR #3151, opened 2019-07-16
Add support for picture element lazy-loading

bpeab:
Add support for `picture` element lazy-loading and support for legacy browsers with the use of Picturefill if available.


bpeab:
I would love to fix the _linting_ problem but it seems that the entire codebase needs to be fixed. Let me know if you want me to do something about it.

DerVaddi:
Hi, 
i test your changes. It works fine for the picture element. but when you have only the image element without picture it doesn't work.

nolimits4web:
Hi, can't merge it. All changes should be done to `src` files only. Why there is `dist` changes, why there is port changed in gulpfile? No polyfills should be used

bpeab:
Thanks @DerVaddi, that's indeed the case. I changed it for the project I am working on but completely forgot to apply the changes to the PR.

@nolimits4web I'll check for the `dist` and `gulpfile` changes but the condition about Picturefill is needed in order for IE11 to work with the `picture` element. The polyfill isn't imported nor specified as a dependency. I only check if it is applied globally and therefore, refresh the element. 

Let me know what you think but that's the only way I think it'll make it work within IE11.

PR #3156, opened 2019-07-19
SlideSize to include paddings for border-box elements in IE (fixes #2897)

### LINKED ISSUE #2897: Miscalculation of Swiper parent width in IE11, unable to scroll to last slide
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: 4.4.2
* Platform/Target and Browser Versions: IE11 (11.407.17134.0) Windows 10 Pro
* Live Link or JSFiddle/Codepen or website with issue: https://codepen.io/saltymouse/full/ZmMPxW/

### What you did
In the above Codepen demo (Bootstrap layout similar to my original project), using the mouse to drag/swipe (from right to left) to go to the final slide in the grey, numbered (0,1,2...) tab-slides.

### Expected Behavior
Be able to drag/swipe all the way right to the bring the final slide into view.
All other modern browsers except IE11 perform as expected.

### Actual Behavior
IE11 stops the drag process at about slide '16 sixteen' unable to continue dragging to the final slide ('23 isEnd'). Checking at the console, IE tells me that `tabbyTop.isEnd // true` even though it has stopped midway.

Interestingly, the bottom Swiper (with photos and articles) works as expected in IE11, dragging to the last slide. The two Swipers are connected only 1-way: `tabbyBottom.controller.control = tabbyTop;`

I tried setting the parent width to 100% as suggested in #1258, but had no effect on how IE makes its calculations.

DKvistgaard:
In IE the width returned from `getPropertyValue('width')` on elements with a `box-sizing: border-box` styling doesn't include paddings as in other browsers.

This is an issue that arrises in IE when using `slidesPerView: 'auto'` and having paddings on each slide together with a `box-sizing: border-box;` styling on the slides.

Here's an example: https://s.codepen.io/DKvistgaard/debug/BXyZyx/XBrGRZbEvGGM

Also see the issue #2897 for more info.

### DIFF
--- a/src/components/core/events/onResize.js
+++ b/src/components/core/events/onResize.js
@@ -29,7 +29,6 @@ export default function () {
     if (params.autoHeight) {
       swiper.updateAutoHeight();
     }
-
   } else {
     swiper.updateSlidesClasses();
     if ((params.slidesPerView === 'auto' || params.slidesPerView > 1) && swiper.isEnd && !swiper.params.centeredSlides) {
--- a/src/components/core/update/updateSlides.js
+++ b/src/components/core/update/updateSlides.js
@@ -1,6 +1,7 @@
 import { window } from 'ssr-window';
 import Utils from '../../../utils/utils';
 import Support from '../../../utils/support';
+import Browser from '../../../utils/browser';
 
 export default function () {
   const swiper = this;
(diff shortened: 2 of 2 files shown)


justingrant:
@DKvistgaard - does this fix work in cases where border-box is applied at a higher level in the DOM and is inherited by each slide?

DKvistgaard:
> @DKvistgaard - does this fix work in cases where border-box is applied at a higher level in the DOM and is inherited by each slide?

Since it's using `getComputedStyles` then it get's all inherited styles as well. So yes, it should also fix the cases where the `box-sizing` styling is inherited.

nolimits4web:
Thanks!

PR #3185, opened 2019-08-22
fixed "swiper-slide-visible" className shows up when it shouldn't

Mrliuchanghao:
Because in the “updateSlidesOffset” method,DOM offsetLeft and offsetTop do not compute decimals,So when slide offestLeft or offsetTop is decimal,  it will be issues. 
When you need to get the visibility parameters of slide, you will find that the invisible slide also has "swiper-slide-visible" className.
I set the error range between 1PX to control the visibility of the slider.
The example is in the demos files: "demos/450-watchSlidesVisibility.html"


nolimits4web:
Looks good to me, merged, thanks!

Issue #3073, opened 2019-05-17
ESM is not trully ESM


Antonio-Laguna:
This is a (multiple allowed):
* [] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 4.5.0
* Platform/Target and Browser Versions: n/a

I had rollup set up but we're switching to Webpack that led me to a path of discovery around formats on JavaScript. Perhaps I'm wrong but what I can see on the `.esm` bundle is not trully ESM and, in fact, rollup[ is not doing](https://github.com/nolimits4web/swiper/blob/master/scripts/build-js.js#L38) `esm` but `es`. It does have untranspiled code.

Rollup does transpile node_modules whereas webpack does not and the two worlds collide. What I found is that providing a `module` on `package.json` pointing to a real `esm` bundle should be ideal. 

Right now I'm solving by doing `import 'swiper/dist/js/swiper.js';` but that breaks tree-shaking .

Perhaps I'm missing something here but would it hurt to switch `es` to `esm` on the `esm` bundles?

nolimits4web:
ESM is ES module = contains `import` and `export` statements. Transpiling to ES5 syntax is not a job of Rollup or Webpack. It is a job of libs like Babel or Buble. Closing as not related to Swiper

Antonio-Laguna:
ESM would leave imports and exports but would not use arrow functions nor let/const.

To me is counter intuitive that you're naming .esm but rollup is set up as es instead.

slorber:
@nolimits4web problem is that in web/browserland, we expect libs to already be transpiled when published on npm, so that we can just consume the lib without requiring transpiling the /node_modules content

In this case, you have ESM output with ES6 imports/exports (which is fine) but the syntax is not transpiled, so importing this makes it painful for those wanting to support IE11/Edge which says "syntax error".

PR #3203, opened 2019-09-12
Swapping main with module

### LINKED ISSUE #3199: Why keep closing un-transpilation issues?
This is a (multiple allowed):
* [ ] bug
* [X] enhancement
* [X] feature-discussion (RFC)

I made an issue long ago with ESM issues: https://github.com/nolimits4web/swiper/issues/3073 today I discover I have an issue using React Id Swiper because of this very same issue. Not only that but this issue is happening for more people https://github.com/kidjp85/react-id-swiper/issues/273

What surprises me is the fact that this was reported and the issue (3047) was deleted too which seems odd enough. 

So, my question is why do you keep silencing these issues when they are real and they make using your library for older browsers a nuisance? I assume you don't want that as, otherwise, you wouldn't have continued with Swiper this long.

Can we move forward and try to fix this situation? If not, what's preventing it from happening?

Thanks.

Antonio-Laguna:
Right now [as per Webpack's documentation](https://webpack.js.org/configuration/resolve/), the `module` version is the one that's going to be resolved when it's required. However, `module` as it stands today, is not transpiled and has ES6 notation. Since mostly everyone excludes `/node_modules/` from their transpilation, it does not work and there are little workarounds around it. This is especially nasty for libraries that depend on Swiper such as React/Vue/Angular/nameYourFramework wrappers.

Either ESM gets transpiled (so no `class` nor `const` notation but imports are good to stay) or this which is the quickest win.

Fixes #3199 #3073
Also https://github.com/kidjp85/react-id-swiper/issues/332 and https://github.com/kidjp85/react-id-swiper/issues/317

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,9 +2,9 @@
   "name": "swiper",
   "version": "4.5.0",
   "description": "Most modern mobile touch slider and framework with hardware accelerated transitions",
-  "main": "dist/js/swiper.js",
+  "main": "dist/js/swiper.esm.bundle.js",
   "jsnext:main": "dist/js/swiper.esm.bundle.js",
-  "module": "dist/js/swiper.esm.bundle.js",
+  "module": "dist/js/swiper.js",
   "scripts": {
     "build:dev": "cross-env NODE_ENV=development gulp build",
     "build:prod": "cross-env NODE_ENV=production gulp build",


nolimits4web:
This can potentially break most of existing apps + see #3199 3199

Antonio-Laguna:
I don't think it would break anything in the mid term. You could also release a major version if you feel like it would break.  There's also the fact that forcing other people to change their config just because of Swiper doesn't seem like an ideal situation. Right?

PR #3206, opened 2019-09-16
fix rtl issue

osamaalsabbagh:
in rtl mode the prev button will be hidden after touch event although if we don't reach the start of the swiper (the right side of it) so I made some update to fix this bug


nolimits4web:
Sorry, changes should be done only to `/src/` files

osamaalsabbagh:
ok I made another pull request in the correct place

PR #3215, opened 2019-09-18
feat(shadow): fix find wrapper

ghaiat:
Fix swiper for use with shadow dom components

### DIFF
--- a/src/components/core/core-class.js
+++ b/src/components/core/core-class.js
@@ -21,465 +21,478 @@ import checkOverflow from './check-overflow/index';
 import defaults from './defaults';
 
 const prototypes = {
-  update,
-  translate,
-  transition,
-  slide,
-  loop,
-  grabCursor,
-  manipulation,
-  events,
-  breakpoints,
-  checkOverflow,
-  classes,
-  images,
+    update,
+    translate,
+    transition,
+    slide,
+    loop,
+    grabCursor,
+    manipulation,
+    events,
+    breakpoints,
+    checkOverflow,
+    classes,
+    images,
 };
 
 const extendedDefaults = {};
 
 class Swiper extends SwiperClass {
-  constructor(...args) {
-    let el;
-    let params;
-    if (args.length === 1 && args[0].constructor && args[0].constructor === Object) {
-      params = args[0];
-    } else {
-      [el, params] = args;
-    }
-    if (!params) params = {};
-
-    params = Utils.extend({}, params);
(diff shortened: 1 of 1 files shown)


nolimits4web:
Can you please keep formatting? Right now it is not possible to understand where changes made

nolimits4web:
Also can you please provide a minimal example of the usecase which it fixes?

ghaiat:
@nolimits4web done. sorry for the mess

basically the use case is if the element is a shadow dom element, you cannot query directly children, you have to use shadowRoot
the code is just checking if the el has a shadowRoot or not

nolimits4web:
You mean when Swiper used as a web component which includes wrapper and you pass slides elements as slots?

ghaiat:
yes

nolimits4web:
Merged, thanks!

ghaiat:
thanks, could you trigger a new release so i can test on my app ?
cheers

taoqf:
I need a example too.
I find it's difficult to use swiper in shadowdom.

```ts
const tpl = `
<link href="//cdn.jsdelivr.net/npm/swiper@5.2.1/css/swiper.css" rel="stylesheet">
<div class="swiper-container">
	<!-- Additional required wrapper -->
	<div class="swiper-wrapper">
		<slot></slot>
	</div>
	<!-- If we need pagination -->
	<div class="swiper-pagination"></div>

	<!-- If we need navigation buttons -->
	<div class="swiper-button-prev"></div>
	<div class="swiper-button-next"></div>

	<!-- If we need scrollbar -->
	<div class="swiper-scrollbar"></div>
</div>
`;

export default class MySwiper extends HTMLElement {
	private dom: ShadowRoot;
	constructor() {
		super();
		this.dom = this.attachShadow({ mode: 'open' }); // 1
		this.dom.innerHTML = tpl;
	}

	public async connectedCallback() {
		const root = this;
		// const root = this.dom.querySelector<HTMLElement>('.swiper-container');  // 2
		const nextEl = this.dom.querySelector<HTMLElement>('.swiper-button-next')!;
		const prevEl = this.dom.querySelector<HTMLElement>('.swiper-button-prev')!;
		new Swiper(root!, {
			navigation: {
				nextEl,
				prevEl
			}
		});
	}
}

window.customElements.define('my-swiper', MySwiper);
```

1. `el && el.shadowRoot` will be null if mode is `closed`;
1. `el && el.shadowRoot` will be null too if I use `swiper-container`.

PR #3219, opened 2019-09-19
listen to touchcancel event and treat it as touchend

hsxfjames:
If the page is loaded via an app's embedded webview, the slides will not be released when we swipe far enough away out of the webview edges. It looks like as it get stuck. This PR solve the problem.


hsxfjames:
In addition, we can also add pointercancel to desktop events, but there is no cancel event for MouseEvent. Maybe need to listen mouseout on document. Currently I have only test the touch events on mobile pages.

nolimits4web:
Thanks!

PR #3225, opened 2019-09-24
Fixed #3025

### LINKED ISSUE #3025: Loop does not work with float number slides per view
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: 4.5
* Live Link or JSFiddle/Codepen or website with issue: https://codepen.io/komarovdesign/pen/YMKOzd 

### What you did
Set float number for slidesPerView with loop mode. For some design cases, you must have just a pick of the next slide

### Expected Behavior
loop should work in both directions 

### Actual Behavior
you can slide endlessly only for prev slides but if you try to slide for next slides, the loop stops working

This issue already was reported #2773 but was not fixed. Please take a closer look at this issue since it is definitely a bug. Thank you

robpop:
Modified **loopCreate()** to set *loopedSlides* to a float rounded up to the nearest whole number.

original line in **loopCreate()**
```
swiper.loopedSlides = parseInt(params.loopedSlides || params.slidesPerView, 10);
```
new line
```
swiper.loopedSlides = Math.ceil(parseFloat(params.loopedSlides || params.slidesPerView, 10));
```
This modification allows *slidesPerView* to be a float value while also being able to swipe next through the loop.

Test: https://plnkr.co/edit/T6Upr5sMv2j7sM7MPCkw?p=preview

### DIFF
--- a/src/components/core/loop/loopCreate.js
+++ b/src/components/core/loop/loopCreate.js
@@ -22,7 +22,7 @@ export default function () {
 
   if (params.slidesPerView === 'auto' && !params.loopedSlides) params.loopedSlides = slides.length;
 
-  swiper.loopedSlides = parseInt(params.loopedSlides || params.slidesPerView, 10);
+  swiper.loopedSlides = Math.ceil(parseFloat(params.loopedSlides || params.slidesPerView, 10));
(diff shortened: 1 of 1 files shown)


nolimits4web:
Looks good, merged, thanks!

PR #3244, opened 2019-09-30
Fixed #3036

### LINKED ISSUE #3036: Pagination bug, reproduces on swiper demo page too.
This is a (multiple allowed):
* [x] bug

Swiper Version: So I see that bug on swiper a long time ago and can be reproduced on many versions.
Platform/Target and Browser Versions: Chrome, macOS
Live Link or JSFiddle/Codepen or website with isssue: http://idangero.us/swiper/demos/200-infinite-loop.html

**Steps to reproduce.**
1. Go to demo page http://idangero.us/swiper/demos/200-infinite-loop.html
2. Try click to bullet points (2-4), then to the last couple of bullets, repeat it a several times.

Result: Swiper show randomly slides from time to time.
Expectation: Swiper must show slides which trigger bullet point.

Here is my gif from the swiper demo page: https://cl.ly/7c3bb1d75462

robpop:
Removed ```swiper.allowClick = false;``` in **onTouchMove** event. Having that line was sometimes causing clicks on pagination bullets to be inaccurate. 

What was the original purpose for having that line there?

Test: https://next.plnkr.co/edit/qGE9irxx8wYgRJik?open=lib%2Fscript.js

### DIFF
--- a/src/components/core/events/onTouchMove.js
+++ b/src/components/core/events/onTouchMove.js
@@ -100,7 +100,6 @@ export default function (event) {
   if (!data.startMoving) {
     return;
   }
-  swiper.allowClick = false;
   if (!params.cssMode) {
     e.preventDefault();
   }
--- a/src/components/core/events/onTouchMove.js
+++ b/src/components/core/events/onTouchMove.js
@@ -100,6 +100,7 @@ export default function (event) {
   if (!data.startMoving) {
     return;
   }
+  swiper.allowClick = false;
   if (!params.cssMode) {
     e.preventDefault();
   }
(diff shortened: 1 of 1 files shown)


nolimits4web:
No, it should be there to prevent clicks on links during swipe.

robpop:
That makes sense. With that in mind, I modified the change so that it checks if the event target is a pagination bullet only. Let me know what you think.

nolimits4web:
Sorry, i think this is still a bit wrong. Swiping over pagination still must be possible without having it being clicked. So such prevention should be done if pagination is clickable and swiping was for a small distance (e.g. few pixels).
But why do we need this fix at all? There are still usecase where it will fail or introduce new issues. I would better just recommend to enable `threshold` swiper parameter instead

robpop:
No worries, I'll admit this is a fringe issue. The original author indicated that this is a non-mobile issue which occurs somewhat randomly. I can continue to fine tune this change to detect the clickable property, or we can close this request and the issue and go with the ```threshold``` fix.

PR #3255, opened 2019-10-06
Fixed #3029

### LINKED ISSUE #3029: Pagination Dynamic Bullets Issue
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: 4.5.0
* Platform/Target and Browser Versions: Windows, Chrome
* Live Link or JSFiddle/Codepen or website with isssue: https://codepen.io/anon/pen/NmGzMN

### What you did
Swiped from the penultimate slide to the final slide (slide 9 to slide 10).

### Expected Behavior
The pagination bullets should show two main bullets (as is specified when configuring swiper)

### Actual Behavior
The pagination bullets show one main bullet.

### Additional Notes
It looks like swiping from slide 1 to slide 10 is fine

robpop:
This fixes the issue of pagination bullets at the end of a loop not having the ```-main``` class as per the ```dynamicMainBullets``` property [as described in the original issue](https://github.com/nolimits4web/swiper/issues/3029#issue-428709036).

Added a conditional in ```pagination.js``` to check if ```bulletIndex``` is at the end of the loop minus ```dynamicMainBullets```.

Test: https://plnkr.co/edit/JbSx3nijaSgs45qwXUr2?p=preview

### DIFF
--- a/src/components/pagination/pagination.js
+++ b/src/components/pagination/pagination.js
@@ -84,22 +84,22 @@ const Pagination = {
           for (let i = firstIndex; i <= lastIndex; i += 1) {
             bullets.eq(i).addClass(`${params.bulletActiveClass}-main`);
           }
-          if (bulletIndex >= bullets.length-params.dynamicMainBullets) {
-            for (let i = params.dynamicMainBullets; i >= 0; i--) {
-              bullets.eq(bullets.length - i).addClass(((params.bulletActiveClass) + "-main"));
+          if (bulletIndex >= bullets.length - params.dynamicMainBullets) {
+            for (let i = params.dynamicMainBullets; i >= 0; i -= 1) {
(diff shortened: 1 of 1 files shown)


nolimits4web:
Didn't get into details but this fix breaks another behavor, try setting `loop: false` and `dynamicManBullets: 5` some bullets during swiping stay small

robpop:
I added a fix for that just now. 

I'm not totally satisfied yet with this result though. Ideally I'd like the ```left``` offset for the bullets to be the same when ```loop: false``` and ```loop: true```. I'm having difficulty getting this to work properly. This behavior is set by ```bulletsOffset``` on line 121 of pagination.js.

nolimits4web:
Ok, let's merge it for now and see how it goes. Thanks!

PR #3259, opened 2019-10-08
Use changedTouches instead of targetTouches inside of onTouchMove

### LINKED ISSUE #2956: Bugsnag: o.targetTouches is undefined
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: 4.0.7
* Vue Awesome Swiper Version: 3.1.3
* Platform/Target and Browser Versions: Chrome Mobile Android 7.0, iOS Safari iOS 12.1
* Live Link or JSFiddle/Codepen or website with issue: [https://www.optusperks.com.au](https://www.optusperks.com.au)

### What you did
We using the Vue Awesome Swiper package in our Vue project.

### Expected Behavior
No `bugsnag` errors.

### Actual Behavior
We are getting consistent bug snags errors `o.targetTouches is undefined` which point to a line in the swiper library:
`node_modules/swiper/dist/js/swiper.js:2623`
`touches.currentX = e.type === 'touchstart' ? e.targetTouches[0].pageX : e.pageX;`

screen shot 2019-01-25 at 2 02 04 pm

The best outcome here is that it's a type error which can be fixed to help suppress future warnings caused by the package.

tom-sherman:
Fixes #2956

[MDN currently describes](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/targetTouches#Browser_compatibility) `targetTouches` as not supported but this isn't the case, at least in iOS 13.

Most of the time referencing `targetTouches` is ok however sometimes it doesn't contain any `Touch` objects inside the `touchmove` event. This happens when the drag was initiated outside of the Swiper container.

### DIFF
--- a/src/components/core/events/onTouchMove.js
+++ b/src/components/core/events/onTouchMove.js
@@ -15,8 +15,8 @@ export default function (event) {
     return;
   }
   if (data.isTouchEvent && e.type === 'mousemove') return;
-  const pageX = e.type === 'touchmove' ? e.targetTouches[0].pageX : e.pageX;
-  const pageY = e.type === 'touchmove' ? e.targetTouches[0].pageY : e.pageY;
+  const pageX = e.type === 'touchmove' ? e.changedTouches[0].pageX : e.pageX;
+  const pageY = e.type === 'touchmove' ? e.changedTouches[0].pageY : e.pageY;
   if (e.preventedByNestedSwiper) {
     touches.startX = pageX;
     touches.startY = pageY;


nolimits4web:
Well, this is pretty weird as such thing as `the drag was initiated outside of the Swiper container.` shouldn't happen at all, so looks like it is iOS bug which triggers `touchstart` event in this case when. Event more weird that it can be fixed by just adding this line of code anywhere:
```
document.addEventListener('touchstart', () => {});
```

So let's merge this and i also add this dummy touch handler

utlime:
Hey @nolimits4web, when do you plan release this fix for 4rd version?

PR #3283, opened 2019-10-16
Add centeredSlidesBounds parameter. Prevents gaps at beginning and end…

jasonlav:
Adds `centeredSlidesBounds` parameter to prevent gaps at the beginning and end of the slider when `centeredSlide` is enabled.

It manipulates the `snapGrid` variable, but not the `slidesGrid` constant. Unknown any side effects associated with this, but it appears to function as expected in a number of situations.

Not intended to be used with `loop` or `pagination`.

### DIFF
--- a/src/components/core/update/updateSlides.js
+++ b/src/components/core/update/updateSlides.js
@@ -255,6 +255,20 @@ export default function () {
     } else slides.filter(slidesForMargin).css({ marginBottom: `${spaceBetween}px` });
   }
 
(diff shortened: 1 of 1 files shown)


nolimits4web:
@robpop from what i see it introduces new behavior - first and last slides will stick to bound, while other will be centered. Why not, it is an optional thing that doesn't break anything when it is not enabled

nolimits4web:
Merged

PR #3288, opened 2019-10-19
remove redundant condition

kashtian:
Remove redundant condition in loopFix function.
The condition make a gap when click prev button.
There is swiper option:
`swiperOption: {
        slidesPerView: 'auto',
        centeredSlides: true,
        loop: true,
        loopedSlides: 3,
        navigation: {
          nextEl: '.swiper-button-next',
          prevEl: '.swiper-button-prev'
        }
      }`


kashtian:
There is a [online example](https://codepen.io/kashtian/pen/YzzpENx).
When you click the prev button three times, you will see the gap.

nolimits4web:
Thanks, merged!

PR #3291, opened 2019-10-19
Fixed #3005

### LINKED ISSUE #3005: Parameters watchOverflow and slidesOffsetBefore/slidesOffsetAfter don't work together
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: 4.5.0
* Platform/Target and Browser Versions: macOS Chrome
* Live Link or JSFiddle/Codepen or website with issue: https://jsfiddle.net/rcug4j1w/ (adapted from swiper demo for Freemode)

### What you did
Freemode swiper with watchOverflow set to true, and slidesOffsetBefore/slidesOffsetAfter set to any number

### Expected Behavior
Swiper should lock scrolling when all slides PLUS the before/after offset space are fully visible

### Actual Behavior
Swiper doesn't lock scrolling, even if viewport is wide enough to fit everything in view. I think that if slidesOffsetBefore/slidesOffsetAfter are set, the before/after offset pixels just need to be factored into the visible width when checking for overflow?

robpop:
Modified checkOverflow() to account for ```slidesOffsetBefore``` and ```slidesOffsetAfter```. This fixes a [bug](https://github.com/nolimits4web/swiper/issues/3005) where ```watchOverflow: true``` wouldn't work with ```slidesOffsetBefore``` and ```slidesOffsetAfter``` set to a non-zero value.

Test: https://plnkr.co/edit/o6gQxkfqcVjYEF3rPskx?p=preview

### DIFF
--- a/src/components/core/check-overflow/index.js
+++ b/src/components/core/check-overflow/index.js
@@ -1,9 +1,16 @@
 
 function checkOverflow() {
   const swiper = this;
+  const params = swiper.params;
   const wasLocked = swiper.isLocked;
+  let lastSlidePosition = (params.slidesOffsetBefore + (params.spaceBetween * (swiper.slides.length - 1)) + ((swiper.slides[0]).offsetWidth) * swiper.slides.length)
 
-  swiper.isLocked = swiper.snapGrid.length === 1;
+  if ((typeof params.slidesOffsetBefore !== 'undefined' && params.slidesOffsetBefore !== 0) || (typeof params.slidesOffsetAfter !== 'undefined' && params.slidesOffsetAfter !== 0)) {
+    swiper.isLocked = lastSlidePosition <= swiper.size;
+  } else {
+    swiper.isLocked = swiper.snapGrid.length === 1;
+  }
+
   swiper.allowSlideNext = !swiper.isLocked;
   swiper.allowSlidePrev = !swiper.isLocked;
 
--- a/src/components/core/check-overflow/index.js
+++ b/src/components/core/check-overflow/index.js
@@ -3,9 +3,9 @@ function checkOverflow() {
   const swiper = this;
   const params = swiper.params;
   const wasLocked = swiper.isLocked;
-  const lastSlidePosition = (params.slidesOffsetBefore + (params.spaceBetween * (swiper.slides.length - 1)) + ((swiper.slides[0]).offsetWidth) * swiper.slides.length);
+  const lastSlidePosition = swiper.slides.length > 0 ? (params.slidesOffsetBefore + (params.spaceBetween * (swiper.slides.length - 1)) + ((swiper.slides[0]).offsetWidth) * swiper.slides.length) : undefined;
 
-  if ((typeof params.slidesOffsetBefore !== 'undefined' && params.slidesOffsetBefore !== 0) || (typeof params.slidesOffsetAfter !== 'undefined' && params.slidesOffsetAfter !== 0)) {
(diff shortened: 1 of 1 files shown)


src/components/core/check-overflow/index.js
```diff
@@ -1,9 +1,16 @@
 
 function checkOverflow() {
   const swiper = this;
+  const params = swiper.params;
   const wasLocked = swiper.isLocked;
+  const lastSlidePosition = (params.slidesOffsetBefore + (params.spaceBetween * (swiper.slides.length - 1)) + ((swiper.slides[0]).offsetWidth) * swiper.slides.length);
```
Maxklebe: When passing no slides this would throw: `Uncaught TypeError: Cannot read property 'offsetWidth' of undefined`
Probably a good idea to conditional check this slides before setting lastSlidePosition.

robpop: Ah, good catch. Added a fix for that now.

src/components/core/check-overflow/index.js
```diff
@@ -3,9 +3,9 @@ function checkOverflow() {
   const swiper = this;
   const params = swiper.params;
   const wasLocked = swiper.isLocked;
-  const lastSlidePosition = (params.slidesOffsetBefore + (params.spaceBetween * (swiper.slides.length - 1)) + ((swiper.slides[0]).offsetWidth) * swiper.slides.length);
+  const lastSlidePosition = swiper.slides.length > 0 ? (params.slidesOffsetBefore + (params.spaceBetween * (swiper.slides.length - 1)) + ((swiper.slides[0]).offsetWidth) * swiper.slides.length) : undefined;
```
Maxklebe: Minor detail but if you'd want to shorten this you can do this instead as this would evaluate to `undefined` if length is 0
```
const lastSlidePosition = swiper.slides.length > 0 && (params.slidesOffsetBefore + (params.spaceBetween * (swiper.slides.length - 1)) + ((swiper.slides[0]).offsetWidth) * swiper.slides.length);
```

src/components/core/check-overflow/index.js
```diff
@@ -3,9 +3,9 @@ function checkOverflow() {
...
 
-  if ((typeof params.slidesOffsetBefore !== 'undefined' && params.slidesOffsetBefore !== 0) || (typeof params.slidesOffsetAfter !== 'undefined' && params.slidesOffsetAfter !== 0)) {
+  if (((typeof params.slidesOffsetBefore !== 'undefined' && params.slidesOffsetBefore !== 0) || (typeof params.slidesOffsetAfter !== 'undefined' && params.slidesOffsetAfter !== 0)) && lastSlidePosition !== 'undefined') {
```
Maxklebe: You could shorten this as well since the if statement would catch both undefined and 0 👍 
```
if(params.slidesOffsetBefore && params.slidesOffsetAfter && lastSlidePosition) {
...
}
```

robpop: Clever shorthands, I like it. Thanks!

nolimits4web:
Thanks, merged!

JohnThomson:
Still doesn't work if slidesOffsetAfter is non-zero but slidesOffsetBefore is zero.
I think the test in the change should be 
(params.slidesOffsetBefore || params.slidesOffsetAfter) && lastSlidePosition

JohnThomson:
(work-around: set slidesOffsetBefore to some very small number)

Issue #3222, opened 2019-09-19
Incorrect order and margin


marwej:
This is a (multiple allowed):
* [x] bug

* Swiper Version: 5.0.3
* Platform/Target and Browser Versions: macOS
* Live Link or JSFiddle/Codepen or website with isssue: N/A

### What you did
17 slides, 5 columns, 2 rows -> 2 groups (the second group displays two slides from the first group too)

```
<Swiper
            slidesPerView={5}
            slidesPerGroup={5}
            slidesPerColumn={2}
            slidesPerColumnFill='row'
            spaceBetween={50}
            threshold={20}
            rebuildOnUpdate
            modules={[ Pagination, Mousewheel, Keyboard ]}
            pagination={{ el: hasManyPages && '.swiper-pagination', clickable: true }}
            mousewheel
            keyboard
            >
```

### Actual Behavior

Still not correct calculation of order and margin. I have named the slides following the order in which they are added (sorted alphabetically). 

Group 1:
<img width="1528" alt="Skärmavbild 2019-09-19 kl  15 06 57" src="https://user-images.githubusercontent.com/5296840/65247068-0573eb80-daf0-11e9-8709-6166293325ff.png">

Group 2:
<img width="1528" alt="Skärmavbild 2019-09-19 kl  15 07 05" src="https://user-images.githubusercontent.com/5296840/65247081-0d339000-daf0-11e9-94ab-07a3b1b09c2c.png">

HTML:
<img width="1917" alt="Skärmavbild 2019-09-19 kl  15 08 05" src="https://user-images.githubusercontent.com/5296840/65247099-158bcb00-daf0-11e9-9fe8-ab53354a88a4.png">

nolimits4web:
Please provide a link to JSFiddle or similar which uses just Swiper, not some third-party implementation or wrapper

marwej:
I'll try to fix a demo. This is just another example in my application. 4 columns, 2 rows, 3 slides. The third slide got a `top-margin`.

![image](https://user-images.githubusercontent.com/5296840/65270849-4d106c80-db1c-11e9-8ac9-19d6ce33e095.png)

marwej:
I managed to get it working with this change here https://github.com/nolimits4web/swiper/blob/master/src/components/core/update/updateSlides.js#L85. I don't know if there are any unintended consequences though. 

```
[...]

PR #3294, opened 2019-10-20
Fix issue with wrong row/column if not full groups

marwej:
Fix to issue #3222. Slides could get misplaced with wrong order and margin if `slidesPerColumnFill = row` and `slidesPerGroup > 1` and `slidesPerColumn > 1` and the last group was not filled up with slides.


nolimits4web:
Merged, thanks!

PR #3302, opened 2019-10-22
Fix duration when freeModeSticky is active

### LINKED ISSUE #3301: Sloooooooow scrolling (10+ seconds per slide) with freeModeSticky
If freeModeSticky is active and the user ends a swipe with a slow-velocity event, then swipe durations can be 20+ seconds to slide one (or zero!) slides.  It's easy to see this when simulating touch with mouse events. To fix this, we should limit single-slide swipes to the default slide duration.  I've built a PR to fix this-- will submit it in a few minutes.

This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: 5.1.0
* Platform/Target and Browser Versions: tested on MacOS Chrome & Firefox, but should repro on all desktop browsers on all OSes because the core problem is caused by math inside Swiper.
* Live Link or JSFiddle/Codepen or website with issue: https://justingrant.github.io/swiper-repro/

### What you did
1. https://justingrant.github.io/swiper-repro/ 
OR 
2. Click and hold the left mouse button over the slide, swipe it slowly to the left, and let go of the mouse. The goal should be to swipe one slide. You may need to try a few times to get it the problem to repro, because the repro depends on the last mouse event covering a small but non-zero distance. 

### Expected Behavior
After the mouse button is released, Swiper will scroll to the next slide in less than 1 second

### Actual Behavior
Scroll to the next slide takes 10+ seconds. 

Screen-Recording-2019-10-21-at-7 41 38-PM

justingrant:
This PR fixes #3301 by limiting the maximum swipe duration when `freeModeSticky` is active, as follows: 
* If the `moveDistance` is less than the current slide size, then the duration is limited to the default swipe speed (`params.speed`)
* If the `moveDistance` is 1x to 2x the current slide size, then the max duration is limited to 1.5x the default swipe speed
* Otherwise the max duration is limited to 2.5x the default swipe speed. 

The underlying problem is caused because the duration was previously calculated without taking into account the extra distance required to snap to the nearest slide when `freeModeSticky` is active.

You can use the same site from #3301 to verify that this PR is working: 

To verify the PR is working, use the same repro steps from #3301.  You should see the desired behavior from #3301.

### DIFF
--- a/src/components/core/events/onTouchEnd.js
+++ b/src/components/core/events/onTouchEnd.js
@@ -164,6 +164,24 @@ export default function (event) {
         } else {
           momentumDuration = Math.abs((newPosition - swiper.translate) / swiper.velocity);
         }
+        if (params.freeModeSticky) {
+          // If freeModeSticky is active and the user ends a swipe with a slow-velocity
+          // event, then durations can be 20+ seconds to slide one (or zero!) slides.
+          // It's easy to see this when simulating touch with mouse events. To fix this,
+          // limit single-slide swipes to the default slide duration. This also has the
+          // nice side effect of matching slide speed if the user stopped moving before
+          // lifting finger or mouse vs. moving slowly before lifting the finger/mouse.
+          // For faster swipes, also apply limits (albeit higher ones).
+          const moveDistance = Math.abs((rtl ? -newPosition : newPosition) - swiper.translate);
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thanks!

PR #3303, opened 2019-10-22
Fix `freeModeSticky` resizing behavior to ensure it stays sticky after a resize

### LINKED ISSUE #2708: [BUG] Freemode breaks swiper responsive
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: 4.3.3
* Platform/Target and Browser Versions: Chrome Win
* Live Link or JSFiddle/Codepen or website with isssue: https://jsfiddle.net/Tucsky/0jLz2dkc/

### What you did
Setup a classic swiper
Enable freemode with sticky (important)
Try resize the swiper container

### Expected Behavior
Each slide width & wrapper translate3d updating accordingly to the container width = responsive OK

### Actual Behavior
Slides width updating but wrapper translate is NOT => responsive KO

the issue

justingrant:
This PR fixes #2708 by handling resizing in the same way regardless of whether `freeModeSticky` is active or not.  See https://github.com/nolimits4web/swiper/issues/2708#issuecomment-518935544 for an explanation of the problem and a description of the thinking behind this PR. 

You can use the same site from #3301 to verify that this PR is working: 

To verify the PR is working, simply resize the browser. If the fix works, after resizing the slide should stick to the edges. (Without this fix, the slide will not be stuck to the edge after resizing.)

### DIFF
--- a/src/components/core/events/onResize.js
+++ b/src/components/core/events/onResize.js
@@ -20,23 +20,13 @@ export default function () {
   swiper.updateSize();
   swiper.updateSlides();
 
-  if (params.freeMode) {
-    const newTranslate = Math.min(Math.max(swiper.translate, swiper.maxTranslate()), swiper.minTranslate());
-    swiper.setTranslate(newTranslate);
-    swiper.updateActiveIndex();
-    swiper.updateSlidesClasses();
-
-    if (params.autoHeight) {
-      swiper.updateAutoHeight();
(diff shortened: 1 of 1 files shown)


nolimits4web:
Great, thanks, not already remember why here was a different freeMode handling, but checked your fix, works great

PR #3304, opened 2019-10-22
Faster & smoother mousewheel inertial scrolling

### LINKED ISSUE #3154: RFC proposal: detect mousewheel deceleration for smoother sticky snapping
Swiper today handles modern trackpads (MacBooks, Surface laptops, etc) via the `mousewheel` option. But when also using `freeMode` and `freeModeSticky`, there's an awkward delay between when the momentum scroll stops and when Swiper snaps to the nearest item.  I want to discuss a possible solution to that awkward delay. I'd consider building a PR for this solution if there's interest.

Here's the idea: momentum scrolling follows a predictable pattern of `wheel` events that are sent during deceleration at the end of a momentum scroll. Instead of the current solution which waits for a timeout after the last wheel event, instead could we detect the deceleration event pattern, bail out early of the system-managed scrolling, and snap to the desired position a few hundred milliseconds earlier than the current implementation can?  

Even if this works as intended, it probably wouldn't be as smooth as the touch solution, but it should be smoother and faster than the current implementation.  As part of the solution I also have an idea how to reduce the "post-scroll" lockout delay discussed here: https://github.com/nolimits4web/swiper/issues/2086#issuecomment-326455917.

Here's a real-world example from a simple React app I built that logs `wheel` events to the console during a momentum scroll. When a momentum scroll is slowing down, the `deltaX` (or `deltaY` for vertical scrolling) of `wheel` events follow a predictable pattern, at least on my 2015 MacBook Pro: 
* There's ~30 events sent in the final 500 msecs of the momentum scroll
* An event is sent every 16ms tick (or almost every tick) 
* `DeltaX` gradually decreases while the number of events for each ΔX value gradually increases: 
   * 1-2 events each for ΔX 9 through 5
   * 3-4 events each for ΔX 4 through 2
   * 8 events for ΔX 1

Below is a table showing these events. The first number in each row is the number of milliseconds before the last event, measured using  `Date.now()`.

```
Time |  ΔT | ΔX
-500 | -16 | 9
-484 | -17 | 8
-467 | -17 | 8
-450 | -17 | 7
-433 | -17 | 6
-416 | -17 | 6
-399 | -16 | 5
-383 | -16 | 5
-367 | -17 | 5
-350 | -17 | 4
-333 | -16 | 4
-317 | -17 | 4
-300 | -17 | 3
-283 | -18 | 3
-265 | -15 | 3
-250 | -17 | 3
-233 | -17 | 2
-216 | -17 | 2
-199 | -15 | 2
-184 | -17 | 2
-167 | -17 | 1
-150 | -17 | 1
-133 | -16 | 1
-117 | -18 | 1
 -99 | -16 | 1
 -83 | -34 | 1
 -49 | -16 | 1
 -33 | -33 | 1
   0 |   0 | 1
```

It'd take some experimentation to get it right, but a starting point for a detection algorithm could be something like: 
* when wheel scrolling starts, store a rolling window of the last N (N=20?) `wheel` events
* do all of those events have decreasing or same ΔX?
* did all events arrive in the last M (M=300?) msecs?
* does the earliest event have a ΔX that's at least P (P=2?) larger than the most recent event?
* does the latest event have a ΔX that's smaller than Q (Q=5?) pixels?

If the answer to all questions above is "yes" then:
* programmatically scroll to the nearest snap point (current Swiper behavior)
* ignore wheel events for 300 msecs after the last `wheel` (current Swiper behavior?), UNLESS...
   * ...a new `wheel` event arrives with a larger ΔX than the most recent event received during deceleration, which indicates that the user has started another scroll.  In this case, clear the timeout and start scrolling.

If for some reason (e.g. overloaded CPU) the algorithm doesn't get triggered, simply fall through to the current behavior.

Does something like this seem like it might work?  If yes, would you be interested in a PR for it?  Also, let me know if you have any advice about the problems inherent in this approach, because you probably know this problem space much better than I do.  Thanks!

### This is a (multiple allowed):
* [ ] bug
* [ ] enhancement

* Swiper Version: 4.5.0
* Platform/Target and Browser Versions: MacOS (Chrome, Safari, Firefox-- all latest version) 
* Live Link or JSFiddle/Codepen or website with isssue: https://codepen.io/justingrant/pen/JgoPWP

### What you did
Scrolled a `Swiper` using a MacBook Pro trackpad with `mousewheel`, `freeMode`, and `freeModeSticky` enabled.

### Expected Behavior
Smooth deceleration to a snap point, like on a touch device.

### Actual Behavior
Smooth deceleration to a non-snap point, then a short delay, then a smooth scroll to the nearest snap point, and finally a significant delay before I can scroll again with the trackpad.

justingrant:
This PR fixes #3154 by enabling smoother and faster inertial scrolling when `freeModeSticky` is on. Changes include:
* Fixes the problem where you're "locked out" of wheel scrolling for a few hundred msecs after the end of a momentum wheel scroll. With this PR you can repeatedly scroll with wheel or two-finger trackpad gestures without any lags or pauses.
* Avoids the delay between the last `wheel` event and snapping to the closest edge. With this PR, the snap happens during the deceleration phase of the momentum scroll instead of waiting 500ms after the last `wheel` event is received, so the snap happens 600ms+ sooner than current behavior.
* Biases snapping in the direction of the scroll, instead of slowing down and then reversing a long distance.  If the projected end of the scroll is >20% of the current slide, then the scroll will continue to the edge of that slide without reversing.

The result isn't perfectly smooth inertial scrolling UX like we get with inertial scrolling for touch events. Even with this PR, there's still a noticeable velocity difference between the decelerating momentum scroll (slow velocity) and the snap (faster velocity). But the velocity difference is much smaller than current behavior, the snap is 600ms+ faster, and (IMHO) the overall UX is smoother than the current behavior.

Here's how it works:

A) Detect the final stages of a decelerating inertial scroll by analyzing the last 15 wheel events within a 500ms rolling window. When a new wheel event comes in, apply the following tests to the rolling window:
  1. do all events have decreasing or same ΔX? (If the delta is going up, then it's not a deceleration. Note: here and below, I'm saying ΔX but replace with ΔY for vertical slides.)
  2. did all events arrive in the last 500 msecs? (only respond to quick, automated decelerations not slow events caused by user actually moving the mousewheel/trackpad.)
  3. does the earliest event have an abs(ΔX) that's at least 1px larger than the most recent event? (otherwise it's not a deceleration!)
  4. does the latest event have a ΔX that's smaller than 6 pixels? (wait until it's almost at the end of a scroll when deltas are small)

B) If 1-4 above are all "yes" then we're near the end of a momentum scroll deceleration. Do the following:
  * i. clear the rolling window
  * ii. programmatically scroll to the next/prev snap point, depending on the scroll direction. Because there's remaining momentum in the scroll (remaining wheel events that we're going to ignore), and because it's more natural UX for scrolling to continue in the same direction as opposed to reversing, we'll bias the snap in the direction of the ongoing scroll.  Therefore, if it's already scrolled more than 20% of the current slide width in the current scroll direction, keep going to the end of that slide.  Otherwise reverse.
  * iii. ignore wheel events for 300 msecs after the last `wheel` event , UNLESS...
    * ...a new `wheel` event arrives with a larger or reverse-sign ΔX than the most recent event received during deceleration, which indicates that the user has started another scroll.  In this case, clear the "ignore wheel events" timeout and start storing events in a new rolling window.

C) However, if 1-4 above aren't satisfied, then wait to snap until 500ms after the last `wheel` event.  In other words, fall through to the existing behavior before this PR. This will handle cases where users are manually scrolling with the wheel (not momentum). This also ensures that if there's a problem (e.g. overloaded CPU) that prevents the algorithm from triggering, the behavior will match the current behavior before this PR.

Here's a test repo you can use to compare this PR to previous behavior: 

### DIFF
--- a/src/components/core/slide/slideToClosest.js
+++ b/src/components/core/slide/slideToClosest.js
@@ -1,19 +1,30 @@
 /* eslint no-unused-vars: "off" */
-export default function (speed = this.params.speed, runCallbacks = true, internal) {
+export default function (speed = this.params.speed, runCallbacks = true, internal, threshold = 0.5) {
   const swiper = this;
   let index = swiper.activeIndex;
   const snapIndex = Math.floor(index / swiper.params.slidesPerGroup);
 
-  if (snapIndex < swiper.snapGrid.length - 1) {
-    const translate = swiper.rtlTranslate ? swiper.translate : -swiper.translate;
+  const translate = swiper.rtlTranslate ? swiper.translate : -swiper.translate;
 
+  if (translate >= swiper.snapGrid[snapIndex]) {
+    // The current translate is on or after the current snap index, so the choice
+    // is between the current index and the one after it.
     const currentSnap = swiper.snapGrid[snapIndex];
     const nextSnap = swiper.snapGrid[snapIndex + 1];
-
-    if ((translate - currentSnap) > (nextSnap - currentSnap) / 2) {
-      index = swiper.params.slidesPerGroup;
+    if ((translate - currentSnap) > (nextSnap - currentSnap) * threshold) {
+      index += swiper.params.slidesPerGroup;
+    }
+  } else {
+    // The current translate is before the current snap index, so the choice
+    // is between the current index and the one before it.
+    const prevSnap = swiper.snapGrid[snapIndex - 1];
+    const currentSnap = swiper.snapGrid[snapIndex];
+    if ((translate - prevSnap) <= (currentSnap - prevSnap) * threshold) {
+      index -= swiper.params.slidesPerGroup;
     }
   }
+  index = Math.max(index, 0);
+  index = Math.min(index, swiper.snapGrid.length - 1);
 
   return swiper.slideTo(index, speed, runCallbacks, internal);
 }
--- a/src/components/mousewheel/mousewheel.js
+++ b/src/components/mousewheel/mousewheel.js
@@ -27,6 +27,8 @@ function isEventSupported() {
 }
 const Mousewheel = {
   lastScrollTime: Utils.now(),
+  lastEventBeforeSnap: undefined,
+  recentWheelEvents: [],
   event() {
     if (window.navigator.userAgent.indexOf('firefox') > -1) return 'DOMMouseScroll';
     return isEventSupported() ? 'wheel' : 'mousewheel';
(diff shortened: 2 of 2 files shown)


nolimits4web:
Awesome, merged!

justingrant:
Cool, thanks @nolimits4web for the quick turnaround on all 4 of my recent PRs!

PR #3306, opened 2019-10-22
Enable sourcemaps in prod builds (for easier debugging of apps using Swiper)

justingrant:
Sourcemaps are useful not only when developing Swiper itself, but also while debugging apps that use Swiper as a dependency.  For example: 
* setting breakpoints inside Swiper code
* clearer call stacks that include Swiper code
* being able to Step Into Swiper's original source in a debugger like VSCode

Currently, sourcemaps are not included with the prod builds that are the only builds published to the npm registry. To support the use-cases above, sourcemaps are needed for the production builds too. 

This PR adds sourcemap support for Swiper 5.x prod builds. Changes include: 
* Copying the /src directory into /package/src so that the source will be published to npm. 
* Enabling sourcemap files to be generated by setting `sourcemap: true` for all configs for various build types in `scripts/build-js.js`.
* Fixing up the paths inside sourcemaps to point to the actual location (`package/src`) where source files will be installed when the package is downloaded and installed from the npm registry.

Note that this PR does not include the actual .map files themselves because those are generated by the build process, and I assumed that I should not include build output in a PR.


nolimits4web:
Looks good, thanks!

PR #3313, opened 2019-10-25
Use Math.pow() instead of ** operator

limonte:
I'm getting `Unexpected token *` exceptions from the production server (Sentry) from Chrome 49 which is the latest possible version of Chrome on Win XP.

The reason is `**` operator which was introduced in ECMAScript 2016 (ES7). 

I can't see reasons not to use `Math.pow()`, it's even a bit more readable IMO because of fewer brackets.

PS. thank you for the outstanding swiper, IMO it's the best from zillions of slider plugins on GitHub.


limonte:
Getting the same issue from Chrome 50

image

PR #3319, opened 2019-10-30
Navigate by page of slides instead of by slide

hnordberg:
We use Swiper to show a set of slots for booking appointments. We show several slots (~8-10) at a time and when the user clicks the arrows we want to shift to the next set of slots (there are ~100 total slots). We made a minimal change to the Swiper code so that the user can specify swiper.params.navigation.navByPage id they want to opt in to this behavior. The choice to call slidePrev() / slideNext() in a loop was both for simplicity and so that the appropriate event handlers (if any are installed) would be called. This code is live in production here: https://www.johnmuirhealth.com/doctor


nolimits4web:
Could be a good addition. But calling slideNext/Prev in loop is wrong solution here for few reasons:
* it will break behavior if `slidesPerGroup > 1`
* it can be potentially resource heavy
So it is better to use `slideTo` method with correct new slide index

hnordberg:
`slidesPerGroup` didn't exist when we implemented this feature. But it appears to do the same thing so we will transition to that solution. Cheers :)

PR #3320, opened 2019-10-31
[feature] thumbs options - add option to control multiple slide activation

cartok:
image

In this image we use slidesPerView of 1.5 for the main images, the connected thumbnail slider uses this value to activate multiple thumbnails. In our usecase we only want the first (real) slide to get activated. CSS workarounds would be ugly or impossible.  

That's why I suggest to add an option for it. I've set the default value to the current behaviour. Additionally i rounded down the thumbsToActivate value in case someone is using it (as we did) with a fractional value (a small logic fix).

### DIFF
--- a/src/components/thumbs/thumbs.js
+++ b/src/components/thumbs/thumbs.js
@@ -110,7 +110,7 @@ const Thumbs = {
       thumbsToActivate = swiper.params.slidesPerView;
     }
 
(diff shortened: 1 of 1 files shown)


nolimits4web:
Thanks, merged! 

P.S. I have renamed the option a bit :)

cartok:
Thanks! Sure, I wasn't so clear about that option name :D

PR #3322, opened 2019-11-04
Fixed #3084

### LINKED ISSUE #3084: .once could be called more than once
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: 4.5.0
* Platform/Target and Browser Versions: MacOS Chrome
* Live Link or JSFiddle/Codepen or website with isssue: [link to codepen](https://codepen.io/Alexnortung/pen/LoJqYq)

### What you did
I created a swiper with loop set to true, where I wanted the first slide to be removed after one swipe and keeping the position. I did this by setting a once listener on the swiper on init.

```javascript
loop: true,
on: {
  init: function() {
    console.log("initted");
    this.once("slideChangeTransitionEnd", () => {
      console.log("this should only be logged once");
      const ri = this.realIndex;
      this.removeSlide(0);
      this.slideToLoop(ri, 0);
      appendFunction(); // my own custom function for testing
    });
  }
}
```

### Expected Behavior
I expected that after swiping once, the once function would be called only once

### Actual Behavior
The once function was called twice, probably because it didn't have time to update before the slideToLoop function was called.

robpop:
Modified class.js ```once``` event listener. Moved ```handler.apply(self, args)``` to run after ```if (onceHandler.f7proxy)```. This should fix a bug where this event runs more than once. See [issue 3084](https://github.com/nolimits4web/swiper/issues/3084).

### DIFF
--- a/src/utils/class.js
+++ b/src/utils/class.js
@@ -30,11 +30,11 @@ class SwiperClass {
     const self = this;
     if (typeof handler !== 'function') return self;
     function onceHandler(...args) {
-      handler.apply(self, args);
       self.off(events, onceHandler);
       if (onceHandler.f7proxy) {
         delete onceHandler.f7proxy;
       }
+      handler.apply(self, args);
     }
     onceHandler.f7proxy = handler;
     return self.on(events, onceHandler, priority);


nolimits4web:
Great, thanks!

PR #3323, opened 2019-11-04
fix: should judge existence of `navigation` in a11y module

hsxfjames:
I found this bug when I was investigating a problem, that user specified `init()` callback not be fired when we have only one initial slide.

My trace route is: `init() > swiper.preloadImages() > swiper.update() > swiper.updateProgress() > swiper.emit('reachEnd toEdge') > a11y::on::toEdge() > swiper.a11y.updateNavigation()`.

Since we haven't met this problem ever before, I think it may be a recent change to cause it easier?

### DIFF
--- a/src/components/a11y/a11y.js
+++ b/src/components/a11y/a11y.js
@@ -61,7 +61,7 @@ const a11y = {
   updateNavigation() {
     const swiper = this;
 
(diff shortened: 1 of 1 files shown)


nolimits4web:
Makes sense, thanks!

Issue #3326, opened 2019-11-05
Mouse wheel: debouncing scroll event instead of throttling it


a-barbieri:
This is a (multiple allowed):
* [x] feature-discussion (RFC)

* Swiper Version: `55496b4a`.
* Platform/Target and Browser Versions: any desktop browser, particularly visible with Apple trackpad or Magic Mouse.
* Live Link or JSFiddle/Codepen or website with isssue: https://codepen.io/a-barbieri/pen/ZEExOdb.

### What you did
Having a full-page slider I'm scrolling to see the content and because there is nothing but this slider I just scroll up and down. Easy.

### Expected Behavior
Every time I scroll I'd like to see the slider reacting movie up or down.

### Actual Behavior
Sometime the slider just doesn't move when I scroll and it gets worst when I try repeatedly to scroll to see if it works because then it don't move for the whole time I try. You will notice in fact that once it's blocked you can keep on trying to scroll continuously for 1 minute and the slider won't react. Only if you stop, wait a second, then retry, it will return to work.

---

### Possible cause: the scroll event keeps being debounced
The problem is in this if statement: 
https://github.com/nolimits4web/swiper/blob/55496b4a999c42aa8f58b007bb8efcf43d8de73d/src/components/mousewheel/mousewheel.js#L147-L158

`swiper.mousewheel.lastScrollTime` is updated even if the user scrolls before the time limit (currently 60ms). So for example:

- User scrolls 
- Swiper reacts moving a slide then it freezes for 60s
- After 50s the user scrolls again (unlikely but possible)
- The above action forces Swiper to reset `lastScrollTime` value to 60ms
- The above reset brings the total time Swiper is frozen to **110ms**
- After 50s the user scrolls again (very unlikely but still possible)
- The above action forces Swiper to reset `lastScrollTime` value to 60ms
- The above reset brings the total time Swiper is frozen to **170ms**
- and so on so forth until the user don't move the wheel anymore

### Is this the behaviour we want? 
I'm thinking about a situation in which the user has 40 slides and he wants to get to the bottom quickly by **scrolling like-there's-no-tomorrow**. He will not be able. He will have to scroll&wait at a specific pace until he decides to use the scrollbar (hoping that the designer didn't request to hide it).

### Throttle might be a better solution

The current behaviour is similar to Lodash `debounce` but what we really want maybe is `throttle`. If you don't know the difference please see [this article](https://css-tricks.com/debouncing-throttling-explained-examples/).

[...]

Issue #3325, opened 2019-11-05
Mouse wheel: unwanted increment of the frozen period


a-barbieri:
This is a (multiple allowed):
* [x] bug
* [x] enhancement

* Swiper Version: `55496b4a`.
* Platform/Target and Browser Versions: any desktop/laptop with trackpad or mouse that has mouse wheel inertia (momentum). All Apple based mouse/trackpad for sure and possibly other ones like this [super cool mouse](https://i.kinja-img.com/gawker-media/image/upload/c_scale,fl_progressive,q_80,w_800/gkv26t09cgylnmbatkdq.mp4) (LOL).
* Live Link or JSFiddle/Codepen or website with isssue: https://codepen.io/a-barbieri/pen/ZEExOdb.

### What you did
Having a full-page slider I'm scrolling to see the content and because there is nothing but this slider I just scroll up and down. Easy.

### Expected Behavior
Every time I scroll I'd like to see the slider reacting, i.e. moving up or down.

### Actual Behavior
Sometime the slider freezes when I scroll and it gets worst when I try repeatedly to scroll again to see if it works because then it doesn't move at all for the whole time I try. You will notice in fact that once it's blocked you can keep on trying to scroll continuously for 1 minute and the slider won't react. Only if you stop, wait a second, then retry, it will return to work.

---

### Possible cause: Unwanted increment of the frozen period

The problem is in this if statement: 
https://github.com/nolimits4web/swiper/blob/55496b4a999c42aa8f58b007bb8efcf43d8de73d/src/components/mousewheel/mousewheel.js#L147-L158

Last line of the above script is parsed several times after last time the mouse wheel event has been triggered. It happens when the `wheel` event is triggered by the **momentum**. Ultimately the user ends up waiting more than the time we set (currently 60ms) depending on how fast he moved the mouse wheel (which on a Mac is not present so the speed is even greater).

So rather than resetting the variable at every scroll wheel event a better idea would be to reset it only when the mouse wheel actually change the current slide.

a-barbieri:
So, I've spent the last 4 hours trying to figure out a solution.

The thing is that `wheel` events triggered **by the user** are not different from `wheel` events triggered  **by the momentum**. 

There is a very old package that tries to solve this issue, but it's not a silver bullet. It's called [lethargy](https://github.com/d4nyll/lethargy).

I bet that library does something similar to what has been done to [detect the end of the momentum](https://github.com/nolimits4web/swiper/pull/3304). 

[...]

PR #3328, opened 2019-11-06
FIX scroll wheel unwanted frozen effect

a-barbieri:
As I've already well explained in this two issues (nolimits4web/swiper#3325 and nolimits4web/swiper#3326) the user experience scrolling the slider was very bad due to the _momentum_ of the mouse/trackpad. This was particularly visible on Mac OS.

I've been fixing the behaviour by instructing the script to animate the slider when only if the scroll direction has changed or the scroll delta has increased. These two are sufficient to differentiate the scroll events triggered by a user and the ones triggered by the _momentum_.

You can try it on my fork using the playground.

### DIFF
--- a/playground/index.html
+++ b/playground/index.html
@@ -34,8 +34,8 @@
     }
     .swiper-container {
         width: 100%;
-        height: 300px;
-        margin: 50px auto;
+        height: 100vh;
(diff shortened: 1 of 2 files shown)


nolimits4web:
Looks good, thanks! 👍

Issue #1422, opened 2015-09-03
slidesPerGroup: auto


jesperveldhuizen:
Is there a way to implement an 'auto' option for the 'slidesPerGroup' option? I'm using this Swiper to create a newspaper view and we use the following structure:
- Slide 1 (width 100%)
- Slide 2 (width 50%)
- Slide 3 (width 50%)
- Slide 4 (width 50%)
- Slide 5 (width 50%)
- Slide 6 (width 100%)

So infact there are 4 views (slide 1, slide 2/3, slide 4/5, slide 6).

When I click the 'next' button from **slide 1** I go to **slide 2/3**. So far so good. When I click 'next' again I go to slide **3/4**... this is wrong as I want to go to slide **4/5**.

Is this possible? 

Set up:

``` javascript
var mySwiper = new Swiper ('.swiper-container',
{
  nextButton: '.swiper-button-next',
  prevButton: '.swiper-button-prev',
  slidesPerView: 'auto'
});
```

``` html
<div class="swiper-container">
    <div class="swiper-wrapper">
        <div class="swiper-slide"><img src="images/pages/page0001_i1.jpg" /></div>
        <div class="swiper-slide"><img src="images/pages/page0002_i1.jpg" /></div>
        <div class="swiper-slide"><img src="images/pages/page0003_i1.jpg" /></div>
        <div class="swiper-slide"><img src="images/pages/page0004_i1.jpg" /></div>
        <div class="swiper-slide"><img src="images/pages/page0005_i1.jpg" /></div>
        <div class="swiper-slide"><img src="images/pages/page0006_i1.jpg" /></div>            
    </div>
    <div class="swiper-button-prev"></div>
    <div class="swiper-button-next"></div>
</div>
```

Custom css:

``` css
.swiper-container
{
    position: absolute;
    top: 0;
    left: 0;
    right: 0;
    bottom: 0;
}

.swiper-slide
{
    width: 50%;
}

.swiper-slide:nth-child(even)
{
    text-align: right;
}

.swiper-slide:nth-child(odd)
{
    text-align: left;
}

.swiper-slide:first-child,
.swiper-slide:last-child
{
    width: 100%;
    text-align: center; 
}

.swiper-slide img
{
    max-width: 100%;
    max-height: 100%;
}
```

nolimits4web:
[...]

PR #3361, opened 2019-12-07
Implement `slidesPerGroupSkip` parameter

tobiasstrebitzer:
**Problem / Current Limitation:**
The current `slidesPerGroup` implementation strictly enforces groups to be of constant size. Previously, a `slidesPerGroup = 'auto'` option has been requested, but implementation seems complex and cumbersome.

**Use Cases:**
- One of the use-cases for a more dynamic `slidesPerGroup` setting was the ability to treat a cover slide differently, to replicate a flipbook / magazine. see #1422
- For such use-case, the new property `slidesPerGroupSkip` allows users to configure the amount of slides to treat as single groups, before starting to swich into grouped mode.

**Example:**
http://cdn.magloft.com/github/swiper/index.html

**The parameter works in the following way:**
1. If `slidesPerGroupSkip` equals `0` (default), no slides are excluded from grouping, and the resulting behaviour is the same as without this change.
2. If `slidesPerGroupSkip` is equal or greater than `1` the first X slides are treated as single groups, whereas all following slides are grouped by the `slidesPerGroup` value.

**Additional Notes:**
- As both parameters`slidesPerGroup` and `slidesPerGroupSkip` support the `breakpoints` configuration, this can also be used to create a responsive reader that switches between `dual-spread` and `single-spread` based on available browser width.

### DIFF
--- a/demos/211-slides-per-group-skip.html
+++ b/demos/211-slides-per-group-skip.html
@@ -1,0 +1,84 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <title>Swiper demo</title>
+  <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1, maximum-scale=1">
+
+  <!-- Link Swiper's CSS -->
+  <link rel="stylesheet" href="../package/css/swiper.min.css">
+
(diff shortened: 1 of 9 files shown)


nolimits4web:
Awesome, thanks!

Issue #3382, opened 2019-12-25
Mousewheel Fail when refresh


Luluda:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: ^5.2.1
* Platform/Target and Browser Versions:  macOS(0.14.3), Chrome(79.0.3945.88), safari(12.0.3)
* Live Link or JSFiddle/Codepen or website with isssue:  https://swiperjs.com/demos/270-mousewheel-control.html

### What you did
Refreshing the website and don't move my mouse, scroll mouse wheel, the page doesn't scroll. If I move the mouse or click, the bug is fixed and I can scroll the page.
This occurs  when I use touchpad too.
### Expected Behavior
When I refresh the page, even I don't move the mouse, I can scroll the website.

### Actual Behavior
This bug happens in my project, too. My config is as below:
```
swiperOption: {
                    direction: 'vertical',
                    mousewheel: {
                        disable: true,
                        sensitivity: 5
                    },
                    speed: 800,
                    parallax: true,
                    slidesPerView: 'auto'
                }
```

Then I debug the source code of swiper, I find the bug is caused by
https://github.com/nolimits4web/swiper/blob/765d9a1c062f5ee7d0a583576082cd1542988673/src/components/mousewheel/mousewheel.js#L124
before I move the mouse, `swiper.mouseEntered = false`, and `releaseOnEdges = false` cause I don't need this.
My solution is to set `releaseOnEdges = true` and this is fixed, but it's not reasonable cause I don't want releaseOnEdges feature。This problem was discussed here: [#2624](https://github.com/nolimits4web/swiper/issues/2624)
I'll commit a pr to fix this, and it works in my project

nolimits4web:
Fixed in #3383

Issue #2624, opened 2018-05-15
Mousewheel Failing?


dejardine:
This is a (multiple allowed):
* [x] bug

* Swiper Version: 4.2.6
* Platform/Target and Browser Versions: macOS Chrome  66.0.3359.139
* Live Link or JSFiddle/Codepen or website with isssue: http://jsfiddle.net/Grafik/7t1oshxw/embedded/result/

### What you did
Refresh the page, and try to mouse scroll. The mousewheel scrolling fails intermittently. Once it fails if you switch to a new tab then back again it works.

### Expected Behavior
Free scroll with mousewheel

### Actual Behavior
Mousewheel slider control does not function

hattamsoyunov:
I am also trying to solve it, I also ((found a solution? Thank you!

miela:
I have the same issue. @dejardine did you solve this?

hanoii:
Interesting, this is a very similar thing to something that's happening to me as well. Basically is like when as long as the arrow doesn't move or is there, swiper is in "no swipping" mode.

You don't need to go to a new tab, just move your mouse around a bit and it starts working again.

A similar thing can be reproduced here: http://idangero.us/swiper/demos/270-mousewheel-control.html

hanoii:
The jsfiddle is very good actually, when you refresh, you even see that the cursor is plain. once you move it it goes to grab mode.

hanoii:
Hmmm, I seem to have found a fix/workaround, although semantically is not quite right.

if you set releaseOnEdges to true, this kind of fixes it, see http://jsfiddle.net/Grafik/7t1oshxw/6/embedded/result/

Looking at the code I see a couple of things, first

https://github.com/nolimits4web/swiper/blob/8ef232e635cadda3f75da12ae6b9bd17816de57a/src/components/mousewheel/mousewheel.js#L113

If `!swiper.mouseEntered` the mousewheel doesn't work. The issue is that when you refresh the page, mouseenter has not triggered yet, so that's why it doesn;'t work.

On my use case, I also noticed that if you click on another window while keeping your swiper window on the back, attempting to scroll it also doesn't work because the mouse "left". I didn't want that to happen either and because of that line, I could remove the mouseenter/mouseleave validation.

Looking further along the code, on

https://github.com/nolimits4web/swiper/blob/8ef232e635cadda3f75da12ae6b9bd17816de57a/src/components/mousewheel/mousewheel.js#L141
and
[...]

PR #3383, opened 2019-12-25
fix mousewheel fail#3382

Luluda:
The pr is to solve [#3382](https://github.com/nolimits4web/swiper/issues/3382) and [#2624](https://github.com/nolimits4web/swiper/issues/2624). when refreshing the page and don't move the mouse, it does't trigger 'mouseenter' event, this result in mousewheel fail.
I add another condition `!target[0].contains(e.target)` to solve it. It means when  trigger `mousescroll` event, we get the event target, and if the event target is in swiper target 
```
let target = swiper.$el;
    if (swiper.params.mousewheel.eventsTarged !== 'container') {
      target = $(swiper.params.mousewheel.eventsTarged);
    }
```
we don't return.
I have tested in my project and it works.

### DIFF
--- a/src/components/mousewheel/mousewheel.js
+++ b/src/components/mousewheel/mousewheel.js
@@ -121,7 +121,11 @@ const Mousewheel = {
       e.preventDefault();
     }
 
-    if (!swiper.mouseEntered && !params.releaseOnEdges) return true;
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thanks!

PR #3389, opened 2020-01-03
Implement ratio breakpoints

tobiasstrebitzer:
**Problem / Current Limitation:**
Currently, the `breakpoints` param only supports absolute unit values, e.g. the width of the window in pixels. For some applications, e.g. on mobile devices, this turns out to be a limitation, preventing an easy way to implement certain logic like portrait / landscape switches.

**Use Cases:**
- Implementing a breakpoint switch between portrait and landscape by using `@1.00` (width/height ratio above 1.00 / square).
- More granular control of slide layout, by tuning the ratio values to guarantee a certain width / height ratio of visible slides.

**Example:**
Located in [`demos/381-ratio-breakpoints.html`](https://github.com/MagLoft/swiper/blob/963f622d329568bc28fb639a6858178f3c6b8cf6/demos/381-ratio-breakpoints.html#L82)

**The parameter works in the following way:**
1. The `breakpoints` array parameter allows keys in string notation `'@X.XX': { ... }`, where `X.XX` represents the minimum aspect ratio (`width / height`) to trigger the breakpoint.
2. Provides compatibility and mixing with integer keys (current behaviour).

### DIFF
--- a/demos/381-ratio-breakpoints.html
+++ b/demos/381-ratio-breakpoints.html
@@ -1,0 +1,103 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <title>Swiper demo</title>
+  <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1, maximum-scale=1">
+
+  <!-- Link Swiper's CSS -->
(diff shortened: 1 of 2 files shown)


nolimits4web:
Merged, thanks!

Issue #3376, opened 2019-12-18
zoom-in/out image by pinch gesture(It doesn't work when one finger is on the picture and the other is on the black background)


mayunhai:
* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 4.5.1
* Platform/Target and Browser Versions: android / Chrome 79
* Live Link or JSFiddle/Codepen or website with isssue: [official demo](https://swiperjs.com/demos/400-zoom.html)

### What you did
try zoom-in/out image by pinch gesture.
It doesn't work when one finger is on the picture and the other is on the black background

### Expected Behavior
zoom-in/out image by pinch gesture

### Actual Behavior
none

mayunhai:
`img{pointer-events: none}` can solve this problem, but it will lead to other problems, such as the inability to long press to save pictures in webView

PR #3394, opened 2020-01-07
#3376  fix bug(change all 'targetTouches' to 'touches') 

mayunhai:
#3376  fix bug

zoom-in/out image by pinch gesture(It doesn't work when one finger is on the picture and the other is on the black background)

I only changed one file, which may cause other bugs

change all 'targetTouches' to 'touches'


mayunhai:
`pointer-events:none;` seems to solve this problem, but it will cause other problems, such as long press to save the picture

Issue #3232, opened 2019-09-24
On Edge, scrollbar moves opposite direction when dragging


ngamanning:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: Swiper 5.0.3
* Platform/Target and Browser Versions: EDGE ON WINDOW 10 (on some machines), CHROME ON WINDOWS 10 (on some machines)
* Live Link or JSFiddle/Codepen or website with isssue: PREFERABLY _(IF YOU WANT YOUR ISSUE TO BE RESOLVED ASAP)_.

### What you did
I made a product slider with scrollbar using swiper.js. Below is the code: 
      <script>
            $(function () {       
            let mySwiper = new Swiper('#product-slider-1 .swiper-container', {
                speed: 600,
                spaceBetween: spaceBetween,
                slidesPerView: cardsPerSlide,
 		grabCursor:false,			
		resistanceRatio: 0.5,	
                scrollbar: {
                    el: '.swiper-scrollbar',
                    snapOnRelease: false,
                    dragSize: 'auto',
                    draggable: true,
                },
                progress: {
                    watchSlidesProgress: true
                },
            });
            mySwiper.scrollbar.updateSize();	  
        });
     </script>

### Expected Behavior
When I drag the scrollbar to the right, I expect the container to shift to the left. In this case, I expect the translateX value on  swiper-wrapper  to be negative. At the same time, the translate X value on swiper-scrollbar-drag should be possible. 
Reversely, when I drag the scrollbar to the left, I expect the container to shift to the right.  

### Actual Behavior
On Edge and Chrome of some machines, when I try to drag the scrollbar to the right, the scrollbar actually moves itself to the left and the container shift to the left. Reversely, when I try to drag the scrollbar to the left, the scrollbar moves to the right. 
I recorded the issue in the video attached. 
[swiper-issue-screen.zip](https://github.com/nolimits4web/swiper/files/3649756/swiper-issue-screen.zip)

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](http://stackoverflow.com/questions/tagged/swiper)
for that.

Before you open an issue, please check if a similar issue already exists or has been closed before.

nolimits4web:
I can assume you just click outside of scrollbar and just drag the swiper. So the behavior is correct. If not, then i need correct reproduction scenario

ngamanning:
[...]

PR #3395, opened 2020-01-09
Amended touch support test as it was returning true on none-touch windows devices

hamzahg:
This change is based on the following open issue: 

https://github.com/nolimits4web/swiper/issues/3232

I have come across the same issue in my project and have amended the script as per the PR, which has resolved the issues on MS Edge and Chrome on Windows based devices.


nolimits4web:
This fix is wrong, where did `window.DocumentTouch && document instanceof window.DocumentTouch` go? There is no `navigator.MaxTouchPoints` property

Jollus88:
#3401 
I'd recommend simply reverting commit `df41dc94c9d9224fcbeb0166e6a5a9634caa6f1d`.
This will check to see if the device has touch-capabilities, which will impact the slider.
`maxTouchPoints` will test to see if the device supports touch, which is ambiguous since what if the user is using a laptop or desktop with a touch screen.

Hirbod:
There is:
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/maxTouchPoints

Issue #3450, opened 2020-02-13
Zoom handles wrong slide #2827 - reopen


tienbuide:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 5.3.1
* Platform/Target and Browser Versions: Chrome - Version 80.0.3987.87 (Offizieller Build) (64-Bit)
* Live Link or JSFiddle/Codepen or website with isssue: https://stackblitz.com/edit/swiper-demo-39-zoom

### What you did
Configured Swiper to simulate touch and use zoom with toggle.

1. Initialize Swiper
2. Double click on slide to zoom in (manually)
3. Call mySwiper.slideNext() or mySwiper.slideTo() to switch to the next slide
4. Call mySwiper.zoom.in()

### Expected Behavior
The current slide should be zoomed.

### Actual Behavior
The previous slide is zoomed

(mostly copy and paste from #2827)

juanpasolano:
I've been having this same exact issue. Only reproducible using the method `zoom.in()` Which is what i need to use

nolimits4web:
Fixed by #3451

Issue #2827, opened 2018-10-11
Zoom handles wrong slide


ms-ox:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: v4.3.3 release, 0dcd6ca
* Platform/Target and Browser Versions: Chrome 69.0.3497.100 on Ubuntu 18.04.1 LTS

### What you did
Configured Swiper to simulate touch and use zoom with toggle.
1. Initialize Swiper
2. Double click on slide to zoom in
3. Call `mySwiper.slideNext()` or `mySwiper.slideTo()` to switch to the next slide
4. Call `mySwiper.zoom.in()`

### Expected Behavior
The current slide should be zoomed.

### Actual Behavior
The previous slide is zoomed

### Notes
- With simulate touch is activated, a double click on the slide, or switching a slide via navigation button invokes `updateClickedSlide(e)`
- This updates `swiper.clickedSlide` and `swiper.clickedIndex` to the current slide
- Calling `mySwiper.slideTo()` **doesn't invoke** `updateClickedSlide(e)`, therefore `swiper.clickedSlide` and `swiper.clickedIndex` were not updated.
- Calling `mySwiper.zoom.in()` again checks for `gesture.$slideEl = swiper.clickedSlide ? $(swiper.clickedSlide) : swiper.slides.eq(swiper.activeIndex);`
- Where `swiper.activeIndex` points the the current slide `swiper.clickedSlide` still points to the previous slide

daliborkovac:
In the previous post by @ms-ox the Expected Behavior and Actual Behavior text should be reversed. The expected behavior is that the current slide should be zoomed.

daliborkovac:
I have done my workaround like this:
In zoom.in and zoom.out functions I've commented out the parts where it uses the clicked slide and left just the part where it uses the activeIndex:

`        gesture.$slideEl = /*swiper.clickedSlide ? $(swiper.clickedSlide) :*/ swiper.slides.eq(swiper.activeIndex);
`

ms-ox:
> In the previous post by @ms-ox the Expected Behavior and Actual Behavior text should be reversed. The expected behavior is that the current slide should be zoomed.

Many thanks to @daliborkovac for bringing this up! I fixed the description.

tienbuide:
[...]

PR #3451, opened 2020-02-13
fix zoom on wrong slide - #3450 #2827

tienbuide:
The API is saying 
- `mySwiper.zoom.in(); | Zoom in image of the currently active slide`
- `mySwiper.zoom.out(); | Zoom out image of the currently active slide`
~ https://swiperjs.com/api/#zoom

The issues #3450 and #2827 describe how to reproduce a bug that does not zoom into the _currently active slide_.

This fix sets the correct slide instead of zooming into the previous slide.


nolimits4web:
Merged, thanks!

PR #3453, opened 2020-02-14
fix autoHeight to work correctly when used with centeredSlides param

### LINKED ISSUE #3448: autoHeight + centeredSlides + slidesPerView bug
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: 
5.3.1

* Platform/Target and Browser Versions: 
macOS 10.15.3
Chrome 79.0.3945.130 (64-bit) & 
Chrome 80.0.3987.106 (64-bit) &
Firefox 72.0.2 & 
Safari 13.0.5

* Live Link: 
https://auto-height-bug.stackblitz.io or 
https://stackblitz.com/edit/auto-height-bug?file=index.html

### What you did
To reproduce, goto above link then slide right 3 times, the height animates to a height that is too short for one of the items. Similarly, the height adjusts to a larger height, one slide too early.

### Expected Behavior
The height should always adjust to the highest div that is currently visible.

### Actual Behavior
The height of the container is incorrectly applied - based on the currently visible slides. It appears that the index of the slide that is driving the height of the container seems to be off by +1.

glomotion:
resolves #3448


nolimits4web:
Hi! Thanks, but there is much simpler fix i pushed in 6deb1c7db9f845cb9d5c0fcff201d7ff1e030682

PR #3468, opened 2020-02-23
Coverflow stretch percentage for responsiveness

NLNicoo:
For the coverflow effect, one can use `stretch` to define how many pixels the prev/next sliders go behind the front one. So if slider is 1000px, you can put `900` and only 10% is shown. However if this slider is responsive (width: 70% or something) and becomes 800px, 500px, etc. width, then the `900` is obviously wrong.

With this tiny PR, one can put `'10%'` instead, so stretch will be relative based on the actual width. I hope this can be approved as it's probably useful in plenty of today's responsive sites. Thanks!


nolimits4web:
Merged, thanks!

Issue #3540, opened 2020-04-17
Error connecting project locally


IvanCen:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 5.3.7
* Platform/Target and Browser Versions: All platforms and browsers.

### What you did
The library started with an error when directly connecting the code to the project (locally by files)
![image](https://user-images.githubusercontent.com/54720503/79596391-284b9a00-80e9-11ea-890c-edd9d6e5b60d.png)
![image](https://user-images.githubusercontent.com/54720503/79597193-83ca5780-80ea-11ea-8e47-c08e3c553851.png)

PR #3541, opened 2020-04-17
Escaped characters in regular expression due to incorrect

IvanCen:
issue #3540

Unshielded regular expression characters lead to an error connecting to the project, is solved by escaping special characters


nolimits4web:
Thanks, i pushed the fix manually, since PRs to `packages` is not allowed

PR #3544, opened 2020-04-21
Fix focus ring rendering glitch due to inherited page line-height.

passiomatic:
**Before**:
Screenshot 2020-04-21 at 11 37 43

**After**:
Screenshot 2020-04-21 at 11 52 55

Looks like the focus ring has a rendering glitch due to inherited page `line-height`, that is likely > 1.0. Set the pseudo-element `line-height` to a neutral value of 1.0 fixes it. 
Tested with latest Safari, Chrome and Firefox.

### DIFF
--- a/src/components/navigation/navigation.less
+++ b/src/components/navigation/navigation.less
@@ -28,6 +28,7 @@
     letter-spacing: 0;
     text-transform: none;
     font-variant: initial;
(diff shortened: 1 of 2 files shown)


nolimits4web:
Merged, thanks!

PR #3550, opened 2020-04-24
Change aria-disabled attribute to disabled

davidbergmannverndale:
Disabled elements are being accessible via keyboard, in order to fix this I'm changing the aria-disabled attribute to disabled.

Screen Shot 2020-04-24 at 11 37 59


masi:
Great!

nolimits4web:
Thanks, but `aria-disabled` is not the same as `disabled`, so we need to keep both then `aria-disabled` and `disabled`

PR #3557, opened 2020-04-29
Add an event on hash change

1sergiogarciap:
When using nested swipers you can check if you have hash navigated to a slide of the nested swiper with the slideChange event and position the main swiper on the slide where the nested swiper is located. However if the nested swiper slide that is being hash navigated is the active slide, the SlideChange event does not happen, hence the need for the HashChange event in this particular case.


nolimits4web:
Merged, thanks!

Issue #3112, opened 2019-06-22
Lazyload for Picture element - it is possible?


7iomka:
I needed more responsive logic, and in reasons of browsers compatibility, I needed to use picture element, instead of simple image.
How to make lazy-loading by swiper to be worked? 
Currently it load only founded img tag :(
![image](https://user-images.githubusercontent.com/18473137/59966022-c51bc600-951e-11e9-8d31-41b361bc4177.png)

PR #3560, opened 2020-05-02
Add support for picture element lazy-loading

zhiru:
Add support for picture element lazy-loading.

I've made some changes on lazy.js to accept data-srcset like used on other pull request example.

Based on #3151 
Maybe solving #3096 #3112 

Test Use: https://aloje.com.br

### DIFF
--- a/src/components/core/images/loadImage.js
+++ b/src/components/core/images/loadImage.js
@@ -1,12 +1,14 @@
 import { window } from 'ssr-window';
+import $ from '../../../utils/dom';
 
 export default function (imageEl, src, srcset, sizes, checkForComplete, callback) {
   let image;
   function onReady() {
     if (callback) callback();
   }
-  const picture = $(imageEl).parent();
-  if (!imageEl.complete || !checkForComplete && (picture[0].tagName !== 'PICTURE')) {
+  const isPicture = $(imageEl).parent('picture')[0];
+
+  if (!isPicture && (!imageEl.complete || !checkForComplete)) {
     if (src) {
       image = new window.Image();
       image.onload = onReady;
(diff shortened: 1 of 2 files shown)


anthonycook:
Great job! Hope this get's merged soon, I've been using a third party lazy-loading plugin just to get it working with the picture element. 🙈

zhiru:
temporary i've used too, this [Lazyload Lib](https://github.com/verlok/lazyload), but i'm thinking of closing this and re-open, i find another adjust that refers to the same on problem.

i need other one to check if is correct and working, can you test the last two commits @anthonycook ?

[ Commits Swiper](https://github.com/zhiru/swiper/commits/master)

nolimits4web:
Thanks! I pushed some minor tweaks, seems to be works fine

JamesIves:
Is there an example of this available in documentation somewhere? From what I can tell you need to attach `class="swiper-lazy"` to each `source` element that is a child of `picture` too? (The following is using Angular so the `attr.` prefix is needed or it will render it as `srcset` directly. 

<picture class="storefront-banner__background-container--asset">
   <source class="swiper-lazy" media="(min-width: 300px)" attr.data-srcset="..." />
   <source class="swiper-lazy" media="(min-width: 600px)" attr.data-srcset="..." />
   
</picture>

maddocnc:
> Is there an example of this available in documentation somewhere? From what I can tell you need to attach `class="swiper-lazy"` to each `source` element that is a child of `picture` too? (The following is using Angular so the `attr.` prefix is needed or it will render it as `srcset` directly.
> 

Looks like you need data-srcset on source, and swiper-lazy class only on  it then takes it's parent <picture> and do the job.

PR #3570, opened 2020-05-08
Mousewheel scrolling problem for vertical slider

ibartsteam:
Fixed Mouse wheel scrolling problem for vertical slider

I was facing this issue for some time so I thought It would be better to fix the code.
only one line is edited so there won't be any problem


nolimits4web:
Thanks, merged manually with small tweak in https://github.com/nolimits4web/swiper/commit/07c3337b453f94186df00c68cc477577f01653d0

PR #3572, opened 2020-05-08
fix: only cancel cancelable touchmove events

pimlie:
We are receiving a reasonable amount of ReportingObserver errors on our swiper component. The error is:
```
ReportingObserver [intervention]: Ignored attempt to cancel a touchmove event with cancelable=false, for example because scrolling is in progress and cannot be interrupted
```

To be honest I am not sure how to test this as I havent been able to trigger this issue myself, but I used this SO as reference and the error it self also exactly explains what the issue is: https://stackoverflow.com/questions/26478267/touch-move-getting-stuck-ignored-attempt-to-cancel-a-touchmove

### DIFF
--- a/src/components/core/events/onTouchMove.js
+++ b/src/components/core/events/onTouchMove.js
@@ -102,10 +102,10 @@ export default function (event) {
     return;
   }
   swiper.allowClick = false;
-  if (!params.cssMode) {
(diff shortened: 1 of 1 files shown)


nolimits4web:
thanks, merged

PR #3586, opened 2020-05-19
fix sub-package.json

mhartington:
Hello! I wanted to follow up on an issue that came across ionic's issue tracker.

https://github.com/ionic-team/ionic/issues/21138

TL;DR, the update to `ssr-window` fixed some ssr bugs, but the package.json in the package dir did not get the ssr-window version bump.

This PR should address that.


nolimits4web:
Thanks! 👍

PR #3590, opened 2020-05-21
Fix in pagination detect uniqueNaVElements

TreVld:
Removed the restriction on one unique pagination element.

I would like to be able to use several instances of swiper on the page, in which there are two pagination elements (on top of the slide and below)


nolimits4web:
Merged, thanks!

PR #3598, opened 2020-05-27
Scale param for coverflow effect added.

SablinSergey:
Now you have possibility to use scale parameter in coverflow effect.


nolimits4web:
Merged, thanks!

PR #3619, opened 2020-06-14
Fix a typo in API.md

Ahmedfathyarafa:
This is a minor typo in "Slides grid" section in Swiper parameters.
Should be `properly ` rather than `property`.


nolimits4web:
Hi, this PR contains a lot of changes. If it is about type then it should be the only change

PR #3658, opened 2020-07-06
link unpkg.com don't work

korom:
Hello,

links for 
https://unpkg.com/swiper/css/swiper.min.css
https://unpkg.com/swiper/js/swiper.min.js

please install redirections !

thanks


korom:
thanks for docs but why not have set up redirects ?

Issue #3692, opened 2020-07-15
Incorrect type definition for Swiper react component.


kintesh:
This is a (multiple allowed):
* [ ] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 6.x.
* Platform/Target and Browser Versions: Typescript 3.8.3, Last two browser versions.

### What you did
Importing Swiper react component in a typescript project that does not set 'skipLibCheck' compile option to true results in the compiler throwing the below error.

### Expected Behavior
Typescript compiler should not complain.

### Actual Behavior

```
node_modules/swiper/swiper-react.d.ts:6:11 - error TS2430: Interface 'Swiper' incorrectly extends interface 'HTMLAttributes<HTMLElement>'.
  Types of property 'onProgress' are incompatible.
    Type '((swiper: Swiper, progress: number) => void) | undefined' is not assignable to type '((event: SyntheticEvent<HTMLElement, Event>) => void) | undefined'.
      Type '(swiper: Swiper, progress: number) => void' is not assignable to type '(event: SyntheticEvent<HTMLElement, Event>) => void'.

6 interface Swiper extends SwiperOptions {
            ~~~~~~
```

### Reason for this issue
This issue is cause by this line `interface Swiper extends React.HTMLAttributes<HTMLElement> {}` in swiper-react.d.ts

It extends the React.HTMLAttributes<HTMLElement> which defines properties like `onProgress` with the type `((event: SyntheticEvent<HTMLElement, Event>) => void) | undefined'`. The Swiper has the same property with type `((swiper: Swiper, progress: number) => void) | undefined'`.

PR #3694, opened 2020-07-15
Fix issue #3692 by omitting generic HTMLElement type properties that …

kintesh:
This issue is cause by this line interface Swiper extends React.HTMLAttributes<HTMLElement> {} in swiper-react.d.ts

It extends the React.HTMLAttributes which defines properties like onProgress with the type ((event: SyntheticEvent<HTMLElement, Event>) => void) | undefined'. The Swiper has the same property with type ((swiper: Swiper, progress: number) => void) | undefined'.

The fix is to omitting generic HTMLElement type properties that are defined in `swiper-events.d.ts` when extending the Swiper interface with React.HTMLAttributes<HTMLElement>.

### DIFF
--- a/src/swiper-react.d.ts
+++ b/src/swiper-react.d.ts
@@ -43,7 +43,11 @@ interface SwiperSlide {
   zoom?: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thanks!

kintesh:
Happy to help :)

PR #3720, opened 2020-07-23
Mousewheel: add threshold for wheel scroll event

timothyhague:
When scrolling, the carousel often jumps two pages of slides instead of one, due to inertia of the scroll. It's particularly noticeable with a mac trackpad.

To improve, I've added preventSwipeThresholdDelta and preventSwipeThresholdTime params to the mousewheel module, so that:

1. only scroll events over that delta threshold will be acknowledged.

and

2. only scroll events which happen after a threshold time in milliseconds will be acknowledged.

To make this work, you can add the following params to the config:

```
{
  slidesPerView: 3,
  slidesPerGroup: 3,
  spaceBetween: 20,
  direction: 'horizontal',
  speed: 500,
  mousewheel: {
    forceToAxis: true,
    // This only allows scrolls with a delta greater than 120 to be acknowledged. 
    preventSwipeThresholdDelta: 120,
    // This only allows scrolls once every 3 seconds to be acknowledged. 
    preventSwipeThresholdTime: 3000,
  },
}
```

### DIFF
--- a/src/components/mousewheel/mousewheel.js
+++ b/src/components/mousewheel/mousewheel.js
@@ -319,10 +319,18 @@ const Mousewheel = {
     const window = getWindow();
 
     if (
-      this.params.mousewheel.preventSwipeThreshold &&
-      newEvent.delta < this.params.mousewheel.preventSwipeThreshold
+      this.params.mousewheel.preventSwipeThresholdDelta &&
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thanks!

PR #3744, opened 2020-07-31
add some types of swiper-options for React

### LINKED ISSUE #3743: swiper/react missing declare some types
* [x] bug

* Swiper Version: 6.0.4
* Platform/Target and Browser Versions: macOS CHROME

### What you did
https://swiperjs.com/react/#usage
I tried to implement swiper using typescript and swiper/react referring to the above page.

**Note, Swiper React component will create required elements for Navigation, Pagination and Scrollbar if you pass these params without specifying its elements (e.g. without navigation.nextEl, pagination.el, etc.)**

It is supposed to be used like this, but boolean type is not declared in types.

### Actual Behavior
Can not compile typescript used by swiper/react.

ninpeng:
Fixes #3743 Added boolean type to swiper-options navigation, pagination and scrollbar props.

### DIFF
--- a/src/types/swiper-options.d.ts
+++ b/src/types/swiper-options.d.ts
@@ -601,10 +601,10 @@ export interface SwiperOptions {
   keyboard?: KeyboardOptions | boolean;
   lazy?: LazyOptions | boolean;
   mousewheel?: MousewheelOptions | boolean;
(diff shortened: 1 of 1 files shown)


nolimits4web:
Thanks, merged

Issue #3755, opened 2020-08-08
swiper/react virtual won't support RTL direction


IronKinoko:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 6.1.1
* Platform/Target and Browser Versions: 5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/84.0.4147.105 Safari/537.36
* Live Link or JSFiddle/Codepen or website with isssue: none

### What you did
At swiper 6.1.1 version, swiper/react virtual won't support RTL direction  
[swiper/react/virtual.js](https://github.com/nolimits4web/swiper/blob/master/src/react/virtual.js)

It may be because you forgot to check whether it is rtlTranslate and set `offsetProp` to `right`

### Expected Behavior
When I set the `dir='rtl'`, it will set the `offsetProp` to 'right' correctly

### Actual Behavior
At 6.1.1 version swiper/react virtual `offsetProp` always be 'left'

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](http://stackoverflow.com/questions/tagged/swiper)
for that.

Before you open an issue, please check if a similar issue already exists or has been closed before.

nolimits4web:
Fixed by #3756

PR #3756, opened 2020-08-08
fix: support react/virtual rtl direction

IronKinoko:
At version 6.1.1  swiper/react won't support RTL direction in virtual mode.

This RP is to fix this. #3755


nolimits4web:
Thanks, merged

Issue #3757, opened 2020-08-10
slideChange event callback function returns empty


Joshanity17:
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 6.1.1
* Platform/Target and Browser Versions: Chrome

### What you did
I am using [ngx-swipper](https://www.npmjs.com/package/ngx-swiper-wrapper). I am watching the slide changes through `slideChange` event. Add looking at the documentation, it has a callback parameter of 
```
slideChange: (swiper: Swiper) => void
```

I implemented this code with this

HTML
```
<div id="videoContainer">
    <swiper [config]="swiperConfig" [(index)]="swiperIndex" (slideChange)="swiperEvent($event)">
        <ng-container *ngFor="let video of videoSource">
            <video>
                <source src="{{video.path}}" type="video/mp4">
            </video>
        </ng-container>
    </swiper>
</div>
```

TS

```
public videoSource: Video[] = [];
public swiperConfig: SwiperConfigInterface = {
  direction: 'vertical',
  slidesPerView: 1,
  navigation: false
};
public swiperIndex: number = 0;

swiperEvent(data) {
  console.log(data)
}
```

What I am getting as `data` is `(0) []` which is an empty set of collection.

### Expected Behavior
I am expecting to get [Swiper](https://swiperjs.com/types/classes/_types_swiper_class_d_.swiper.html) class as a callback function parameter for the `slideChange` event.

### Actual Behavior
What I am getting as callback function parameter is `(0) []` which is an empty set of collection.

nolimits4web:
Swiper instance is there:
https://swiperjs.com/demos/010-default.html
```
swiper.on('slideChange', console.log)
```

PR #3758, opened 2020-08-10
Add swiper as parameter on slideChange callback function event

Joshanity17:
Fix for slideChange event callback function returns empty #3757, kindly see the issue that I raised for the whole details.


nolimits4web:
It is not needed, Swiper automatically `prepends` Swiper instance to all emits:
https://swiperjs.com/demos/010-default.html
swiper.on('slideChange', console.log)

Issue #3417, opened 2020-01-27
Don't use A11y when renderBullet is used


babielmam:
This is a (multiple allowed):
* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 5.2.1
* Platform/Target and Browser Versions: all
* Live Link or JSFiddle/Codepen or website with isssue: -/-

### What you did
I used the renderBullet function of the Pagination component to create a fully custom pagination bullet. I also enabled the A11y component.

### Expected Behavior
SwiperJS takes the string that I return, sets active classes and that's it.

### Actual Behavior
SwiperJS does what I expected, but also adds attributes for tabindex, role and aria-label, because I have enabled the A11y component.
This is unexpected and not wanted, when I use renderBullet. We use semantically meaningful HTML and thus don't need the tabindex and role. Aria-label as a convention for us is a last resort, we aim to give meaning to buttons through hidden text content. 

### Suggested solution
When renderBullet function is used to create pagination bullets, A11y should not transform the bullets. 
Alternatively you could also set the A11y options more granular, so that you can enable/disable it for specific components (such as the Pagination component).

babielmam:
Also, why should you be able to focus a bullet but not be able to hit Enter/Space in order to go to that slide? This is the default, when A11y and Keyboard components are used.

anneke:
Hi there, could I request to reopen this issue? I now feel forced to adapt the script myself to not add aria-label and role="button", because the code we use for the paginationBullets is using buttons (which do not need a `role="button"`, and we add a screenreader-only label which negates the need for an aria-label (and rule #1 of ARIA is.. do not use it, as I've learned). 

I would ask that if pagination/renderBullet is set, these properties (aria-label and role) are not set, and that in the docs this is reflected with a note and a warning (as I do realise that people who are not experienced with accessibility might harm the slider, not knowing).

nolimits4web:
Fixed by #3774

PR #3774, opened 2020-08-23
#3417 if pagination.renderbullet is set, don’t add role/aria-label

anneke:
Hi there,

This is a proposed fix for issue #3417. Please go easy on me, I hope this is useful as I don't want to raise an issue and not try t work towards a solution.

If pagination.renderbullet is set, (and the user wishes to implement their own html code to render each bullet), hand off responsibility of making that code accessible by not forcing `role="button"` and `aria-label` on it. Adding these unasked might even hurt the accessibility of what the user is trying to achieve.

I tested this locally and it worked as expected for me as summarized above!

### DIFF
--- a/src/components/a11y/a11y.js
+++ b/src/components/a11y/a11y.js
@@ -96,8 +96,8 @@ const A11y = {
       swiper.pagination.bullets &&
       swiper.pagination.bullets.length
     ) {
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thanks!

PR #3804, opened 2020-09-07
docs: Fix simple typo, momuntum -> momentum

timgates42:
There is a small typo in src/components/mousewheel/mousewheel.js.

Should read `momentum` rather than `momuntum`.


nolimits4web:
Thanks!

Issue #3149, opened 2019-07-15
Improve Accessibility Features


fregiese:
This is a (multiple allowed):
* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: v4.5.0
* Platform/Target and Browser Versions: EVERY PLATFORM

## Expected Behavior
The slider should implement some of the features and html structure of the w3c slider example: https://www.w3.org/TR/wai-aria-practices/examples/carousel/carousel-1/carousel-1.html
The w3c example also has a list of reasons why the features and html structure they used is accessible.
This would make SwiperJS a lot more accessible. Some of the accessibility enhancements can be set with options, but this as the default would help users who don't know how to make it more accessible. It would also be a good reason to use this library instead of other slider libraries, because SwiperJS would be a good option for everyone who wants to apply wcag 2.1 standards. https://www.w3.org/TR/WCAG21/

Some of the features could only be implemented in the accessibility module, for users who don't want to use them.

## Actual Behavior
SwiperJS is only partially accessible.

## Features to implement:
* [x] `<section>` as the outer container, instead of an `<div>` element
* [x] outer container has `aria-roledescription="carousel"`
* [x] outer container has `aria-label="Example Content"` (Label can be set like the next/prev slider buttons text)
* [x] next and previous buttons are `<a>`-tags
* [x] next and previous buttons have`role="button"`
* [x] next and previous buttons has `aria-controls="swiperID"` (this connects with the id of the content)
* [x] content container has `id="swiperID"` (this connects with the aria-attribute of the next and previous buttons)
* [x] content container has `aria-live="off"` when autoplay is on and `aria-live="polite"` when autoplay is off
* [x] content item has `role="group"`
* [x] content item has `aria-roledescription="slide"`
* [x] content item has the slide number as aria-label. Example:`aria-label="1 of 6"`
* [ ] Prevent tabbing to focusable items within out-of-view slides
* [ ] Make sure disabled controls/buttons are not focusable

## Related Issues
The w3c example doesn't use clones, but this issue is also important to improve accessibility: https://github.com/nolimits4web/swiper/issues/2929
Maybe the improved keyboard accessiblity would also fix issues like this: https://github.com/nolimits4web/swiper/issues/2945

aglotoff:
[The specification](https://www.w3.org/TR/wai-aria-practices/#carousel) for the carousel design pattern says: 
[...]

PR #3834, opened 2020-09-24
A11y enhancements

jenemde:
This pull request is related to https://github.com/nolimits4web/swiper/issues/3149.
It contains three new parameters for a11y:

- containerMessage (String)
- containerRoleDescriptionMessage (String)
- itemRoleDescriptionMessage (String)

Their default value is null, considering that the a11y community doesn't share a common opinion about aria-roledescription (s. https://github.com/nolimits4web/swiper/issues/3149).

There are some new functions in the a11y section:
- getRandomNumber(size), returning a random number, used to create a random id for the wrapper element.
- addElRoleDescription($el, description), adding the attribute "aria-roledescription" with value "description" to $el
- addElControls($el, controls), adding the attribute "aria-controls" with value "controls" to $el
- addElId($el, id), adding the attribute "id" with value "id" to $el
- addElLive($el, live), adding the attribute "aria-live" with value "live" to $el

In the code you can find the desired enhancement features:
- `<section>` as the outer container, instead of a `<div>` element: changed in playground/index.html
- outer container has aria-roledescription="carousel": changed in a11y (optional, default value is null)
- outer container has aria-label="Example Content" (Label can be set like the next/prev slider buttons text): changed in a11y (optional, default value is null)
- next and previous buttons are `<a>`-tags: changed in playground/index.html, but used `<button>`-tags instead of `<a>`-tags
- next and previous buttons have role="button": changed in a11y (role is only set if buttons have no `<button>`-tag)
- next and previous buttons has aria-controls="swiperID" (this connects with the id of the content): changed in a11y (relates to next feature)
- content container has id="swiperID" (this connects with the aria-attribute of the next and previous buttons): changed in a11y (a random id which is also used for aria-controls in next and previous buttons is generated)
- content container has aria-live="off" when autoplay is on and aria-live="polite" when autoplay is off: changed in a11y (checks autoplay status and sets aria-live accordingly)
- content item has role="group": changed in a11y
- content item has aria-roledescription="slide": changed in a11y (optional, default value is null)
- content item has the slide number as aria-label. Example :aria-label="1 / 6": changed in a11y

There is also a pull request for an according change in the documentation for swiper-website: https://github.com/nolimits4web/swiper-website/pull/74

### DIFF
--- a/src/components/a11y/a11y.js
+++ b/src/components/a11y/a11y.js
@@ -2,6 +2,10 @@ import $ from '../../utils/dom';
 import { bindModuleMethods } from '../../utils/utils';
 
 const A11y = {
+  getRandomNumber(size = 16) {
+    const randomChar = () => Math.round(16 * Math.random()).toString(16);
+    return 'x'.repeat(size).replace(/x/g, randomChar);
+  },
   makeElFocusable($el) {
     $el.attr('tabIndex', '0');
     return $el;
@@ -14,8 +18,24 @@ const A11y = {
     $el.attr('role', role);
     return $el;
   },
+  addElRoleDescription($el, description) {
+    $el.attr('aria-role-description', description);
+    return $el;
+  },
+  addElControls($el, controls) {
+    $el.attr('aria-controls', controls);
+    return $el;
+  },
   addElLabel($el, label) {
     $el.attr('aria-label', label);
+    return $el;
+  },
+  addElId($el, id) {
+    $el.attr('id', id);
+    return $el;
+  },
+  addElLive($el, live) {
+    $el.attr('aria-live', live);
     return $el;
   },
   disableEl($el) {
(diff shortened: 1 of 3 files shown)


src/components/a11y/a11y.js
```diff
@@ -111,11 +131,43 @@ const A11y = {
...
+    if (params.containerRoleDescriptionMessage) {
+      swiper.a11y.addElRoleDescription($containerEl, params.containerRoleDescriptionMessage);
+    }
+    if (params.containerMessage) {
+      swiper.a11y.addElLabel($containerEl, params.containerMessage);
+    }
+
+    // Wrapper
+    const $wrapperEl = swiper.$wrapperEl;
+    const randomWrapperId = `swiper-wrapper-${swiper.a11y.getRandomNumber(16)}`;
```
nolimits4web, line 149: I would suggest to reuse existing wrapper ID if it is there, something like:
```
const wapperId = $wapperEl.attr('id') || `swiper-wrapper-${swiper.a11y.getRandomNumber(16)}`;
```
Otherwise people may complain if their custom ID (if the have it for some reason) is overwritten

nolimits4web (commented):
Thanks for this PR, appreciate it! Added one comment about wrapper ID and can you please also add docs for new parameters to TypeScript definitions here https://github.com/nolimits4web/swiper/blob/master/src/types/components/a11y.d.ts ?

src/components/a11y/a11y.js
jenemde: Thanks for your kind and fast reaction.
You are right, I changed the part for the wrapperId as you suggested and added the docs for the new parameters to TypeScript definitions.

nolimits4web:
Merged, thanks!

PR #3868, opened 2020-10-12
Fix no swiping class in shadow copmonent

weiz18:
Problem 

`$targetEl.closest` does not find element with swiping class if the element is within a web component with shawroot. 

Fix 
Replace targetEl to be the first element in the event touch path

### DIFF
--- a/src/components/core/events/onTouchStart.js
+++ b/src/components/core/events/onTouchStart.js
@@ -31,7 +31,7 @@ export default function onTouchStart(event) {
     $targetEl = $(event.path[0]);
   }
 
-  if (params.noSwiping && ($targetEl.closest(params.noSwipingSelector ? params.noSwipingSelector : "." + params.noSwipingClass)[0] || shadowRootWithSwipingClass)) {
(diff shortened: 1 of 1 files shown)


src/components/core/events/onTouchStart.js
```diff
@@ -24,15 +24,18 @@ export default function onTouchStart(event) {
...
-    swiper.allowClick = true;
-    return;
+
+  // change target el for shadow root componenet
+  const swipingClassHasValue = !!params.noSwipingClass && params.noSwipingClass !== '';
+  if (swipingClassHasValue && e.target && e.target.shadowRoot && event.path && event.path[0]) {
+    $targetEl = $(event.path[0]);
   }
+
+  if (params.noSwiping && ($targetEl.closest(params.noSwipingSelector ? params.noSwipingSelector : "." + params.noSwipingClass)[0] || shadowRootWithSwipingClass)) {
```
nolimits4web, line 34: What is `shadowRootWithSwipingClass` ?

weiz18: removed. 🙏

nolimits4web:
Thanks, merged!

PR #3899, opened 2020-10-27
#3897 | Type-check the `index` argument

### LINKED ISSUE #3897: slideTo() / No type check for the `index` argument
This is a:
* [x] bug
* [ ] enhancement

---

* **Swiper Version**: 6.3.4
* **Platform/Target and Browser Versions**: (doesn't really matter) Google Chrome 88.0.4298.4 (Official Build) dev (x86_64) @ macOS Catalina 10.15.7 (19H2)
* **Live Link**: https://codesandbox.io/s/swiper-slideto-nan-ufzml

### What you did
I've called `Swiper.slideTo()` method. It occurred, that the argument I've been passing to the method was `NaN`. Of course it has been a direct result of the bug in my own code, but I'm sure that the provided error message could've been much better 🙂

### Expected behavior
The error message should say something like: `Swiper.slideTo(): Expected type 'number' or 'string', but got [NaN]`.

### Actual behavior
I've been given a `TypeError` with the following message: `Cannot read property 'classList' of undefined`.

staszek998:
Perform the type checking on the `index` argument before using its value inside the `slideTo()` method.

Fixes #3897.

### DIFF
--- a/src/components/core/slide/slideTo.js
+++ b/src/components/core/slide/slideTo.js
@@ -4,6 +4,37 @@ export default function slideTo(
   runCallbacks = true,
   internal,
 ) {
+  if (typeof index !== 'number' && typeof index !== 'string') {
+    throw new Error(
+      `The 'index' argument cannot have type other than 'number' or 'string'. [${typeof index}] given.`,
+    );
+  }
+
+  if (typeof index === 'string') {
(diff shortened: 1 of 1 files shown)


nolimits4web:
Thanks, merged

PR #3909, opened 2020-10-30
fix(types): add null types

### LINKED ISSUE #3866: TypeScript: NavigationOptions in navigation.d.ts can't take null as value for nextEl/prevEl
This is a (multiple allowed):
* [x] bug
* [x] enhancement

* Swiper Version: 6.3.3

### What you did

I wanted to assign a useRef(null) Hook for a NavigationOptions configuration:

```typescript
navigation={{
  prevEl: prevEl.current,
  nextEl: nextEl.current,
}}
```

### Expected Behavior

That I could assign null to prevEl and nextEl, especially as it is hinted in the [Type Definition](https://github.com/nolimits4web/swiper/blob/master/src/types/components/navigation.d.ts):

```typescript
export interface NavigationOptions {
  /**
   * String with CSS selector or HTML element of the element that will work
   * like "next" button after click on it
   *
   * @default null
   */
  nextEl?: CSSSelector | HTMLElement;

  /**
   * String with CSS selector or HTML element of the element that will work
   * like "prev" button after click on it
   *
   * @default null
   */
  prevEl?: CSSSelector | HTMLElement;
```

### Actual Behavior

This causes a TypeScript Error: `Type 'null' is not assignable to type 'HTMLElement | CSSSelector | undefined'.ts(2322)`

I usually apply the following patch after `yarn install`, but I could also create a Pull Request if you prefer that.

```patch
diff --git a/node_modules/swiper/types/components/navigation.d.ts b/node_modules/swiper/types/components/navigation.d.ts
index 5b93fcc..58902d0 100644
--- a/node_modules/swiper/types/components/navigation.d.ts
+++ b/node_modules/swiper/types/components/navigation.d.ts
@@ -36,7 +36,7 @@ export interface NavigationOptions {
    *
    * @default null
    */
-  nextEl?: CSSSelector | HTMLElement;
+  nextEl?: CSSSelector | HTMLElement | null;
 
   /**
    * String with CSS selector or HTML element of the element that will work
@@ -44,7 +44,7 @@ export interface NavigationOptions {
    *
    * @default null
    */
-  prevEl?: CSSSelector | HTMLElement;
+  prevEl?: CSSSelector | HTMLElement | null;
 
   /**
    * Toggle navigation buttons visibility after click on Slider's container
```

vltansky:
This PR adds `null` types. Example:

```typescript

  /**
   * String with CSS selector or HTML element of the element that will work
   * like "prev" button after click on it
   *
   * @default null
   */
  prevEl?: CSSSelector | HTMLElement;
```

Optional character `?` tells Typescript that it accepts `undefined` but not null, but according to the `@default` null is acceptable.
So `null` type is missing here and in places like this. 

closes #3866

### DIFF
--- a/src/types/components/a11y.d.ts
+++ b/src/types/components/a11y.d.ts
@@ -57,19 +57,19 @@ export interface A11yOptions {
    *
    * @default null
    */
-  containerMessage?: string;
+  containerMessage?: string | null;
 
   /**
    * Message for screen readers describing the role of outer swiper container
    *
    * @default null
    */
-  containerRoleDescriptionMessage?: string;
+  containerRoleDescriptionMessage?: string | null;
 
   /**
    * Message for screen readers describing the role of slide element
    *
    * @default null
    */
-  itemRoleDescriptionMessage?: string;
+  itemRoleDescriptionMessage?: string | null;
 }
--- a/src/types/components/mousewheel.d.ts
+++ b/src/types/components/mousewheel.d.ts
@@ -58,12 +58,12 @@ export interface MousewheelOptions {
    *
    * @default null
    */
-  thresholdDelta?: number;
+  thresholdDelta?: number | null;
 
   /**
    * Minimum mousewheel scroll time delta (in ms) to trigger swiper slide change
    *
    * @default null
    */
-  thresholdTime?: number;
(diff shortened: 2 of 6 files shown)


nolimits4web:
Thanks!

PR #3913, opened 2020-11-01
feat: support multiple classes in pagination buttonClass

### LINKED ISSUE #3905: Option to Handle Multiple Class through bulletClass and bulletActiveClass
I posted a question to [**StackOverflow**](https://stackoverflow.com/q/64548304/2637261); however, I got no response on this. I tried to check the code as well and I think this is not being handled currently.

This is a (multiple allowed):
* [ ] bug
* [x] enhancement

* Swiper Version: **6.3.3**
* Platform/Target and Browser Versions: **Chrome** (But I believe this can be replicated anywhere)
* Live Link or JSFiddle/Codepen or website with isssue: https://stackblitz.com/edit/react-ts-sa4ex7?file=index.tsx

### What you did
I tried to add multiple class to the `bulletClass` property of `pagination`. 
```
pagination={{
    clickable: true,
    bulletClass: classNames('swiper-pagination-bullet', style['feature-pagination']),
    bulletActiveClass: classNames('swiper-pagination-bullet-active', style['active'])
}}
```
I am trying to retain the default class of the bullets, while adding my custom styles. Of course I could have easily add my class globally, but I am working with `CSS Modules`, so I want this class only implemented to specific components.

### Expected Behavior
Expecting to all of the classes are implemented, and the functional use case will be retain.

### Actual Behavior
The styles has been applied; however, the function of pagination is not working (e.g. out of the box changing of active class, clickable pagination)

I believe the reason behind this is because of the way `swiper` selects the class. In this [line of code](https://github.com/nolimits4web/swiper/blob/master/src/components/pagination/pagination.js#L218), it returns `null` since it cannot handle multiple class.

vltansky:
closes #3905

### DIFF
--- a/src/components/a11y/a11y.js
+++ b/src/components/a11y/a11y.js
@@ -71,7 +71,10 @@ const A11y = {
         swiper.a11y.notify(params.prevSlideMessage);
       }
     }
-    if (swiper.pagination && $targetEl.is(`.${swiper.params.pagination.bulletClass}`)) {
+    if (
+      swiper.pagination &&
+      $targetEl.is(`.${swiper.params.pagination.bulletClass.replace(/ /g, '.')}`)
+    ) {
       $targetEl[0].click();
     }
   },
@@ -205,7 +208,7 @@ const A11y = {
     ) {
       swiper.pagination.$el.on(
         'keydown',
-        `.${swiper.params.pagination.bulletClass}`,
+        `.${swiper.params.pagination.bulletClass.replace(/ /g, '.')}`,
         swiper.a11y.onEnterKey,
       );
     }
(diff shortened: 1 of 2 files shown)


nolimits4web:
Merged, thanks!

Devotee007:
I'm looking for a way to have different colours on the pagination for each slide. I want some to have a blue active dot and some a red active dot when active. Is this above a way to achieve that?

vltansky:
@Devotee007 just use CSS.
You can get help in [discussions](https://github.com/nolimits4web/swiper/discussions) or [StackOverflow](https://stackoverflow.com/)

Devotee007:
@vltansky, thanks for the tip, I have started a thread over in the discussions board now. I have used CSS for a custom colour, but I want to have a second class so that I can render a slider with different colour on the pagination. I don't know how I tell the different sliders a part.

PR #3918, opened 2020-11-03
feat: Angular support

vltansky:
Hi, I would be happy to contribute to swipers Angular support.  
Especially as the most popular Angular swiper library has [End of life notice](https://github.com/zefoy/ngx-swiper-wrapper#eol-notice).
This PR far away from to be ready. But many functionalities implemented, I'm trying to have the same API as React, Vue, and Svelte.
There are a lot of types that were missing, I've added them, but without description (as I don't know what to write there).
Also right now I'm importing everything from the `build` directory, as a temporary solution.
Currently working on virtual slides implementation.

I would like to know if it's welcomed and if yes I would really appreciate your help and review. I understand if you will not want to include it, I'll make from it a standalone wrapper then.


nolimits4web:
Thanks, I am for it with all hands, it was requested a lot of times. But again, I have close to 0 experience with Angular, and if you will help with issues with it in future then let's make it.

As I understood Angular uses template rendering which in my understanding is not so flexible as in React/Vue where they use render functions, and allow such powerful features as cloning and duplication children slides (for example for Virtual Slides and Loop mode). If it is true, then you probably should use Svelte's version for this as a starting point, it is much more simpler with less features.

vltansky:
Yes, of course, I'll help to maintain the Angular version. Meantime I'm trying to help in issues too, in order to understand swiper's internal code & potential issues better.

I'll make a separate PR for types, so you will be able to check if there are mistakes there / fulfill descriptions and etc.

I'm looking at Svelte and a bit at Vue as an example. Yes, Angular templating less flexible than Vue & React. But currently, I'm trying to implement virtual slides as it is in Vue (have some good progress with it).
Will be glad if you will play with the demo of Angular, and say what's missing or should be implemented another way in order to make some kind of roadmap.

vltansky:
Hi, I've made some progress with virtual slides. Looks like it works.
Can you play with the demo and share your thoughts?
`npm run build:dev` (currently angular uses `build` folder for types and etc)
`npm run angular:dev`

wilson-young:
Hi, @vltansky 
is it useable now? or soon?
I am doing a project and trying to use swiperjs but it's really hard to work with for now.

and really thank you for doing it for angular.
respect!

vltansky:
@wilson-young no, it's under development. You shouldn't use it right now.
For now better use this: https://github.com/zefoy/ngx-swiper-wrapper

wilson-young:
> @wilson-young no, it's under development. You shouldn't use it right now.
> For now better use this: https://github.com/zefoy/ngx-swiper-wrapper

Ok sir,
thanks for hard work
can't wait to see official angular supported Swiper

nolimits4web:
> @nolimits4web I've made some progress with virtual slides. Looks like it works.
> Can you play with the demo and share your thoughts?

@vltansky thanks, I will have time to look deeper on a weekend

nolimits4web:
Hey @vltansky, the demo looks good 👍 The only thing I noticed in Virtual demo is that Slides are missing -prev/-active/-next classes. I guess you need to listen for `_slideClass` event to update it correctly like it is done in React/Vue/Svelte

vltansky:
Okay, will work on this in nearest days.
I'm importing everything from build folder right now. I need or to import somehow from src, or to move angular to build folder. So the imports will be relative. Do you have any idea for the best solution?

nolimits4web:
> I need or to import somehow from src, or to move angular to build folder. So the imports will be relative. Do you have any idea for the best solution?

Good question 😄 Doesn't it work with relative imports inside of `src`? I mean if you will import Swiper core from same src folder?

vltansky:
> Hey @vltansky, the demo looks good 👍 The only thing I noticed in Virtual demo is that Slides are missing -prev/-active/-next classes. I guess you need to listen for `_slideClass` event to update it correctly like it is done in React/Vue/Svelte

Done. Will work now on `loop` mode. Is `loop` the last thing to finish or am I missing something?
I will leave imports for the end.

nolimits4web:
@vltansky Yes, loop is left. Also check it works/re-renders/updated on breakpoint change. And slideData like https://swiperjs.com/react/#swiperslide-render-function or https://swiperjs.com/vue/#swiperslide-slot-props or https://swiperjs.com/svelte/#swiperslide-slot-data (if something similar is achievable with Angular)

vltansky:
slideData done. 
Not sure about something. Take a look on example.
Do you prefer providing all variables as `slideData` / every variable separately / both (now both) ?
```html

    <swiper-slide>
      <ng-template swiper-slide-content let-active="isActive">
        test 1 {{ active ? 'active' : 'nope' }}
     </ng-template>
    </swiper-slide>
    <swiper-slide>
      <ng-template swiper-slide-content let-somename="isActive" let-next="isNext">
        test 2 {{ somename ? 'active' : 'nope' }} {{ next }}
     </ng-template>
    </swiper-slide>
    <swiper-slide>
      <ng-template swiper-slide-content let-active="isActive" let-data="slideData">
        test 3 {{ active ? 'active' : 'nope' }} {{ data.isNext }}
     </ng-template>
    </swiper-slide>
```

nolimits4web:
Let's keep it in align with other frameworks and just use single data object, like this I guess:
```
<ng-template swiper-slide-content let-data="slideData">
        test 3 {{ data.isActive ? 'active' : 'nope' }} {{ data.isNext }}
</ng-template>
```

vltansky:
Got stuck with the loop.
Looks like you cant clone `<ng-content>` in Angular
https://stackoverflow.com/questions/51104405/angular-cloning-ng-content-elements-and-its-functionality

Although works fine with `<ng-template>`, but that's forces to use this markup:
```html
<swiper>
 <swiper-slide>
     <ng-template swiper-slide-content let-data>1 {{data.isNext}}</ng-template>
   </swiper-slide>
 <swiper-slide>
     <ng-template swiper-slide-content let-data>2 {{data.isNext}}</ng-template>
  </swiper-slide>
</swiper>
```

Does somebody have any ideas?

nolimits4web:
> but you can see in the example that buttons on duplicated slides are not working.

Yes, this is because Swiper core duplicated slides and it has no idea (or angular) about those buttons. Svelte also doesn't allow to duplicate elements so I left duplication to core.

I think if it is possible in Angular with a bit of extra markup then let's better go with it as it should work better

vltansky:
But do we want to force using `ng-template`? Or just to warn that loop is working only with `ng-template`?

vltansky:
@nolimits4web how loop should work with virtual slides?
https://stackblitz.com/edit/swiper-demo-40-virtual-slides-dn7613?file=index.html

nolimits4web:
> But do we want to force using ng-template? Or just to warn that loop is working only with ng-template?

I think we should use `ng-template` only when `loop` enabled and cover it in docs, that with loop mode we need to use `ng-template`

vltansky:
Maybe better to force the use of `ng-template` to prevent issues with `loop` (I guess this is a popular option)?
This kind of markup is possible:
```html
<swiper>
    <ng-container *ngFor="let slide of virtualSlides">
      <ng-template swiperSlide let-anyName>Slide {{ slide }} {{anyName.isActive}}</ng-template>
    </ng-container>
    <ng-template swiperSlide let-data>Slide {{data.isActive}}</ng-template>
    <ng-template swiperSlide let-slideData>Slide {{slideData.isActive}}</ng-template>
</swiper>
```
The only downside that you cant use `*ngFor` on `ng-template`, so you will need to wrap it in `ng-container`.

nolimits4web:
To be honest not sure, but having `<swiper-slide>` just looks more clean for me (for the case we don't use loop)

vltansky:
It is cleaner, but I think `ng-template` is better here:
* Will prevent issues with loop mode
* `slideData` are available (no need for extra `ng-template`)
* As a result of above 2, as a user you don't need to remember markups for all use cases

I just pushed a version with following markup:
```html
<swiper
    #swiperRef
    [navigation]="navigation"
    (swiper)="log('swiper')"
    (slideChange)="log('slideChange')"
    [slidesPerView]="3"
    [spaceBetween]="50"
    [scrollbar]="scrollbar"
    [loop]="true"
    [pagination]="{ type: 'fraction' }"
  >
    <ng-template swiperSlide>Best slide ever 1</ng-template>
    <ng-template swiperSlide>Best slide ever 2</ng-template>
    <ng-template swiperSlide>Best slide ever 3</ng-template>
    <ng-container *ngFor="let slide of slides>
      <ng-template swiperSlide let-data>
        {{ slide }} / active: {{ data.isActive }} / next: {{ data.isNext }}</ng-template
      >
    </ng-container>
  </swiper>
```

vltansky:
@nolimits4web  I'm not sure about something.
When breakpoint param changes I was just updating param value on `swiperRef` as in vue/reacts/svelte.
But nothing happened, slider didnt change layout (i changed slidesPerView). `swiperRef.update()` didnt help.
In the end the solution I found is: 
```js
this.swiperRef.currentBreakpoint = null;
this.swiperRef.setBreakpoint();
```
How is it working in other frameworks (is it working?)?
If yes what am I missing?

vltansky:
As for me, it looks like we can plan some alpha/`next` tag release soon?
Would be happy if you could take a closer look and see if something important missing.
Also, we should decide about markup.

nolimits4web:
> @nolimits4web I need help here with breakpoints.
When breakpoint param changes I was just updating param value on swiperRef as in vue/reacts/svelte.
But nothing happened, slider didnt change layout (i changed slidesPerView). swiperRef.update() didnt help.
In the end the solution I found is:

Ok, it can be a little bit tricky. So here is how Core version works.
- on window resize it recalculates and update swiper parameters based on breakpoint (window size)
- on same resize handler swiper internals check parameters and update layout/slides's sizes, grid, etc. based on parameters

This may not work correctly in React/Vue/Svelte, because they don't know (don't watch for internal parameters) and can override something that was set by Core based on swiper props (let's call it external parameters).

In Swiper React/Vue/Svelte i am using a hack forcing to update swiper component right breakpoints changes. I have a dummy internal state property `breakpointChanged` (like [here](https://github.com/nolimits4web/swiper/blob/master/src/react/swiper.js#L26)) I just update this property within `_beforeBreakpoint` event (like [here](https://github.com/nolimits4web/swiper/blob/master/src/react/swiper.js#L76-L78)). This property does nothing and doesn't do anything but it is enough to trigger update of component in React/Vue/Svelte versions.

> Also, we should decide about markup.

If it is a common practice in Angular land, then let's go with unified `<ng-template>` syntax for all the cases.

> As for me, it looks like we can plan some alpha/next tag release soon? 
> Would be happy if you could take a closer look and see if something important missing.

I think most of things looks good. So we can just release it, let's say, as 6.4.0 update (without beta/next tag)

vltansky:
I think we are talking about different things in the breakpoints subject.
I don't have a problem updating the view. And it works fine when I'm changing the breakpoint that I'm not on currently.
See JSON in the page footer.
Example changing 768 breakpoints (we are initially at >1024):
https://skr.sh/v5VjUL3nbBZ?a

Here is an example of changing the current breakpoint (in the env of the video I'm uncommenting two lines that made this work for me)
https://skr.sh/v5VRHgiWzdl?a

> If it is a common practice in Angular land, then let's go with unified <ng-template> syntax for all the cases.

I think it's fine. Anyway, we always can add this functionality if it will be demanded.

> I think most of things looks good. So we can just release it, let's say, as 6.4.0 update (without beta/next tag)

I still have problems with imports. I've tried to solve it with `tsconfig` [path](https://github.com/nolimits4web/swiper/pull/3918/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0R15-R18), but when using this as a node_module getting this warning from webpack:
```
Warning: Entry point 'angular' contains deep imports into 'D:/projects/tailwind-test/node_modules/swiper/types/components/a11y', 'D:/projects/tailwind-test/node_modules/swiper/types/components/autoplay', 'D:/projects/tailwind-test/node_modules/swiper/types/components/controller', 'D:/projects/tailwind-test/node_modules/swiper/types/components/effect-coverflow', 'D:/projects/tailwind-test/node_modules/swiper/types/components/effect-cube', 'D:/projects/tailwind-test/node_modules/swiper/types/components/effect-fade', 'D:/projects/tailwind-test/node_modules/swiper/types/components/effect-flip', 'D:/projects/tailwind-test/node_modules/swiper/types/components/hash-navigation', 'D:/projects/tailwind-test/node_modules/swiper/types/components/history', 'D:/projects/tailwind-test/node_modules/swiper/types/components/keyboard', 'D:/projects/tailwind-test/node_modules/swiper/types/components/lazy', 'D:/projects/tailwind-test/node_modules/swiper/types/components/mousewheel', 'D:/projects/tailwind-test/node_modules/swiper/types/components/navigation', 'D:/projects/tailwind-test/node_modules/swiper/types/components/pagination', 'D:/projects/tailwind-test/node_modules/swiper/types/components/scrollbar', 'D:/projects/tailwind-test/node_modules/swiper/types/components/thumbs', 'D:/projects/tailwind-test/node_modules/swiper/types/components/virtual', 'D:/projects/tailwind-test/node_modules/swiper/types/components/zoom', 'D:/projects/tailwind-test/node_modules/swiper/types/swiper-events'. This is probably not a problem, but may cause the compilation of entry points to be out of order.
```
Will look for another solution...

nolimits4web:
Ok, you mean to change breakpoints object. Actually React/Vue/Svelte doesn't support it at all :) And this:
```
this.swiperRef.currentBreakpoint = null;
this.swiperRef.setBreakpoint();
```
yes, looks good to me in this case. I guess we need to add it to the React/Vue/Svelte as well then.

As for imports, I guess this can be a solution:

- build project to `package/` with the imports config that works
- add `scripts/build-angular.js` node script that will read angular package files and fix/change the imports

Can this work?

vltansky:
Ok, done as you proposed. Seems ready for me..?

vltansky:
Didn't work.. getting this error when trying to use swiper/angular:
```
Error: The target entry-point "angular" has missing dependencies:
 - ../core
```
I actually searched for the deep imports subject, and it looks like native script using deep imports just the same way:
```
Warning: Entry point '@nativescript/angular' contains deep imports into '/yourproject/node_modules/@nativescript/core .......
```

Maybe the deep imports solution is not so bad?

https://github.com/NativeScript/nativescript-angular/wiki/Updating-and-developing-for-@nativescript-angular-v9-with-Ivy-or-without#problem-2

vltansky:
Imports - done.
as for me, it's ready to be merged.
Take a look at the last commits, I needed to add few files (as `types.d.ts`, `swiper-angular.ts`, `core/index.js`), hope it's fine.
I believe that it's not perfect, but good enough for the first version.

nolimits4web:
Merged 🎉

wilson-young:
> Merged 🎉

so is it usable now? haha
btw thanks for hardwork!!!

nolimits4web:
It is published on npm as 6.4.0 version

vltansky:
Tried to use 6.4.0 version: getting warning about deep imports from webpack again...
```
Warning: Entry point 'angular' contains deep imports into 'D:/projects/tailwind-test/node_modules/swiper/types'. This is probably not 
a problem, but may cause the compilation of entry points to be out of order.
```

Although now only one. Don't understand how I missed it yesterday. It's not critical, everything works. But I'll make a PR with a fix to this.

vltansky:
@nolimits4web
1. Do we want to include CSS files with component? or developers should import it manually like in skeleton-elements?

2. I think we should support some of the `ngx-swiper-wrapper` functionality, so it will be easier to migrate for current users and also it just looks useful?
I propose to add:
* [global config](https://github.com/zefoy/ngx-swiper-wrapper#load-the-module-for-your-app-with-global-configuration)
```js
import { SwiperModule } from 'swiper/angular';
import { SWIPER_CONFIG } from 'swiper/angular';
import { SwiperConfigInterface } from 'ngx-swiper-wrapper';

const DEFAULT_SWIPER_CONFIG: SwiperConfigInterface = {
  direction: 'horizontal',
  slidesPerView: 'auto'
};

@NgModule({
  ...
  imports: [
    ...
    SwiperModule
  ],
  providers: [
    {
      provide: SWIPER_CONFIG,
      useValue: DEFAULT_SWIPER_CONFIG
    }
  ]
})
```
* setting params via config, input&output slideIndex variable `[(index)]`, disabled state

```html
<swiper [config]="config" [(index)]="index" [disabled]="false"></swiper>
```

* [functions](https://github.com/zefoy/ngx-swiper-wrapper#available-control--helper-functions-provided-by-the-directive):
```js
swiper()                          // Returns reference to the Swiper instance.

init()                            // Starts Swiper (when init set to false).
update()                          // Updates Swiper elements / classes / etc.

getIndex(real?)                   // Returns the active or real slide index.
setIndex(index, speed?, silent?)  // Runs transition to slide with given index.

nextSlide(speed?, silent?)        // Runs transition to the next slide index.
prevSlide(speed?, silent?)        // Runs transition to the previous slide index.

stopAutoplay(reset?)              // Stops and optionally resets the autoplay.
startAutoplay(reset?)             // Starts and optionally resets the autoplay.
```


What do you think?

nolimits4web:
> Do we want to include CSS files with component? or developers should import it manually like in skeleton-elements?

I think having it separated gives a bit more flexibility like in React/Svelte/Vue versions, especially when we can import only styles for selected modules (Navigation, Pagination) and not the whole bundle

nolimits4web:
> I think we should support some of the ngx-swiper-wrapper functionality, so it will be easier to migrate for current users and also it just looks useful?

You mean to extend defaults for all Swipers? To be honest I am not sure it is a pretty common/popular usecase. And it can be done much easier with core:
```js
import Swiper from 'swiper';

Swiper.extendDefaults({
  direction: 'vertical',
  // ...
}),
```

vltansky:
> I think having it separated gives a bit more flexibility like in React/Svelte/Vue versions, especially when we can import only styles for selected modules (Navigation, Pagination) and not the whole bundle

Ok

> You mean to extend defaults for all Swipers? To be honest I am not sure it is a pretty common/popular usecase. And it can be done much easier with core:

Make sense.

What about the rest of things?


* setting params via config, input&output slideIndex variable `[(index)]`, disabled state

```html
<swiper [config]="config" [(index)]="index" [disabled]="false"></swiper>
```

* [functions](https://github.com/zefoy/ngx-swiper-wrapper#available-control--helper-functions-provided-by-the-directive):
```js
swiper()                          // Returns reference to the Swiper instance.

init()                            // Starts Swiper (when init set to false).
update()                          // Updates Swiper elements / classes / etc.

getIndex(real?)                   // Returns the active or real slide index.
setIndex(index, speed?, silent?)  // Runs transition to slide with given index.

nextSlide(speed?, silent?)        // Runs transition to the next slide index.
prevSlide(speed?, silent?)        // Runs transition to the previous slide index.

stopAutoplay(reset?)              // Stops and optionally resets the autoplay.
startAutoplay(reset?)             // Starts and optionally resets the autoplay.
```

should we move the discussion to issue?

PR #3920, opened 2020-11-03
fix: add missing types

### LINKED ISSUE #3912: Why NavigationMethods doesn't export init() and destory()?
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: 6.3.4
* Platform/Target and Browser Versions: Chrome or Firefox. Not specific.
* Live Link or JSFiddle/Codepen or website with isssue: Nothing.

### Expected Behavior
I expect that typed file exports NavigationMethods is containing `init()` and `destory()`.

Did I miss something? Are the methods like private? Sorry for this, I'm kind of new to reading program.

vltansky:
closes #3912
related #3918

### DIFF
--- a/src/types/components/controller.d.ts
+++ b/src/types/components/controller.d.ts
@@ -1,6 +1,8 @@
 import Swiper from '../swiper-class';
 
-export interface ControllerMethods {}
+export interface ControllerMethods {
+  control?: Swiper;
+}
 
 export interface ControllerEvents {}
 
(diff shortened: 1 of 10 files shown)


src/types/components/controller.d.ts
```diff
@@ -1,6 +1,8 @@
 import Swiper from '../swiper-class';
 
-export interface ControllerMethods {}
+export interface ControllerMethods {
+  control?: Swiper;
```
vltansky, line 4: Is it the right place for it?
e.g Vue:
```js
if (needControllerInit) {
    swiper.controller.control = currentParams.controller.control;
  }
```

nolimits4web:
Thanks! I added some docs and moved keyboard and mouse wheel events to relevant modules

Issue #3895, opened 2020-10-27
Keyboard navigation with onlyInViewport: true triggers slide change in hidden swiper


acromatica:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: v6.3.4
* Platform/Target and Browser Versions: macOS 10.15.7 Google Chrome 86.0.4240.80
* Live Link or JSFiddle/Codepen or website with isssue: https://codepen.io/acromatica/pen/jOrBeex

### What you did
I added a swiper with enabled keyboard control and onlyInViewport parameter set to true. After that I've hidden the swiper, but I still see slideChange event being triggered when I press arrow keys.

### Expected Behavior
Slides shouldn't change when the swiper is hidden and onlyInViewport parameter set to true.

### Actual Behavior
slideChange event is being triggered when arrow keys are pressed.

vltansky:
fixed by #3922

waleedart:
Dear Sir,
Please unsubscribe me from this list.
Thanks,
Waleed
    On Saturday, January 23, 2021, 09:20:55 PM GMT+4, Vlad Tansky <notifications@github.com> wrote:  
 
 


Closed #3895.

PR #3922, opened 2020-11-04
Ignore the origin [0,0] inView.

ygj6:
Fix issue #3895, by ignore the origin [0,0] inView.
[0,0] not inView could avoid keyboard trigger slide change in hidden swiper.

When hidden swiper , the swiperCoord was:
```
0: (2) [0, 0]
1: (2) [NaN, 0]
2: (2) [0, NaN]
3: (2) [NaN, NaN]
```

### DIFF
--- a/src/components/keyboard/keyboard.js
+++ b/src/components/keyboard/keyboard.js
@@ -69,6 +69,7 @@ const Keyboard = {
       for (let i = 0; i < swiperCoord.length; i += 1) {
         const point = swiperCoord[i];
         if (point[0] >= 0 && point[0] <= windowWidth && point[1] >= 0 && point[1] <= windowHeight) {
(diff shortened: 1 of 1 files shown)


nolimits4web:
Thanks! Merged

domstubbs:
I’m reluctant to submit a PR without doing more testing, but for the project I’m working on currently this change isn’t preventing slides inside an inactive tab (`display: none`) from changing when arrow keys are pressed.

If I swap `continue` for `break` it all works as anticipated.

Before:

    if (point[0] === 0 && point[1] === 0) continue; // eslint-disable-line

After:

    if (point[0] === 0 && point[1] === 0) break; // eslint-disable-line

The swiperCoord I’m seeing for the hidden element looks like:

```
0: (2) [0, 0]
1: (2) [1051, 0]
2: (2) [0, 24]
3: (2) [1051, 24]
```

PR #3925, opened 2020-11-05
fix(pagination): bullet class mistake

vltansky:
I've made a mistake in PR #3913 
Extra braces after the class name:
```js
`.${params.bulletClass.replace(/ /g, '.')}}<-- two braces instead of one
```
this PR fixes it.


nolimits4web:
Thanks!

PR #3927, opened 2020-11-06
fix a typo in updateSize()

steve3d:
I believe this is a real typo, when detecting if there is a height in options, you always test height and width, then without the width in options, this will always fall, and there are so many issues with this height problems.


nolimits4web:
Thanks!

PR #3932, opened 2020-11-07
fix(lazy): update slider sizes after lazy image loaded

### LINKED ISSUE #3929: Lazy Loading Swiper Jumps
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: 6.3.5
* Platform/Target and Browser Versions: osx firefox 83.0
* Live Link or JSFiddle/Codepen or website with isssue: https://stackblitz.com/edit/lazy-loading-swiper-bug?file=index.js

### What you did
Create a swiper with lazy loading. Offset of the images is wrong, unless I resize the window when everything is loaded. ( The same appears when adding a video to a slide).

### Expected Behavior
Correct Image offset, also with `{ centeredSlides: true }`

### Actual Behavior
offset is wrong

vltansky:
closes #3929 
This is the easiest way to fix the issue, but maybe there is a more efficient way. Take a look at the issue for more details.

### DIFF
--- a/src/components/lazy/lazy.js
+++ b/src/components/lazy/lazy.js
@@ -91,9 +91,7 @@ const Lazy = {
           }
         }
         swiper.emit('lazyImageReady', $slideEl[0], $imageEl[0]);
-        if (swiper.params.autoHeight) {
-          swiper.updateAutoHeight();
-        }
+        swiper.update();
       });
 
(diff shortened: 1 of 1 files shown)


nolimits4web:
I am not sure we need it automatically as it is only makes sense for the case mentioned in the original issue. Otherwise it will do extra `update()` when it is unnecessary which is considered as heavy operation. So I would propose to use the solution you mentioned in the issue

Issue #3816, opened 2020-09-11
Sometimes when I click on a pagination bullet, the slider goes to a wrong slide


SilencerWeb:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 6.2.0 (of course, if your demos are using the latest version)
* Platform/Target and Browser Versions: macOS 10.15.6, Chrome 85.0.4183.102 (Official Build) (64-bit)
* Live Link or JSFiddle/Codepen or website with isssue: [Demo](https://swiperjs.com/demos/070-pagination-custom.html), [Source Code](https://github.com/nolimits4web/Swiper/blob/master/demos/070-pagination-custom.html)

### What you did
Sometimes when I click on a pagination bullet, the slider goes to a wrong slide. [Here](https://streamable.com/h45igg) you can see the video

### Expected Behavior
The slider should go to a correct slide

### Actual Behavior
The slider went to a wrong slide

pam81:
Hi, I did same test but only clicking quicly between 2 items ( page 3 and 4) and it moves to page 2 in a moment. 
It is lossing the real Index when you click wuicly between pages. 
Also I try to change default speed from 300 to 600 but the issue is still happening.

vltansky:
@SilencerWeb I can't reproduce the issue.
Maybe you accidentally swiping? Try to reproduce this issue on this example:
https://stackblitz.com/edit/swiper-demo-7-custom-pagination-svb5wz?file=index.html

SilencerWeb:
@vltansky in your example everything is ok, that's probably an accident swiping but I'm not sure why the slider thinks I'm swiping when I'm not doing anything close to this

aykutkardas:
I was having the same problem. It was enough to give a value of `20` for `threshold`.

| Name | Type | Default | Description |
| -- | -- | -- | --|
| threshold | number | 0 | Threshold value in px. If "touch distance" will be lower than this value then swiper will not move|

**Source:** https://swiperjs.com/swiper-api (threshold)

JoaoPauloLisboa:
@vltansky in your example, if you enable simulateTouch, you can reproduce the error.

domstubbs:
[...]

PR #3933, opened 2020-11-07
disable event allowTouchMove when click pagination

VinhVictor:


vltansky:
closes #3816

vltansky:
Will be fixed in upcoming `v7` version with `touchEventsTarget` default changes to `wrapper`

PR #3954, opened 2020-11-17
Problem: swiper.swipeDirection after swipe and use other event

MaximusFT:
Problem: after use swipe, all other events (that change slide) not clear param swiper.swipeDirection. Consequently difficult understand what the next event changed slide.

fix swiper.swipeDirection param when
* click on navigation
* click on pagination
* use mousewhell
If before used swipe event.


nolimits4web:
Hi `swipeDirection` is only for the internal usage. And this fix is not really correct, as basically setting it to `undefined` must be done in before any `.slideTo` calls everywhere, which I don't want to introduce. For your custom solution, you can just listen for touchstart/move/end events to calc the swipe direction and decide on how it was done

Issue #3928, opened 2020-11-06
Swiper lazyload loads images outside viewport


rhandrade:
I'm using Swiper with lazyload to create rich carousels and improve usability. But I noticed the Swiper loads the first image of slider no matter when or not is visible or in viewport. This creates a strange behavior and is detect by Google Pagespeed / Lighthouse. Images on website footer is loaded together in page load.

This is a: bug.

* Swiper Version: 6.1.2
* Platform: macOS, Windows, iOS, Android on Chrome / Firefox / Edge
* Browser Versions: Chrome, Firefox, Edge, Safari
* JSFiddle: https://jsfiddle.net/gnzqou49/

### What you did
Implements Swiper with lazyload to prevent images outside viewport to impact page load. The code below show options set for Swiper. The code is based on Lazyload example available in Swiper website.

```javascript
var swiper = new Swiper('.swiper-container', {
  preloadImages: false,
  watchSlidesProgress: true,
  watchSlidesVisibility: true,
  lazy: {
    loadOnTransitionStart: true,
  },
  pagination: {
    el: '.swiper-pagination',
    clickable: true,
  },
  navigation: {
    nextEl: '.swiper-button-next',
    prevEl: '.swiper-button-prev',
  },

});
```

### Expected Behavior
Swiper carousel is initiated and just first image (or nexts if enabled) inside viewport / visible is loaded.

### Actual Behavior
Swiper carousel is initiated and  first image (or nexts if enabled) is loaded even not visible or outside viewport.

vltansky:
fixed in #3955

PR #3955, opened 2020-11-18
Fix issue 3928, should not lazy load when swiper not inView.

ygj6:
Fix issue #3928.

### DIFF
--- a/src/components/lazy/lazy.js
+++ b/src/components/lazy/lazy.js
@@ -1,3 +1,4 @@
+import { getWindow } from 'ssr-window';
 import $ from '../../utils/dom';
 import { bindModuleMethods } from '../../utils/utils';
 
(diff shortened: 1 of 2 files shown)


nolimits4web:
Hi! Like the idea, but it is not so easy here. What I think:
- First of all it should be configurable (and disabled by default) not to introduce breaking/different behavior for existing setups
- Checking `window.scroll` is not a valid for all the cases as Swiper can be in some inner element with `overflow: auto`. So "scrolling element" should be also configurable (with defaults to window/document) that should be considered for offsets/scroll calculations

ygj6:
Yes, you're right, I've overlooked these details. I will take some time to look at it again.

ygj6:
How about always checking `window.onscroll` by default?

nolimits4web:
Thanks! I just pushed couple of fixes, so now it will listen for scroll only on one element, `window` or any custom specified. Also `onscroll` should not be used as it can be easily overwritten by user.

Hirbod:
@nolimits4web @ygj6  this changes are missing in the documentation.
And don't you think using an IntersectionObserver might be a better and performant solution? Listening to scroll events is not the right thing to do anymore imho.

Hirbod:
And why is it disabled when `loop` is true

Issue #3946, opened 2020-11-12
Swiper React - swiper.updateSize() not firing


tomhill991:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: ^6.3.5
* Platform/Target and Browser Versions: swiperjs for React
* Live Link or JSFiddle/Codepen or website with issue: 

### What you did
want to be able to update the size of the swiper-wrapper after a max-height animation is triggered on the slide

### Expected Behavior
the mySwiper.updateSize() should update the height of the swiper-wrapper so that the user can scroll down to the bottom of the active slide after the animation has occurred

### Actual Behavior
user cannot scroll down to the bottom of the slide because mySwiper.updateSize() does not fire (it returns undefined). Consequently swiper-wrapper does not increase in height even though the height of the current slide has changed (due to the max-height animation being triggered)

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](http://stackoverflow.com/questions/tagged/swiper)
for that.

Before you open an issue, please check if a similar issue already exists or has been closed before.

nolimits4web:
It is not clear for me what you mean without live example

tomhill991:
managed to resolve this by setting a timeout and using updateAutoHeight() instead. Not sure if this is the best solution but maybe this will help someone in the future. Code looks like this: 

`import React, { useState } from 'react';

function ParentComponent(props) {
  let [swiper, setSwiper] = useState(null)

  return(
    <Swiper 
      onSwiper={(swiper) => {setSwiper(swiper)}}
      autoHeight
    >
      <ChildComponent
          title={"title"}
          text={"text"}
          swiper={swiper}
      />
    </Swiper>
  ) 
}

function ChildComponent(props) {
    let [active, setActive] = useState(false)
    let swiper = props.swiper

    // toggle active class
    const handleClick = () => {
        if (active) {
            setActive(false)
            setTimeout(() => {
                swiper.updateAutoHeight()
            }, 200)
        }
        else {
            setActive(true)
             setTimeout(() => {
                swiper.updateAutoHeight()
            }, 250)
        }
    }

    return (
    <div className={"infos" + (active ? ' active' : '')}>
        <div onClick={handleClick}>
            <h4>{props.title}<span></span></h4>
        </div>

        <div className="text">
            <p>{props.text}</p>
        </div>
    </div>
    )
}

export default ChildComponent`

[...]

PR #3967, opened 2020-11-23
Fix issue 3946.

ygj6:
Fix issue #3946.


nolimits4web:
Well, I would avoid such fixes as it is more like a hack and seems pretty much usage specific. For example what and why is it `200`. I think we should leave such solutions to the end user implementation as per app requirements

ygj6:
@nolimits4web I agree, so I think [this issue](https://github.com/nolimits4web/swiper/issues/3946) can be closed.

Issue #3984, opened 2020-11-28
Different slide with different height


lawchihon:
This is a (multiple allowed):
* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 5.
* Platform/Target and Browser Versions: IOS
* Live Link or JSFiddle/Codepen or website with isssue: https://codepen.io/lawchihon/pen/ZEpYbra

### What you did
EXPLAIN WHAT YOU DID, PREFERABLY WITH CODE EXAMPLES, HERE.
I am trying to set different height on different slide.
Tried to set height as auto or fixed pixel, but either the style get overwritten or behaviour become weird

### Expected Behavior
Allow different height on each slide

### Actual Behavior
Forced to have all slide to have same height

ygj6:
@lawchihon Hi, the following is the author’s opinion, Is this what you want?
> That will break many use cases. I don't really understand the issue. If you need different slide widths/heights this is exactly what slidesPerView: 'auto' mode exist, where it respects slide sizes and doesn't override anything

PR #3995, opened 2020-12-02
Fix issue 3984.

ygj6:
Fix issue #3984.


nolimits4web:
That will break many use cases. I don't really understand the issue. If you need different slide widths/heights this is exactly what `slidesPerView: 'auto'` mode exist, where it respects slide sizes and doesn't override anything

PR #3997, opened 2020-12-02
fix: swiper-slide-duplicate-next class typo

vltansky:
Vue, Svelte and Vue checking `swiper-slide-duplicate next` class to identicate `slideData.isNext` variable.
But I see that the right class is `swiper-slide-duplicate-next` (**with** dash)
image


nolimits4web:
Yes, thanks!)

PR #4015, opened 2020-12-08
 chore: fix eslint/prettier endOfLine warnings for windows users

vltansky:
On windows, eslint showing error on every line due to using `crlf` end of the line and not `lf` that forced by `plugin:prettier/recommended`. Adding `"endOfLine": "auto",` to `.prettierrc` solves this for me.
The only thing I'm not sure about is why my VSCode using `crlf` although `.editorconfig` has `end_of_line = lf`. Maybe there is another solution in this direction.

This PR also update `prettier` to v2.2.1
Untitled


nolimits4web:
Merged, thanks!

PR #4016, opened 2020-12-08
build: use terser 5.5.1

vltansky:
This PR updates Terser from version 4.7.0 to 5.5.1 and changes few lines in the build scripts (adding `await` and etc according to Terser changes). As a result: `swiper-bundle.min.js` went from ~140kb to ~137kb


nolimits4web:
Thanks!

PR #4022, opened 2020-12-09
optimized updateslides code

ygj6:
Modify:
 `const previousSlidesGridLength = swiper.snapGrid.length`
    to
 `const previousSlidesGridLength = swiper.slidesGrid.length`

### DIFF
--- a/src/components/core/update/updateSlides.js
+++ b/src/components/core/update/updateSlides.js
@@ -15,7 +15,7 @@ export default function updateSlides() {
   const slidesGrid = [];
   const slidesSizesGrid = [];
 
-  function slidesForMargin(slideIndex) {
+  function slidesForMargin(slideEl, slideIndex) {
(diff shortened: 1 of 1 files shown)


src/components/core/update/updateSlides.js
```diff
@@ -15,7 +15,7 @@ export default function updateSlides() {
   const slidesGrid = [];
   const slidesSizesGrid = [];
 
-  function slidesForMargin(slideEl, slideIndex) {
+  function slidesForMargin(slideIndex) {
```
nolimits4web, line 18: It can not be deleted, now `slideIndex` will represent the slide element and it will work wrong

ygj6: Sorry, I will restore it.

nolimits4web:
Merged, thanks!

PR #4028, opened 2020-12-09
fix(angular): imports

vltansky:
This PR contains:
1. fix for deep import warning:
```
Warning: Entry point 'angular' contains deep imports into 'D:/projects/tailwind-test/node_modules/swiper/types'. This is probably not 
a problem, but may cause the compilation of entry points to be out of order.
```
https://github.com/nolimits4web/swiper/pull/3918#issuecomment-740677749

by adding `types/package.json` (creating entry point). Is it fine as for you?

2. `angular/package.json` name changed to `swiper_angular`. I couldn't rename to `swiper/angular`, build is failing with the following error:

```
Building Angular Package

------------------------------------------------------------------------------
Building entry point 'swiper/angular'
------------------------------------------------------------------------------
✖ Compiling TypeScript sources through NGC
ERROR: Maximum call stack size exceeded
An unhandled exception occurred: Maximum call stack size exceeded
See "C:\Users\vltan\AppData\Local\Temp\ng-FqgGio\angular-errors.log" for further details.
```

I honestly don't know why this happens and what ng-packagr trying to do. 

3. Changed Input types to use `SwiperOptions` interface (easier maintenance) e.g:
```js
@Input() url: string;// from
@Input() url: SwiperOptions['url'];//to
```
need to create a node script that will generate Input and outputs, so it will be easier in the future.


nolimits4web:
Merged, thanks!

PR #4029, opened 2020-12-09
chore: update dependencies

vltansky:
I ran `npm run build:prod`, seems like everything works just fine.

If you want to see which dependencies updated take a look at the first commit, as in the second commit npm rearranged JSON. The second commit only updates vue-loader.

https://github.com/nolimits4web/swiper/pull/4029/commits/ce800e490bc8dea5a27bb7e6855eec0b34c76b87


nolimits4web:
Thanks!

PR #4060, opened 2020-12-17
build(styles): refactor & add css components

### LINKED ISSUE #4053: CSS of the swiper is bundled by components for PostCSS users
This is a (multiple allowed):
* [ ] bug
* [x] enhancement

* Swiper Version: 6.4.1
* Platform/Target and Browser Versions: All

### What you did
I use postCSS in the project but still want to benefit from smaller bundle size of the swiper by importing css components that only relate to components I am using such as Pagination, Navigation, Zoom, etc. - similar to how SASS people have it.

### Expected Behavior
CSS of the swiper is bundled by components, not as one monolithic css file

vltansky:
closes #4053 
Adds CSS generation to components folders.
I'm not sure if we should write both `**.css` and `**.min.css`, also if the name of CSS and less/scss files will be the same, you can't control what you are importing. That's why I decided that we can just output only `**.min.css`.
@nolimits4web @AndrewBogdanovTSS what do you say?
image

Also, this PR refactor styles build script.


AndrewBogdanovTSS:
I guess you mean that this problem may happen if you omit an extension when importing scss/less files? But I'm okay with having it as **.min.css since those shouldn't be editable anyway

nolimits4web:
> That's why I decided that we can just output only `**.min.css.`

Yes, good idea

nolimits4web:
For me everything looks and works good, merging, thanks!

nolimits4web:
Ok, it was actually missing the main core `swiper.min.css`. I've added it

PR #4082, opened 2020-12-30
fix(swiper5): postinstall echo on windows

### LINKED ISSUE #4081: Line breaks can't be quoted on Windows on Swiper 5
https://github.com/nolimits4web/swiper/issues/3985 should be meged in 5.x.x branch too.

Also maybe we should use `"postinstall": "node -e \"try{require('./postinstall')}catch(e){}\""` like [core-js](https://github.com/zloirock/core-js/blob/4645db4ce9efa01ddaa3d5e951f3825a934bf57e/packages/core-js/package.json) do?

vltansky:
closes #4081

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,3 +5,4 @@ node_modules
 dist
 package
 src/swiper.js
+postinstall.js
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ custom
 
 .versions
 .idea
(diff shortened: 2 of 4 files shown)


nolimits4web:
I guess we already have this fix merged with another PR

vltansky:
it was targeting `v5` branch

nolimits4web:
Ok, merged then :)

PR #4117, opened 2021-01-21
fix(angular): slide change render

vltansky:
https://stackoverflow.com/questions/65570802/angular-swiper-not-working-when-using-ngif


nolimits4web:
Merged, thanks!

Issue #3772, opened 2020-08-21
Touch drag no longer works


jwoodwardtfx:
There is a bug introduced by the change "Simplify touch events detection" (11dcfe5).

It looks like there has been an attempt to simplify detection of "touch" support, however the way it's been changed will break most desktop/tablet use cases.

The newer code uses either:

'ontouchstart' in window => this has been deprecated and taken out of Chrome for ages (https://www.chromestatus.com/feature/4764225348042752)
DocumentTouch => this has been removed from the standards and shouldn't be used either (https://developer.mozilla.org/en-US/docs/Web/API/DocumentTouch)
The old method checks for "window.navigator.maxTouchPoints", which (at least on Windows desktop/tablets) will function correctly.

This commit needs reverting.

See also issues 3589 and 3649.

ksn-systems:
I Have had the same issue with 6.1.2 on Chrome.

However 5.4.0 is also broken on Chromium / linux 

My fallback release is 5.3.8 where swiping works everywhere but the current page blinks on both Chrome windows and Chromium / Linux as the start of a drag / swipe

matadorhernan:
Yeah I need to fallback as well, the swiper api is now broken for anything that isnt either an IOs or an Android device, I agree with the author of this issue

jwoodwardtfx:
Any further progress on this?

Thanks.

ksn-systems:
Just tried 6.0.2 on Chromium / linux 

Touch drag is still Broken.

antiero:
Is the current workaround for this to revert to using 5.3.8?

jwoodwardtfx:
Yes that is one workaround, but I use Framework7 which includes Swiper. I've had to manually hack the code to revert this change.

It's looking like Swiper is becoming a dead project for anything other than a Android/iOS Smartphone as there seems to be no desire to fix this issue, which basically stops use of Swiper on any Windows-based tablet.

wizardnet972:
So many problems it caused my clients.
28+ days and yet no fix :(
Demos are not working (stackblitz). 
Sad.

wizardnet972:
This problem should be critical.
Many problems it caused to me and my clients.
28+ days and yet no fix :(
Demos are not working (stackblitz).
Sad.

mberezinski:
Any progress? Demo https://swiperjs.com/demos/ is not working on any touch enabled windows system like Microsoft Surface devices.

vltansky:
[...]

Issue #3649, opened 2020-07-03
Swipe gestures dont work on tablet or desktop


antonioberetini:
I tested swiper components on windows machine with touch interface, swipe gestures don't work. I tried this out in Chrome and Chromium Edge. Is this a known issue? I am working on a progressive web app which is targeting touch enabled tablets with windows

kebien6020:
In #3589  there is a couple of cases with similar issues.

vltansky:
Duplicate of #3772

Issue #3589, opened 2020-05-21
Touch drag no longer works (possible bug)


ksn-systems:
Hi

Using version 5.3.8 with a simple scroller of a few slides and the cube effect, I can successfully manually change slides as expected by dragging on a touch screen (A HP PC). If I update to 5.4.0 or 5.4.1 whilst I can still trigger a transition  via the API, when I drag the swioer, it moves a very small amount and "sticks".

(I set direction, effect and speed from vars)

     swipers[data.block_id] = new Swiper ('#' + data.block_id + ' > .swiper-container', {
        // Optional parameters
        direction: direction,
        'effect' : effect,
        'speed' : speed,
        //'roundLengths' : true,
        pagination: {
            el: '.swiper-pagination_' + data.block_id,
        },
        coverflowEffect: {
            rotate: 60,
            slideShadows: false
        },
        flipEffect: {
            rotate: 30,
            slideShadows: false
        },
        cubeEffect: {
            slideShadows: false,
            shadow: showShadow
        },
        loop: true,
    });

bariscc:
Can confirm, touch drag is not working on demos.

Windows 10
Chrome Version 83.0.4103.97
Edge Version 83.0.478.45
No issues on Firefox

kebien6020:
Also can confirm, not working on demos.

Windows 10
Chrome Version 83.0.4103.116
In Firefox 78.0.1 having trouble with vertical swiping. Horizontal works fine.

CSS scrolling demo works fine for both.

antonioberetini:
Same here

jorovipe97:
Same here, it works on chrome android but not on windows.

jwoodwardtfx:
I think it's to do with the change "Simplify touch events detection" (https://github.com/nolimits4web/swiper/commit/11dcfe55bd5bcee62fc1c89da37ca4a607c80b0f)

It looks like there has been an attempt to simplify detection of "touch" support, however the way it's been changed will break most desktop/tablet use cases.

The newer code uses either:
- 'ontouchstart' in window => this has been deprecated and taken out of Chrome for ages (https://www.chromestatus.com/feature/4764225348042752)
- DocumentTouch => this has been removed from the standards and shouldn't be used either (https://developer.mozilla.org/en-US/docs/Web/API/DocumentTouch)

The old method checks for "window.navigator.maxTouchPoints", which (at least on Windows desktop/tablets) will function correctly.

This commit needs reverting.

[...]

PR #4122, opened 2021-01-22
fix: swipe on pointer events devices

### LINKED ISSUE #3728: Swiper not transitioning using Touchscreen on Windows 10 in Chrome.
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version:  6.0.4
* Platform/Target and Browser Versions: 
Windows 10. Touch Screen (Dell XPS 15). Chrome Version 84.0.4147.89

### What you did
Attempt to change slides using a touch screen in Windows 10 in Chrome.

### Expected Behavior
Slider should translate to the next slide.

### Actual Behavior
Swiper does not change between slides.
When the screen is first touched to swipe a direction it immediate appears to start to transition but then gets stuck in place mid transition.
Moving the mouse over the slider then transitions the slides left / right as if the mouse button is being held down.

I don't believe this is an issue specific to my project as this issue can also be replicated using the top swiper on https://swiperjs.com/

**Note:**
Works in Firefox 78.0.2 
This wasn't an issue in my old 5.2.1 version.

vltansky:
closes #3728

#3589
#3649
#3772

thanks @paultuckey your workaround gave the right direction :)

### DIFF
--- a/src/components/core/classes/addClasses.js
+++ b/src/components/core/classes/addClasses.js
@@ -1,11 +1,14 @@
 export default function addClasses() {
   const swiper = this;
-  const { classNames, params, rtl, $el, device } = swiper;
+  const { classNames, params, rtl, $el, device, support } = swiper;
   const suffixes = [];
 
   suffixes.push('initialized');
   suffixes.push(params.direction);
 
+  if (support.pointerEvents) {
+    suffixes.push('pointer-events');
+  }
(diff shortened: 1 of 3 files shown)


nolimits4web:
Thanks! I think it should be pan-x or pan-y based on Swiper direction - vertical or horizontal.

PR #4125, opened 2021-01-22
chore(deps): update deps

vltansky:
sorry for the duplication. It was few node modules I wasn't sure about before


nolimits4web:
By the way, do we need `tslib` as a dependency of Swiper, and not in `devDependencies`?

vltansky:
> tslib is not a dependency of angular itself. But rather a dependency of the TypeScript compiler when you enable TypeScript importHelpers.
> Any library that is compiled using importHelpers should always have a direct dependency on tslib and not rely on other libraries to install it.

It looks like no!

PR #4129, opened 2021-01-24
refactor(types): use const in examples

vltansky:
https://github.com/nolimits4web/swiper-website/pull/97

Maybe you prefer to change variable name from `mySwiper` to something else? e.g `swiperRef`?


nolimits4web:
> Maybe you prefer to change variable name from mySwiper to something else? e.g swiperRef?

Let's rename `mySwiper` to just `swiper` to avoid confusion with `swiperRef` for example in React/Vue where it is a bit different thing

vltansky:
done

PR #4131, opened 2021-01-25
fix(angular): Cannot set property 'classNames' of undefined

### LINKED ISSUE #4130: Angular: Cannot set property 'classNames' of undefined
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Swiper Version: 6.4.8
* Platform/Target and Browser Versions: Windows, CHROME
* Live Link or JSFiddle/Codepen or website with isssue: https://stackblitz.com/edit/angular-swiper-classname?file=src/app/app.component.ts

### What you did
Create new slide after swiper init, by button

### Expected Behavior
Create slide like usual

### Actual Behavior
Slide create, but in console write: Cannot set property 'classNames' of undefined
And the usual slider operation breaks down

vltansky:
closes #4130

### DIFF
--- a/src/types/swiper-events.d.ts
+++ b/src/types/swiper-events.d.ts
@@ -223,6 +223,14 @@ export interface SwiperEvents {
    * INTERNAL: Event will fired after setting CSS classes on swiper slide element
    */
   _slideClass?: (swiper: Swiper, el: HTMLElement, classNames: string) => void;
+
+  /**
+   * INTERNAL: Event will fired after setting CSS classes on all swiper slides
+   */
+  _slideClasses?: (
+    swiper: Swiper,
+    slides: { el: HTMLElement; classNames: string; index: number }[],
+  ) => void;
   /**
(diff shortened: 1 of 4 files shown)


src/types/swiper-events.d.ts
```diff
@@ -223,6 +223,14 @@ export interface SwiperEvents {
...
    */
   _slideClass?: (swiper: Swiper, el: HTMLElement, classNames: string) => void;
+
+  /**
+   * INTERNAL: Event will fired after setting CSS classes on all swiper slides
+   */
+  _slideClasses?: (
+    swiper: Swiper,
+    slides: { el: HTMLElement; classNames: string; index: number }[],
+  ) => void;
```
vltansky, line 233: I use it in angular to rerender class names. So I think it will be better to rerender it once, instead of on every slide.

nolimits4web:
Thanks!

Issue #4134, opened 2021-01-25
New Issue Template Violates Stack Exchange Terms and Conditions


tomjn:
```
P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](http://stackoverflow.com/questions/tagged/swiper)
for that.
```

This is a violation of the stack exchange terms and conditions. You can't do this

nolimits4web:
Your issue is closed because it is not filled correctly. Before you open an issue please review the [contributing guideline](https://github.com/nolimits4web/Swiper/blob/master/CONTRIBUTING.md).

Please fill the issue template correctly and:
* Clearly describe the issue including steps to reproduce when it is a bug.
* Make sure you fill in the earliest version that you know has the issue.
* Provide live link or JSFiddle/Codepen or website with isssue

It will save a lot of time for other contributors to check your issue and fix it.

PR #4135, opened 2021-01-25
Fix violation of stack exchange terms and conditions

tomjn:
Stops issues directing users to stack exchange for support, stack overflow is not a dumping ground for user support


nolimits4web:
See https://github.com/nolimits4web/swiper/issues/4134#issuecomment-766977266

PR #4141, opened 2021-01-26
Playground refactor

vltansky:
move all `playground-*` folders under `./playground`
from:
```
playground
playground-react
playground-vue
playground-svelte
playground-angular
```
to
```
playground
- core
- react
- vue
- svelte
- angular
```


nolimits4web:
Thanks, was also thinking about it :)

PR #4145, opened 2021-01-26
docs: add badges (all contributors)

vltansky:
https://github.com/nolimits4web/swiper/issues/3149


vltansky:
There is some problem generating readme..

PR #4153, opened 2021-01-28
ci: split test to three ci workflows

vltansky:
split test workflow into three workflows:
- Formatting (`npm run check-format`)
- Lint (`npm run lint`)
- Build (`npm run build:prod`)

I think this way indication will be more informative and in the future, it will be easier to add new tests, like e2e and etc.
What do you think?


nolimits4web:
Makes sense, I agree it will better

PR #4170, opened 2021-01-31
feat: cypress

vltansky:


cypress/integration/modules/core.js
```diff
@@ -0,0 +1,97 @@
...
+
+context('Core', () => {
+  describe('centeredSlides & slidesPerView', () => {
+    beforeEach(() => {
+      cy.swiperPage();
+      cy.initSwiper({
+        centeredSlides: true,
+        slidesPerView: 3,
+        spaceBetween: 10,
+        navigation: {
```
nolimits4web, line 11: Not sure why it happens, but specifying navigation here will throw an error. Because somehow it looses default parameters (e.g. rest navigation params like `swiper.params.navigation.disabledClass` will be undefined)

cypress/support/commands.js
```diff
@@ -0,0 +1,111 @@
...
+// https://on.cypress.io/custom-commands
+// ***********************************************
+
+Cypress.Commands.add('getNextSlide', { prevSubject: 'optional' }, () => {
+  return cy.get('.swiper-button-next');
+});
+Cypress.Commands.add('getPrevSlide', { prevSubject: 'optional' }, () => {
+  return cy.get('.swiper-button-prev');
+});
+Cypress.Commands.add('nextSlide', { prevSubject: 'optional' }, () => {
```
nolimits4web, line 17: We shouldn't rely here on some modules, better here will be to use Swiper API and call it as `swiper.slideNext()`. Otherwise we can't slide next if it doesn't have navigation enabled

cypress/support/commands.js
```diff
@@ -0,0 +1,111 @@
...
+Cypress.Commands.add('getSliderWrapper', { prevSubject: 'optional' }, () => {
+  return cy.get('.swiper-wrapper');
+});
+
+Cypress.Commands.add('getSliderContainer', { prevSubject: 'optional' }, () => {
+  return cy.get('.swiper-container');
+});
+
+Cypress.Commands.add('getSlide', { prevSubject: 'optional' }, (subject, slideIndex) => {
+  return cy.get(`.swiper-slide:nth-child(${slideIndex})`);
```
nolimits4web, line 37: If we pass slide index which yes, makes sense, then I guess it should be `:nth-child(${slideIndex + 1})` (indexes in CSS starts from `1`) :)

vltansky: Fixed

cypress/support/commands.js
vltansky: The idea was to rely on swiping, not navigation.
I agree that it should be renamed to a more representative name `navigationNextSlide` maybe?

I think relying on swiping is better than on method `swiper.slideNext()`

Issue #4181, opened 2021-02-02
It is not disabled when props of navigation, pagination and scrollbar are false


mya-ake:
- Swiper Version: https://github.com/nolimits4web/swiper/commit/71e3d15bd5f8375304a0125b23f2b7f0187c910b
- Platform/Target and Browser Versions: any

## What You Did

https://codesandbox.io/s/ancient-haze-3g201?file=/src/App.js

<img width="876" alt="スクリーンショット 2021-02-02 23 38 04" src="https://user-images.githubusercontent.com/16383815/106615569-ea542f80-65af-11eb-8c8f-31bf5b9cfa61.png">

 

## Expected Behavior

navigation, pagination and scrollbar are not displayed.

## Actual Behavior

navigation, pagination and scrollbar are displayed.

mya-ake:
I will continue to issue the fix PR.

PR #4182, opened 2021-02-02
fix: navigation, pagination, scrollbar props #4181

mya-ake:
fix: #4181 

In the case of `{}`, `needsNavigation`, `needsPagination`, and `needsScrollbar` will be true, so if each props is false, the initial value is given so that element is null.

### DIFF
--- a/src/react/get-params.js
+++ b/src/react/get-params.js
@@ -33,7 +33,17 @@ function getParams(obj = {}) {
     }
   });
   ['navigation', 'pagination', 'scrollbar'].forEach((key) => {
-    if (params[key] === true || params[key] === false) params[key] = {};
+    if (params[key] === true) params[key] = {};
+    if (params[key] === false) {
+      switch (key) {
+        case 'navigation':
+          params[key] = { nextEl: null, prevEl: null };
+          break;
+        default:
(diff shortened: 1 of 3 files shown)


src/react/get-params.js
```diff
@@ -33,7 +33,17 @@ function getParams(obj = {}) {
     }
   });
   ['navigation', 'pagination', 'scrollbar'].forEach((key) => {
-    if (params[key] === true || params[key] === false) params[key] = {};
```
vltansky, line 36: Why do you need to specify default value?
Wouldnt it be enough to change it to:
```js
if (params[key] === true) params[key] = {};
```

vltansky:
You can contribute to vue & svelte versions with the same change.

Take a look at `src/svelte/get-params.js` and `src/vue/get-params.js`

src/react/get-params.js
mya-ake: I wondered if I needed to convert it to an object, so I set the default value.
That change seems to be enough, so I will change it.

mya-ake:
I also updated svelte and vue.

nolimits4web:
Merged, thanks! Checked and everything seems fine

PR #4184, opened 2021-02-02
feat: slim version

vltansky:
There are still a lot of people using Swiper via CDN without tree shaking.
At the same time, most of the comparison of swiper versus other carousels (e.g slick) ends up with comparing Kilobytes. A lot of people see swiper as a huge library. And this is true when `swiper-bundle` used. But it's containing a lot of features that regular usage of carousel probably won't need.
So I thought that some kind of slim version for CDN usage would be appropriate.
I included modules that I think are most used. But I'm sure that this list can be improved.

includes following modules:
- navigation
- pagination
- lazy
- a11y
- autoplay

End up with `24.97 KB` vs `35.63 KB` of `swiper-bundle`


nolimits4web:
Like the idea, I was thinking for a while about it:

- I don't think we need a scrollbar
- also would be good to detach some core functionality that has relatively large amount of codes, like `freeMode`, `slidesPerRow`, maybe something else.

vltansky:
I've added a scrollbar
Guess you should prepare a list of settings that should be extracted to modules and show an example of how you are separating it. Then I could help, right now I'm not sure how to do it properly.

Also, is removing features from the Core version mean the need for a major release?

nolimits4web:
I created a new RFC thread https://github.com/nolimits4web/swiper/issues/4227

nolimits4web:
And moved it to discussions https://github.com/nolimits4web/swiper/discussions/4228?converting=1 :)

PR #4221, opened 2021-02-13
docs: add tree-shakable and types badges

vltansky:
adds following badges to readme:
- tree-shakeable
- types included


nolimits4web:
Nice, thanks!

PR #4222, opened 2021-02-13
Add features section to readme

vltansky:
Maybe we should add all modules here?

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,25 @@ Swiper - is the free and most modern mobile touch slider with hardware accelerat
 
 Swiper is not compatible with all platforms, it is a modern touch slider which is focused only on modern apps/platforms to bring the best experience and simplicity.
 
+## Features
+- **Tree-shakeable**: Only modules you use will be imported into your app's bundle.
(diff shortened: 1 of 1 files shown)


vltansky:
@nolimits4web guess the list can be improved & more modules can be added?

nolimits4web:
@vltansky you mean modules like Navigation, Pagination, etc.? What is missing, I guess worth mention is Virtual and maybe Zoom modules. All other basically covered in existing list in some way

vltansky:
Guess we should extract this bullet:
"And many more ...: All Swiper well-known features are also here: Responsive, Scroll prevention, Resistant bounds, Autoplay, Loop mode, Nested Swipers" to multiple. As loop mode, nested swipers, Autoplay, Scrollbar (+progress bar) also worth a bullet for each of them?

Also, I think we should add React, Vue, Svelte and Angular versions as features.

nolimits4web:
@vltansky I agree :)

vltansky:
I'll make a PR with additions.

PR #4238, opened 2021-02-16
refactor(core): addClasses minify

vltansky:


src/components/core/classes/addClasses.js
```diff
@@ -1,45 +1,38 @@
...
-  }
-  if (params.slidesPerColumn > 1) {
-    suffixes.push('multirow');
-    if (params.slidesPerColumnFill === 'column') {
-      suffixes.push('multirow-column');
+function pushClasses(settings, containerModifierClass) {
+  const result = [];
+  settings.forEach((item) => {
+    if (typeof item === 'object') {
+      for (const [condition, classNames] of Object.entries(item)) {
```
nolimits4web, line 5: I would avoid using `for of` loops, babel can transpire it to be used as `iterators`. 
```js
Object.entries(item).forEach(([condition, classNames]) => {...}) 
```
will do basically the same

PR #4241, opened 2021-02-16
refactor(core): update slides

vltansky:


src/components/core/update/updateSlides.js
```diff
@@ -2,6 +2,27 @@ import { getWindow } from 'ssr-window';
 import { extend } from '../../../utils/utils';
 
 export default function updateSlides() {
+  const getDirectionLabel = (property) => {
+    const swiper = this;
```
nolimits4web, line 6: Let's move `const swiper = this` above the `const getDirectionLabel`, because right below we have same `const swiper = this` in upper scope

src/components/core/update/updateSlides.js
```diff
@@ -2,6 +2,27 @@ import { getWindow } from 'ssr-window';
...
+      'padding-left': 'padding-top',
+      'padding-right': 'padding-bottom',
+      'marginRight': 'marginBottom',
+    }[property];
+  };
+  const getDirectionPropertyValue = (node, label) => {
+    return parseFloat(node.getPropertyValue(getDirectionLabel(label)) || 0);
+  };
+
   const swiper = this;
```
nolimits4web, line 26: 👆

nolimits4web (changes requested):

src/components/core/update/updateSlides.js
vltansky: ?

nolimits4web:
All good, thanks!

Issue #4272, opened 2021-02-27
Swiper props are not working on ios safari


faskade:
- Swiper Version: 6.4.15
- Platform/Target and Browser Versions: ios 12.4.8/Safari

## What You Did
I just tried to provide some props to swiper. Even in vue playground from this repo, this bug reproduces.  
For instance, you can just try "npm run vue" and check it on ios. none of the props are working.

## Expected Behavior

Swiper gets the props

## Actual Behavior
Swiper uses default props in any case. 

## How it can be fixed
The reason is that `Object.keys(props)` returns an empty array in function `getParams`. 
I suggest you convert props into a plain object before using `Object.keys()`. 
It could be reached for example by: `plainObj = JSON.parse(JSON.stringify(obj));`
Or: `plainObj = Object.assign({}, obj)`

vltansky:
@faskade would you want to open PR?

faskade:
@vltansky done

nolimits4web:
Fixed in #4273

PR #4273, opened 2021-02-27
Fix Issue #4272

faskade:
Fix Issue [#4272](https://github.com/nolimits4web/swiper/issues/4272)

### DIFF
--- a/src/vue/get-params.js
+++ b/src/vue/get-params.js
@@ -15,7 +15,9 @@ function getParams(obj = {}) {
 
   const rest = {};
   const allowedParams = paramsList.map((key) => key.replace(/_/, ''));
-  Object.keys(obj).forEach((key) => {
+  // Prevent empty Object.keys(obj) array on ios.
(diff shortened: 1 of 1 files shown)


nolimits4web:
Can you elaborate a bit, as I don't understand, how and why this object becomes empty on iOS? :)

faskade:
The object doesn't become empty. Everything okay with the object. But since it's some special reactive vue object it has some problems with methods like Object.keys(). Object.keys(props) returns an empty array instead of the key's array on IOS. At least on my iphone.

nolimits4web:
Ok, merged, thanks!

Saltibarsciai:
Hey, @nolimits4web  when are you planning to release this? 
I have such prop `:navigation="{ nextEl: '.nextArrow--c', prevEl: '.prevArrow--c' }"` but on iphone 6 arrows do nothing, I think this pull request fixes it (on other phones it works)

Issue #4274, opened 2021-02-27
Lazy loading in free mode may not work


bcmk:
- Swiper Version: 5d15292
- Platform/Target and Browser Versions: Ubuntu Linux 20.04 / Chrome 88.0.4324.182

## What You Did

If both lazy loading and free mode are enabled, lazy loading may not work in some circumstances.
Basically if you move to not loaded yet slides with a mouse and completely stop moving before releasing, slides are not loaded.

I recorded a video showing it: https://youtu.be/zvfhp20sccA.
At first I'm opening a tab with an example built using current master branch of Swiper.
Then I'm opening a tab with the version patched with this proposed change. 

Here are examples:
The version from master branch: https://bcmk.github.io/swiper/master.html
Fixed version: https://bcmk.github.io/swiper/fixed.html

Complete examples: https://github.com/bcmk/swiper/tree/lazy-free-fix-example
Look for master.html built from current master (5d15292) and fixed.html built from proposed patch (I'm sending the pull request as well)

## Expected Behavior

Talking about provided video, slides number 9 and 10 should be loaded

## Actual Behavior

Slides number 9 and 10 were not loaded

PR #4275, opened 2021-02-27
Fix for lazy loading in free mode

bcmk:
If both lazy loading and free mode are enabled, lazy loading may not work in some circumstances.
Basically if you move to not loaded yet slides with a mouse and completely stop moving before releasing, slides are not loaded.

I recorded a video showing it: https://youtu.be/zvfhp20sccA.
At first I'm opening a tab with an example built using current master branch of Swiper.
Then I'm opening a tab with the version patched with this proposed change. 

Here are examples:
The version from master branch: https://bcmk.github.io/swiper/master.html
Fixed version: https://bcmk.github.io/swiper/fixed.html

I haven't created test cases yet and I would be grateful if someone points me to the correct way to add them.

Corresponding issue: https://github.com/nolimits4web/swiper/issues/4274

P.S. It fixes the behavior only for `loadOnTransitionStart: true`.
The issue for `loadOnTransitionStart: false` still exists.

### DIFF
--- a/src/components/core/events/onTouchEnd.js
+++ b/src/components/core/events/onTouchEnd.js
@@ -218,7 +218,7 @@ export default function onTouchEnd(event) {
             });
           }, 0);
         });
-      } else if (swiper.velocity) {
+      } else {
         swiper.updateProgress(newPosition);
         swiper.setTransition(momentumDuration);
         swiper.setTranslate(newPosition);
@@ -230,8 +230,6 @@ export default function onTouchEnd(event) {
             swiper.transitionEnd();
           });
         }
-      } else {
-        swiper.updateProgress(newPosition);
       }
 
       swiper.updateActiveIndex();


vltansky:
We just started to write cypress tests, thats why autoplay still not exists. 
Just create `swiper\cypress\integration\modules\autoplay.js` file and write the tests there.
See `swiper\cypress\integration\modules\core.js` for examples and `swiper\cypress\support\commands.js` for commands.

If you want to help & contribute more you're welcome to include more tests, it would help a lot :)

bcmk:
@vltansky Thanks for pointing me to the tests! However, I didn't get what autoplay has to do with free mode? I guess I should write my test cases in `swiper\cypress\integration\modules\core.js`.

vltansky:
Hey, right, I meant `swiper\cypress\integration\modules\lazy-load.js`, as it's a module https://swiperjs.com/swiper-api#lazy-loading it should have a separate file.

But if it will help you - you can write the test in `swiper\cypress\integration\modules\core.js`, I'll organize it later :)

nolimits4web:
Hi! Thanks for pointing out the issue and for the fix. But this fix introduces some unnecessary calculation and actions running transitions (when there is not). So in https://github.com/nolimits4web/swiper/commit/82bcc5cc2ed3dc5c7995df45cfee1d76ed170139 I just introduced new internal event for this, and do the lazy load

Issue #4297, opened 2021-03-04
No 'exports.__esModule = true;' inside of generated cjs files


Filyus:
Swiper 6.4.15

Files, generated by Svelte don't contain the line `exports.__esModule = true;`
- `node_modules\swiper\cjs\svelte\swiper.js`
- `node_modules\swiper\cjs\svelte\swiper-slide.js`
```js
class Swiper extends SvelteComponent {
  //...
}
//exports.__esModule = true; //missing
exports.default = Swiper;
```

Therefore `_interopRequireDefault()` in `node_modules\swiper\swiper-svelte.cjs.js` returns `.default.default` property instead of simple`.default`:
```js
function _interopRequireDefault(obj) {
  return obj && obj.__esModule ? obj : {
    default: obj
  };
}

var _swiper = _interopRequireDefault(require("./cjs/svelte/swiper"));
exports.Swiper = _swiper.default;

var _swiperSlide = _interopRequireDefault(require("./cjs/svelte/swiper-slide"));
exports.SwiperSlide = _swiperSlide.default;
```

Possible solutions:
1. (**Dirty**) Add `swiper += '\nexports.__esModule = true;'`;
and `swiperSlide += '\nexports.__esModule = true;'`;
before `await fs.writeFile ...` in `swiper/scripts/build-svelte.js`
2. Or change `_interopRequireDefault()` in `node_modules\swiper\swiper-svelte.cjs.js` to check `.default` instead of `.__esModule`:
```js
function _interopRequireDefault(obj) {
  return obj && obj.default ? obj : {
    default: obj
  };
}
```

Filyus:
I still have this issue...

PR #4351, opened 2021-03-18
Fix running with Snowpack

Filyus:
These two small changes are the result of one hard day of work as a programmer with 19 years of experience trying to make Swiper work with Svelte and Snowpack.

1. `peerDependencies` is needed to avoid a second Svelte instance that almost completely stops Swiper from working.
There redundant`^` in the PR, you can remove it like this: `"svelte": "3.x"`, but everything works with it too.
2. `export const __esModule = true;` in `swiper.esm.js` is needed because of double `.default` attribute bug.
I'm tried to describe it here: https://github.com/nolimits4web/swiper/issues/4297, there same problem, but that commit didn't solve it completely.

### DIFF
--- a/package/package.json
+++ b/package/package.json
@@ -67,5 +67,8 @@
   "dependencies": {
     "dom7": "^3.0.0",
     "ssr-window": "^3.0.0"
+  },
+  "peerDependencies": {
+    "svelte": "^3.x"
   }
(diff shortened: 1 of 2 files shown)


nolimits4web:
Merged, thanks! 👍

PR #4366, opened 2021-03-24
feat: a11y.slideLabelMessage

### LINKED ISSUE #4353: Changing swiper-slide aria-label value
Hi, it's possible to change the swiper-slide aria-label value? I need something like "1 of 5" instead of "1/5".  I didn't find anything in the documentation.
Thank you! Swiper is amazing!

vltansky:
closes #4353


nolimits4web:
Thanks, merged!

PR #4367, opened 2021-03-24
Fix getPropertyValue call in updateSlides

yamnikov-oleg:
This is a fix for a bug I have encountered. 

In my case, the swiper with `slidesPerView: 'auto'` has an extra empty space after the last slide, which is visible if you scroll to the end of the gallery.

I don't know if this fixes any other issues that have been reported before.


nolimits4web:
Merged, thanks!

PR #4368, opened 2021-03-24
Avoid using cached dimensions for keyboard inView calculations

domstubbs:
This PR follows on from [my comments](https://github.com/nolimits4web/swiper/pull/3922#issuecomment-806026531) on a previous inView PR.

If a Swiper instance is initialised and subsequently hidden, the inView flag will be set based on the dimensions that the element previously had, because `swiper.height` and `swiper.width` are cached values. Referencing the element’s height properties directly addresses this.

(I wondered whether `swiper.updateSize()` should be used here, but it’s no help as it will only update width and height values that are greater than 0, which hidden elements won’t be.)

### DIFF
--- a/src/components/keyboard/keyboard.js
+++ b/src/components/keyboard/keyboard.js
@@ -56,15 +56,19 @@ const Keyboard = {
       ) {
         return undefined;
       }
+
+      const $el = swiper.$el;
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thanks!

PR #4410, opened 2021-04-07
Remove svelte as peerDependency

### LINKED ISSUE #4409: Svelte listed as peerDependency
- Swiper Version: 6.5.3
- Platform/Target and Browser Versions: N/A

## What You Did

EXPLAIN WHAT YOU DID, PREFERABLY WITH CODE EXAMPLES, HERE.

with yarn v1, yarn add or yarn install if already installed

## Expected Behavior

no warning at install

## Actual Behavior

`warning: .... swiper@6.5.3" has unmet peer dependency "svelte@^3.x".`

JSteunou:
fixes #4409

### DIFF
--- a/package/package.json
+++ b/package/package.json
@@ -67,8 +67,5 @@
   "dependencies": {
     "dom7": "^3.0.0",
     "ssr-window": "^3.0.0"
-  },
-  "peerDependencies": {
-    "svelte": "^3.x"
   }
 }


nolimits4web:
That was added not without a reason I guess https://github.com/nolimits4web/swiper/pull/4351 Maybe @Filyus knows what happens?

JSteunou:
Humm strange... why not just a peerDep in swiper svelte? And how could be 2 instances of svelte if swiper does not add another?

Filyus:
It was the simplest solution to get Snowpack to work, I guess peerDependency can be moved inside swiper/svelte. In theory, you can also completely remove it, but then you have to fix Snowpack, which is probably more correct but more difficult.

nolimits4web:
Let's merge it for now. I will look into Snowpack/Svelte issue today

Issue #3863, opened 2020-10-11
Swiper React->Thumbs->onSwiper


FEMotong:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: EXACT RELEASE VERSION OR COMMIT HASH, HERE.
* Platform/Target and Browser Versions: PLATFORM CLIENT YOU ARE TARGETING SUCH AS macOS, Windows, CORDOVA, IOS, ANDROID, CHROME, ETC.
* Live Link or JSFiddle/Codepen or website with isssue: PREFERABLY _(IF YOU WANT YOUR ISSUE TO BE RESOLVED ASAP)_.

### What you did
![](https://ftp.bmp.ovh/imgs/2020/10/d3d53f0c66e68257.png)
### Expected Behavior
Hope to trigger onSwiper event normally.
### Actual Behavior

onSwiper event is not triggered.

stkaspar:
same here.
it won't fire to change the first slider

nolimits4web:
Fixed by #4440

Issue #3852, opened 2020-10-02
Thumbs in React is not working on version ^6.3.0


DCitalo:
This is a (multiple allowed):
* [x] bug

* Swiper Version: 6.3.2
* Platform/Target and Browser Versions: Possibly all plataforms

### What you did
I followed exactly what Timo did in his [video tutorial](https://www.youtube.com/watch?v=l4kFO6VQPfA&feature=emb_logo). But on my code it was not working, so i tryed to [copy his code](https://github.com/valnub/swiper-react-demo/blob/master/src/App.js), and it still wasn't working... In last resource I downgrade to version 6.0.2, the same version that is used on the video, just after that the thumbs work perfectly. I don't know exactly version that broken between 6.0.2 - 6.3.2 but in the last version the Thumbs aren't working.

### Expected Behavior
Thumbs to control the main carousel

### Actual Behavior
Click in thumbs and nothing happens

#### Edit 08/12/2020
I've [recorded an loom](https://www.loom.com/share/ad358f835c6d43bcae3632a5c2ae7ec6) and [create a codesandbox for this](https://codesandbox.io/s/vigilant-lamarr-7fvkn?file=/src/App.js)

spgwebdev:
can confirm the issue is still present in 6.3.3

ygj6:
@DCitalo I've used the `master` branch test and there's no problem.

DCitalo:
@ygj6 I've used the master as well, as you can see in the codesandbox I posted in 14 Oct. 
I've record a loom with the error, if you can see in the release `6.2.0` it was working but in the release `=> 6.3.0` it is not working:
https://www.loom.com/share/ad358f835c6d43bcae3632a5c2ae7ec6

Btw in the last release (`6.4.0`) it still not working

code-mat:
it doesn't work still in `6.4.1` :(

valnub:
I can confirm this issue still exists in 6.4.5, see https://github.com/valnub/swiper-react-demo/issues/4

As Swiper is open source let's have a look at the source code. There must have been a change somewhere between version 6.0.2 and 6.3.0. If we can find the problem maybe we can send in a PR to fix it.

nolimits4web:
For me it seems like some bug in React, if you disable StrictMode it works.
Replace this:
```
ReactDOM.render(
  <React.StrictMode>
    <App />
  </React.StrictMode>,
  rootElement
);
```
with this:
```
ReactDOM.render(
  <App />,
[...]

PR #4440, opened 2021-04-13
Make swiper update for virtual mode React.StrictMode compatible

TheAnkurPanchani:
Due to react strict mode additional checks, swiper was not being updated correctly, specifically in virtual mode. Code change of the PR will make sure that it works on strict mode without any issues.

You can replicate issue by creating react virtual swiper and adding slides dynamically in development with react strict mode enabled. After these changes, it seems to be working just fine in development with react strict mode.

### DIFF
--- a/src/react/swiper.js
+++ b/src/react/swiper.js
@@ -39,16 +39,6 @@ const Swiper = forwardRef(
 
     const { slides, slots } = getChildren(children);
 
-    const changedParams = getChangedParams(
-      passedParams,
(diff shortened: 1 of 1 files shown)


nolimits4web:
Awesome thanks! I wasn't able to figure out why it doesn't work correctly in strict mode.

Also:
fixes #3863
fixes #3852

valnub:
This is awesome! Thanks so much!! :-)

PR #4514, opened 2021-05-02
Fix control type in ControllerOptions interface

wwwebman:
Regarding the documentation `control` could be  Swiper instance or array with Swiper instances.
Read more - https://swiperjs.com/swiper-api#controller-methods.


nolimits4web:
Thanks!

PR #4533, opened 2021-05-08
feat(react): add the missing render function type

yoychen:
Add the missing type for [SwiperSlide render function](https://swiperjs.com/react#swiper-slide-render-function) of React,

### DIFF
--- a/src/swiper-react.d.ts
+++ b/src/swiper-react.d.ts
@@ -27,6 +27,14 @@ interface Swiper extends SwiperOptions {
   // MODULES_EVENTS
 }
 
+interface SlideData {
+  isActive: boolean;
+  isVisible: boolean;
+  isDuplicate: boolean;
+  isPrev: boolean;
+  isNext: boolean;
+}
+
 interface SwiperSlide {
   /**
    * Slide tag
@@ -48,6 +56,13 @@ interface SwiperSlide {
    * @default false
    */
   virtualIndex?: number;
+
+  /**
+   * Slide's child element or render function
+   *
+   * @default undefined
+   */
+  children?: React.ReactNode | ((slideData: SlideData) => React.ReactNode);
 }
 
 interface Swiper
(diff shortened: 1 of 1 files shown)


src/swiper-react.d.ts
```diff
@@ -66,6 +81,6 @@ interface Swiper
 interface SwiperSlide extends React.HTMLAttributes<HTMLElement> {}
 
 declare const Swiper: React.FunctionComponent<Swiper>;
-declare const SwiperSlide: React.FunctionComponent<SwiperSlide>;
+declare const SwiperSlide: React.VoidFunctionComponent<SwiperSlide>;
```
nolimits4web, line 84: Wondering why `React.VoidFunctionComponent` here and not `React.FunctionComponent`?

yoychen: Sorry for the late reply, and thanks for giving me a chance to contribute to this awesome project. 

Because the difference between `FunctionComponent` and `VoidFunctionComponent` [1] is that the props type of `FunctionComponent` is wrapped with generic  `PropsWithChildren` . The `PropsWithChildren` [2] will declare the type of children to `ReactNode`, so we can use the render function feature (the type declaring in interface `SwiperSlide`) to declare the slide content when use React with TypeScript.

1.
    ```ts
    interface FunctionComponent<P = {}> {
        (props: PropsWithChildren<P>, context?: any): ReactElement<any, any> | null;
        propTypes?: WeakValidationMap<P>;
        contextTypes?: ValidationMap<any>;
        defaultProps?: Partial<P>;
        displayName?: string;
    }

    interface VoidFunctionComponent<P = {}> {
        (props: P, context?: any): ReactElement<any, any> | null;
        propTypes?: WeakValidationMap<P>;
        contextTypes?: ValidationMap<any>;
        defaultProps?: Partial<P>;
        displayName?: string;
    }
    ```

2.
    ```ts
    type PropsWithChildren<P> = P & { children?: ReactNode };
    ```
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts

nolimits4web: Got it, thanks for your help and contribution!

Issue #4553, opened 2021-05-14
Swiper React scss interpolation build error


lukeMWI:
- Swiper Version: 6.6.1
- Platform/Target and Browser Versions: MacOS Chrome 90.0.4430.212 

## What You Did

Originally tried to use non-react version and switched to react version within NextJS project. Now getting Sass errors

```
WARNING: You probably don't mean to use the color value aqua in interpolation here.
It may end up represented as cyan, which will likely produce invalid CSS.
Always quote color names when using them as strings or map keys (for example, "aqua").
If you really want to use the color value here, use '"" + $paginationColorName'.

    ╷
147 │   .swiper-pagination-#{$paginationColorName} {
    │                        ^^^^^^^^^^^^^^^^^^^^
    ╵
    node_modules/swiper/components/pagination/pagination.scss 147:24  @import
    src/styles/theme/style.scss 606:9                                 root stylesheet
```

## Expected Behavior

Project Builds

## Actual Behavior

Build fails

PR #4554, opened 2021-05-14
Fix scss interpolation error

lukeMWI:
Fixes nolimits4web/swiper#4553

Currently the implementation of scss interpolation causes a build error. This PR updates the syntax to prevent that.

### DIFF
--- a/src/components/pagination/pagination.scss
+++ b/src/components/pagination/pagination.scss
@@ -144,8 +144,8 @@
   }
 }
 @each $paginationColorName, $paginationColorValue in $colors {
-  .swiper-pagination-#{$paginationColorName} {
-    --swiper-pagination-color: #{$paginationColorValue};
+  .swiper-pagination-#{'' + $paginationColorName} {
+    --swiper-pagination-color: #{'' + $paginationColorValue};
   }
 }
 .swiper-pagination-lock {


lukeMWI:
@nolimits4web Is there any chance I could get some eyes on this since this bug prevents CI from working? Hopefully this PR is simple enough that is doesn't need much of your time.

nolimits4web:
Thanks! But why it doesn't work for you and works for everyone else as far as I know. Even CodeSandbox works fine https://codesandbox.io/s/swiper-react-pagination-forked-c8rq0?file=/src/App.js . Maybe you have some `sass`/`scss` outdated?

Jared-Dev:
Hi @nolimits4web! 👋🏻 

I can't speak as to why we are the ones seeing this. The only thing I know to pass along is that this is the recommendation for using colors within string interpolation!
https://github.com/sass/sass/issues/1687


Apparently, (this was new info for me) this stems from the fact that colors in SCSS are objects! TIL!

The other question I have is there a potential issue that you'd forsee in using the updated syntax since it is the recommendation that the `sass` repo recommends?

_They do present another alternative which is to put the keys in the source list within quotes, but I haven't looked at that list to see how the code is currently written._

nolimits4web:
Merged, thanks!

Issue #3386, opened 2019-12-30
Swiper cube loop problem


Micu22:
* [x] bug

* Platform/Target and Browser Versions: both chrome 79 and edge produce a problem
* Live Link or JSFiddle/Codepen or website with isssue: [https://biiird.com/slider/](url) (there are two swipers, the issue is about the cube one

### What I did
I have created a cube slider with the parameter loop: true,
The slider glitches between the first and the last slide when swiping backwards
Here is the swiper code:
`var swiper = new Swiper( '.custom-swiper > .elementor-container', {
        speed: 400, 
        loop: true, 
        parallax: true, 
        effect: 'cube',
        grabCursor: true,
        cubeEffect: {
            shadow: true,
            slideShadows: false,
            shadowOffset: 20,
            shadowScale: 0.94,
        },
        navigation: {
            nextEl: '.custom-swiper .swiper-button-next',
            prevEl: '.custom-swiper .swiper-button-prev',
        },`

### Expected Behavior
slides going in a loop without glitches

### Actual Behavior
glitch between the first and last slide when sliding backwards

![image](https://user-images.githubusercontent.com/51614436/71597236-a9831f00-2b42-11ea-94eb-6e4df92a0e26.png)
![image](https://user-images.githubusercontent.com/51614436/71597254-ba339500-2b42-11ea-9334-deda68592c08.png)
![image](https://user-images.githubusercontent.com/51614436/71597279-ce779200-2b42-11ea-87b5-772c069cbea8.png)

Yi98:
Can you provide the html code? I would like to replicate your problem :)

Micu22:
You can see the problem here on codepen: 
[https://codepen.io/iamrc1/pen/MdGKYQ](url)
I can provide my code, but this is simpler and shows the same problem.
When going from first slide directly to the last one by draging the animation glitches.

Yi98:
I found out that there must be at least 4 swiper-slide components inside a swiper-wrapper for it behave normally. In your case, you have only 3 swiper-slide components which is also why the unexpected behavior occurred. This is indeed a bug that need to be fixed. For now, you may choose to add on 1 more swiper-slide to temporarily solve the issue :)

Gumbraise:
Still here

Dupflo:
Still there

EmanuelOcampo:
still there

Issue #1772, opened 2016-07-05
[Bug] loop on cube effect


zhaoyao91:
If I set effect as cube and enable loop, and then slide it reversely, after the first image, the pagination and the 'inner mode' should be at the last image, but the displayed slide is the still the first image.

```
 this.swiper = new Swiper(this.refs.container, {
            pagination: this.refs.pagination,
            observer: true,
            autoplay: 5000,
            autoplayDisableOnInteraction: false,
            loop: true, // there is bug, see
            paginationClickable: true,
            preventClicks: false,
            effect: 'cube'
        })
```

![2016_07_05_15_48_42](https://cloud.githubusercontent.com/assets/3808838/16577481/1edcc0f6-42c8-11e6-943c-f7737b9b7186.gif)

nolimits4web:
Hmm, can't replicate it. Can you please post link to example with the issue if it is still actual

zhaoyao91:
sorry, this project has been updated for a long time, so I don't have any demo for now

pe6o:
I entered the number of slides in the `loopedSlides` setting and it works. Without it, I have the same problem.

nolimits4web:
Issue is closed because of outdated/irrelevant/not actual/needed more information/inactivity.

If this issue is still actual and reproducible for latest version of Swiper, please create new issue and fill the issue template correctly:
* Clearly describe the issue including steps to reproduce when it is a bug.
* Make sure you fill in the earliest version that you know has the issue.
* Provide live link or JSFiddle/Codepen or website with issue

Issue #1752, opened 2016-06-16
In 3d Cube animation last slide image or first slide image dublicates.


Cyclops0571:
In saphari when i go back from first slider to last slider with sliding left to right first image shows up instead of last image.
In other browsers there is not such a problem.
Here is the sample for the problem.
http://www.galepress.com/banners/service_view/683

datamann10:
i can second this
http://danian.website/photography/

nolimits4web:
Issue is closed because of outdated/irrelevant/not actual/needed more information/inactivity.

If this issue is still actual and reproducible for latest version of Swiper, please create new issue and fill the issue template correctly:
* Clearly describe the issue including steps to reproduce when it is a bug.
* Make sure you fill in the earliest version that you know has the issue.
* Provide live link or JSFiddle/Codepen or website with issue

PR #4566, opened 2021-05-17
3-side looped cube face overlapping fix

adriangornypl:
The changes applies to the behaviour of duplicated slides (from 3rd to 1st and from 1st to 3rd) for the issue shown with 3-sides cube with loop option enabled. This issue has been previously reported although not followed, the following fix, should resolve these edge-cases that only happen when:

1. Swiper has 3 slides only
2. The loop option is enabled
3. The "Cube effect" is activated.

Related issues:
https://github.com/nolimits4web/swiper/issues/3386
https://github.com/nolimits4web/swiper/issues/1772
https://github.com/nolimits4web/swiper/issues/1752

Probably this can be done better however this solves the issue in a pretty easy way.

### DIFF
--- a/src/components/effect-cube/effect-cube.less
+++ b/src/components/effect-cube/effect-cube.less
@@ -56,4 +56,12 @@
       filter: blur(50px);
     }
   }
+  .swiper-slide-prev:not(.swiper-slide-duplicate-next) {
+    pointer-events: auto;
+    visibility: visible;
+  }
+  .swiper-slide-prev.swiper-slide-duplicate-next.swiper-slide-duplicate {
+    pointer-events: none;
+    visibility: hidden;
+  }
 }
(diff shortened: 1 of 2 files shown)


nolimits4web:
Hi! Thanks for your PR, I checked it and it actually doesn't solve issue much for me, it is still kinda "glitching". What does solve it is just by setting `loopedSlides: 3` parameter

adriangornypl:
> Hi! Thanks for your PR, I checked it and it actually doesn't solve issue much for me, it is still kinda "glitching". What does solve it is just by setting `loopedSlides: 3` parameter

Hmm let me check it then and if not I'll create a codepen of the issue this one solves.

adriangornypl:
It seems that loopedSlides also works :) Closing this one then, thanks @nolimits4web!

Issue #4682, opened 2021-06-16
Safari Scrollbar Media Query


mariomc:
- Swiper Version: v6.7.0 (happens with all, really 😅 )
- Platform/Target and Browser Versions: macOS Safari (14.1.1) but happens with all. Only happens when the page has vertical overflow AND the user has a scrollbar setting on "always visible" or has a mouse connected

## What You Did

When in visible scrollbar mode, macOS Safari discounts the width of the scrollbar to the window.innerWidth for its own media queries.
This creates a dissonance between swiper and how media queries are enacted for those devices

https://stackblitz.com/edit/swiper-demo-1-default-c22ogy?file=index.html

## Expected Behavior

The logic used by Swiper breakpoint election should mirror the CSS media queries behaviours in all devices

## Actual Behavior

To test this out, go to `System Preferences > General ` and either turn the "Always" option ON or have a mouse connected with "Automatic based on mouse or trackpad" (default for macOS)

<img width="513" alt="Screenshot 2021-06-16 at 10 41 51" src="https://user-images.githubusercontent.com/887163/122196829-7b58e280-ce8f-11eb-968c-24ce281b5788.png">

After that is done use Safari to go to: https://swiper-demo-1-default-c22ogy.stackblitz.io

When in a resolution around 15px (or the width of the scrollbar) of the edge of the breakpoints you can notice the mismatch between an element shown using CSS media queries and the breakpoints used in Swiper.

https://user-images.githubusercontent.com/887163/122198365-fff83080-ce90-11eb-8e1a-67db0a3150fe.mov

I'll open a PR fixing the issue soon after. 😄

PR #4683, opened 2021-06-16
Fixes safari scroll issue

mariomc:
Fixes issue: https://github.com/nolimits4web/swiper/issues/4682

If merged, this will use the [window.matchMedia API](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) instead of doing a comparison with the window.innerWidth, which can be flakey in some devices.

This also attains the benefit of making the breakpoint election closer to the CSS counterpart in media queries.

Browser support for window.matchMedia is wide ( https://caniuse.com/matchmedia ) and has no known gotchas associated (apart from using addEventListener on it in Safari, which we don't need).


nolimits4web:
Merged, thank you!

PR #4692, opened 2021-06-21
fix(core): fixed __proto__ pollution

hydRAnger:
FIX this vulnerability for Swiper V5: 
- https://www.npmjs.com/advisories/1694
- https://snyk.io/vuln/SNYK-JAVA-ORGWEBJARS-1244698

refs:
- https://github.com/nolimits4web/swiper/commit/9dad2739b7474f383474773d5ab898a0c29ac178

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -87,7 +87,7 @@ const Utils = {
     for (let i = 1; i < args.length; i += 1) {
       const nextSource = args[i];
       if (nextSource !== undefined && nextSource !== null) {
-        const keysArray = Object.keys(Object(nextSource));
+        const keysArray = Object.keys(Object(nextSource)).filter((key) => key !== '__proto__');
         for (let nextIndex = 0, len = keysArray.length; nextIndex < len; nextIndex += 1) {
           const nextKey = keysArray[nextIndex];
           const desc = Object.getOwnPropertyDescriptor(nextSource, nextKey);


lobosan:
Does this also need to check for `constructor` and `prototype` like this?
https://github.com/ionic-team/ionic-framework/pull/23344/files

nolimits4web:
I am afraid that will required a more work and not extend '__proto__', 'constructor', 'prototype' props. Plus it needs correctly reference other Swiper instances (required for Thumbs and Controller to work correctly). To be honest last Swiper v5 was released almost a year ago, and I didn't plan new v5 releases for sure. Why just not copy v5 package locally and do the local changes instead?

lobosan:
@nolimits4web @hydRAnger I thought the one-line code was the fix in version 6 and it seems it would work for v5. Is it possible to deploy a new v5 with this change?

hydRAnger:
> @nolimits4web @hydRAnger I thought the one-line code was the fix in version 6 and it seems it would work for v5. Is it possible to deploy a new v5 with this change?

according to your suggestion, what should I do for now?

kashifshamaz21:
@hydRAnger @lobosan I'm also trying to do a local patch of v5.4.5 Swiper with the fix for the `proto` pollution. did you guys just do a patch similar to 
`core: fixed proto pollution` ([ec358de](https://github.com/nolimits4web/Swiper/commit/ec358deab79a8cd2529465f07a0ead5dbcc264ad)) that was done on v6.5.1 ?

PR #4707, opened 2021-06-27
Add Vue component typings

### LINKED ISSUE #3916: swiper/vue missing type declarations
This is a (multiple allowed):
* [ ] bug
* [x] enhancement

I see that Swiper has already types for `swiper/react` component included. I would be really cool to add type definitions for `swiper/vue` component as well. This would allow us to integrate swiper in vue.js projects with typescript easier.

Screenshot 2020-11-03 at 8 35 50

I'm willing to work on this feature with a little guidance from maintainers.

* Swiper Version: v3.6.5
* Platform/Target and Browser Versions: Typescript

MrTimscampi:
This adds proper TypeScript definitions to the Vue components.

I'm not fully sure everything is 100% good, but it was tested on another project and seems to work fine there, so I figured I'd upstream it.

~~It is probably missing some things like the emits, but it has all the (current) attributes defined.~~ Emits have been added, so it should have everything now.

I was a bit lost with the multiple package.json files in multiple places, so feel free to correct me if I messed something up :smile:

Fixes #3916 

**Edit:** I went over the typings again and fixed a bunch of props, all the defaults and added typings for all the emits/event callbacks.  
This should now be pretty much complete.

### DIFF
--- a/src/swiper-vue.d.ts
+++ b/src/swiper-vue.d.ts
@@ -1,0 +1,567 @@
+import {
+  A11yOptions,
+  AutoplayOptions,
+  ControllerOptions,
+  CoverflowEffectOptions,
+  CubeEffectOptions,
+  FadeEffectOptions,
+  FlipEffectOptions,
+  HashNavigationOptions,
+  HistoryOptions,
+  KeyboardOptions,
+  LazyOptions,
+  MousewheelOptions,
+  NavigationOptions,
+  PaginationOptions,
+  ParallaxOptions,
+  ScrollbarOptions,
+  ThumbsOptions,
+  VirtualOptions,
+  ZoomOptions,
+} from './types/public-api';
+import { ComponentOptionsMixin, DefineComponent, PropType } from 'vue';
+import { SwiperOptions } from './swiper';
+import SwiperClass from './types/swiper-class';
+
+declare const Swiper: DefineComponent<
+  {
+    tag: {
+      type: StringConstructor;
+      default: string;
+    };
+    wrapperTag: {
+      type: StringConstructor;
+      default: string;
+    };
+    init: {
+      type: BooleanConstructor;
+      default: undefined;
+    };
+    direction: {
+      type: PropType<SwiperOptions['direction']>;
+      default: SwiperOptions['direction'];
+    };
+    touchEventsTarget: {
+      type: PropType<SwiperOptions['touchEventsTarget']>;
+      default: undefined;
+    };
+    initialSlide: {
+      type: NumberConstructor;
+      default: undefined;
+    };
+    speed: { type: NumberConstructor; default: undefined };
+    cssMode: { type: BooleanConstructor; default: undefined };
+    updateOnWindowResize: {
+      type: BooleanConstructor;
+      default: undefined;
+    };
+    resizeObserver: {
+      type: BooleanConstructor;
+      default: undefined;
+    };
+    nested: { type: BooleanConstructor; default: undefined };
+    focusableElements: {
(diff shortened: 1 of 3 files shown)


MrTimscampi:
Side note, I originally wanted to port the components to TypeScript, to get automatic typings, but I couldn't figure out how to add TypeScript support to the current build system.

If someone is willing to help to get TypeScript up and running with the existing system, porting the Vue components to TypeScript should be fairly straightforward and will yield better typings than this.

MrTimscampi:
Drafting this while I look at some of the properties, as some are incorrect.

src/swiper-vue.d.ts
```diff
@@ -0,0 +1,567 @@
...
+      default: undefined;
+    };
+  },
+  () => JSX.Element,
+  unknown,
+  {},
+  {},
+  ComponentOptionsMixin,
+  ComponentOptionsMixin,
+  {
```
nolimits4web, line 471: I think events here are in totally same format as `SwiperEvents` in `./types/swiper-events.d.ts`, maybe we can reuse it here?

MrTimscampi: I can't seem to make it work with `SwiperEvents`, as it's complaining about it not respecting a constraint for `EmitsOptions`.

Using something like `[event in keyof SwiperEvents]?: SwiperEvents[event]` gets rid of the errors, but then all the events disappear from auto-completions provided by Vetur or Volar.

They are pretty much the same though, so it'd be good to figure out a way to use them here 😕

nolimits4web:
Merged, thanks! I will add build pipeline

vltansky:
> Side note, I originally wanted to port the components to TypeScript, to get automatic typings, but I couldn't figure out how to add TypeScript support to the current build system.
> 
> If someone is willing to help to get TypeScript up and running with the existing system, porting the Vue components to TypeScript should be fairly straightforward and will yield better typings than this.

@nolimits4web maybe migrating the source to TS should be considered?

nolimits4web:
> > Side note, I originally wanted to port the components to TypeScript, to get automatic typings, but I couldn't figure out how to add TypeScript support to the current build system.
> > If someone is willing to help to get TypeScript up and running with the existing system, porting the Vue components to TypeScript should be fairly straightforward and will yield better typings than this.
> 
> @nolimits4web maybe migrating the source to TS should be considered?

I think yes, will move Vue to .ts in upcoming releases

JonathanDn:
Hey @MrTimscampi and @nolimits4web I browsed through the correspondense here and the code merged in this PR and had hopes that it would help my case of Vue having a hard time to inffer types inside a `Vue.extend({...})` component's `$refs`. 

Any clue to what could help ? I am getting the following error:

```
TS2339: Property '$swiper' does not exist on type 'Vue | Element | (Vue | Element)[]'.
  Property '$swiper' does not exist on type 'Vue'.
```


What I had done:
* Upgraded Swiper.js to above version 6.5.2(I assume that is when vue types were added based on this merged PR)
* Installed `"@types/swiper": "5.4.3"` types in hopes this will help

(attached image shows the culprit throwing the type error)

I would appreciate any best practices how to tackle this.

image

MrTimscampi:
> it would help my case of Vue having a hard time to infer types inside a `Vue.extend({...})` component'

I have usually had to explicitly type all refs in components (using `topRef.$swiper as Swiper` or something similar, for Swiper).

JonathanDn:
Yeah I though about doing the same. @MrTimscampi do you know in which module I can possibly import the interface to than type with `as Swiper` , like what does the import look like for `Swiper` type?

PR #4732, opened 2021-07-02
refactor(touchevent) -  improve shadow swiping class check

weiz18:
Previously, we modified the targetEl to ensure `element.closest(swipingClass)` to always finding the correct element. But it does not work cross browser since event path is different from safari to chrome. This refactor introduce a new method to find the closest matching element with nested shadowRoot, so its now working the same way as the non shadow components. 

So swipe action trigger inside shadow-component will be blocked if either `<Element>` or `<Inner-element>` contains no swiping class. 
`<Element class="no-swiping">`
  `<Inner-element class="no-swiping">`
   `<shadow-comp />`
`     <Inner-element>`
`</Element>`

### DIFF
--- a/src/components/core/events/onTouchStart.js
+++ b/src/components/core/events/onTouchStart.js
@@ -2,6 +2,21 @@ import { getWindow, getDocument } from 'ssr-window';
 import $ from '../../../utils/dom';
 import { extend, now } from '../../../utils/utils';
 
+// https://stackoverflow.com/questions/54520554/custom-element-getrootnode-closest-function-crossing-multiple-parent-shadowd
+function closestElement(selector, base = this) {
+  function __closestFrom(el) {
+      if (!el || el === document || el === window)
+          return null;
+      if (el.assignedSlot)
+          el = el.assignedSlot;
+      let found = el.closest(selector);
+      return found
+          ? found
+          : __closestFrom(el.getRootNode().host);
+  }
+  return __closestFrom(base);
+}
+
 export default function onTouchStart(event) {
   const swiper = this;
   const document = getDocument();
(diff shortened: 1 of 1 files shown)


nolimits4web:
Hey @weiz18, thanks! But can you give me some minimal live example of what it fixes (where now it will be broken)? Just for better understanding what we are fixing here :)

weiz18:
Sure. will attach a minimal live example when i have some time.  Basically, current implementation of finding the closest target on shadow root only works with one level. So if you have a deep nested dom structure where the no swiping class is on the outer most element, it will fail to find it in the path.  
<div class="swiper-no-swiping"><shadow-comp><shadow-comp-2></shadow-comp-2></shadow-comp></div>,

nolimits4web:
Still not sure what is the issue, because right now tested the following:
``` html
    <div class="swiper-slide">
      <span class="swiper-no-swiping" style="display: inline-block;">
        <custom-1>
          <custom-2>
            <p style="line-height: 1;">Hello world</p>
          </custom-2>
        </custom-1>
      </span>
    </div>
```
And everything works as expected for me in Safari, Chrome and FF. There is no swiping  over this elements

weiz18:
Hey, @nolimits4web  sorry for the late response, was away for a couple of days.  I have attached a demo reproducing the issue. https://jsfiddle.net/qhzwom26/31/ . as you can see in the demo, the child element is still moving despite no-swiping class being set on the parent element.  I think the key here is to set the shadow root mode to `open` in order to reproduce.  let me know if you need more info. 
thanks

nolimits4web:
Merged, thanks!

weiz18:
nice. thanks a lot

PR #4750, opened 2021-07-11
feat(pagination): hide 1 dot pagination

### LINKED ISSUE #2572: Request - hide pagination dots if only 1 "page"
This is a (multiple allowed):
* [] bug
* [x] enhancement

* Swiper Version: 4.2.2.
* Platform/Target and Browser Versions: ALL.
* Live Link or JSFiddle/Codepen or website with issue: N/A

### What you did
* Created a multi-row swiper which has it's contents dynamically updated (slides added/removed).
* On initial load, only 1 "page" may have slides, so the pagination button is a distraction.
* Swipers may or may not have slides dynamically added as the user interacts with the application

### Expected Behavior
* If a swiper contains only 1 "page", dynamically hide the pagination dots.
* If more slides are dynamically added (e.g lazy loaded) and this results in total "pages" > 1, dynamical add the pagination buttons.
* If total "pages" > 1 and slides are destroyed, evaluate and dynamically hide the pagination dots.

### Actual Behaviour
If pagination is enabled, page dots are shown, even if only 1 "page" is visible (aka no swiping) - this decreases the usability and is likely to confuse end users - after all, pagination is only useful if there are multiple pages 😃

vltansky:
closes #2572

### DIFF
--- a/src/components/pagination/pagination.less
+++ b/src/components/pagination/pagination.less
@@ -77,6 +77,10 @@
   .swiper-pagination-clickable & {
     cursor: pointer;
   }
+
+  &:only-child {
+    display: none;
+  }
 }
 .swiper-pagination-bullet-active {
   opacity: 1;
--- a/src/components/pagination/pagination.scss
+++ b/src/components/pagination/pagination.scss
@@ -69,7 +69,7 @@
   }
 
   &:only-child {
-    visibility: hidden;
+    display: none;
   }
 }
 .swiper-pagination-bullet-active {
--- a/src/components/pagination/pagination.scss
+++ b/src/components/pagination/pagination.scss
@@ -67,6 +67,10 @@
   .swiper-pagination-clickable & {
     cursor: pointer;
   }
+
+  &:only-child {
+    visibility: hidden;
+  }
 }
 .swiper-pagination-bullet-active {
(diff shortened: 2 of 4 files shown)


src/components/pagination/pagination.less
```diff
@@ -77,6 +77,10 @@
   .swiper-pagination-clickable & {
     cursor: pointer;
   }
+
+  &:only-child {
+    display: none;
```
nolimits4web, line 82: `display: none` here

src/components/pagination/pagination.scss
```diff
@@ -67,6 +67,10 @@
   .swiper-pagination-clickable & {
     cursor: pointer;
   }
+
+  &:only-child {
+    visibility: hidden;
```
nolimits4web, line 72: `visibility: hidden;` here? 😀

nolimits4web (changes requested):
I that is so simple :D let's make it for v6 then :)

src/components/pagination/pagination.scss
vltansky: you're right

vltansky:
closed in favor of #4786 (that target `master` instead of `v7`)

Issue #4781, opened 2021-07-21
Count aria-label without duplicated item in loop option


dpxgit:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://swiperjs.com/demos#infinite-loop

### Bug description

If the loop option is set, the aria-label count all DOM elements including the duplicated. There are 5 elements, but 7 are counted by the duplicates. 

### Expected Behavior

If the loop option is set, the aria label may only be attached to the non-duplicated elements.  Only the elements that are not duplicated are counted.

### Swiper version

6.7.0

### Platform/Target and Browser Versions

Independent of platform/browser

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

PR #4782, opened 2021-07-21
fix(a11y): Count aria-label without duplicated item in loop option

dpxgit:
If the loop option is set, the aria label may only be attached to the non-duplicated elements. Only the elements that are not duplicated are counted. 

FIX: #4781

### DIFF
--- a/src/components/a11y/a11y.js
+++ b/src/components/a11y/a11y.js
@@ -160,12 +160,19 @@ const A11y = {
       swiper.a11y.addElRoleDescription($(swiper.slides), params.itemRoleDescriptionMessage);
     }
     swiper.a11y.addElRole($(swiper.slides), params.slideRole);
-    swiper.slides.each((slideEl) => {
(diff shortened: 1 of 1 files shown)


nolimits4web:
Thanks for your PR! Don't you think maybe it would be better if duplicated slides will have same aria label index as their original copies? You can see in loop mode each slide has `data-swiper-slide-index` attribute with "realIndex"

dpxgit:
The `aria-label` is usefull for screenreader to tell the user which slide is active and how many slider there are. But on duplicated slides we have the attribute `aria-hidden=true` this tell the screenreader that he shouldn't read it.

nolimits4web:
We don't set aria-hidden on duplicated slides, so I made what I suggeted

dpxgit:
Hey, thats correct, it was my fault, we build in our project the aria-hidden option. Thanks for fixing. 
A PullRequest for aria-hidden will follow. Many Thanks.

PR #4788, opened 2021-07-22
fix: fix server side judge (close #4787)

### LINKED ISSUE #4787: bug: isNode judge is inaccurate in server side rendering

- [X] I confirm

### Reproduction link

https://github.com/zhangyuang/swiper-bug-isnode

### Bug description

```js
// src/utils/utils.js
function isNode(node) {
  // eslint-disable-next-line
  if (typeof window !== 'undefined') {
    return node instanceof HTMLElement;
  }
  return node && (node.nodeType === 1 || node.nodeType === 11);
}
```

The judge is very inaccurate, in server side render, framework self or users maybe mock a window object in server side is frequent

Maybe there should judge at a deeper level。for example, `window.document`,`window.document.documentElement` or `window.screen`

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Swiper version

6.8.0

macos

### Validations

zhangyuang:
Mock window object is very frequent in server side rendering, the `isNode` judge is inaccurate。
Maybe we should judge further

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -94,7 +94,7 @@ function isObject(o) {
 }
 function isNode(node) {
   // eslint-disable-next-line
-  if (typeof window !== 'undefined' && typeof window.screen !== 'undefined') {
+  if (typeof window !== 'undefined' && typeof window.HTMLElement !== 'undefined') {
     return node instanceof HTMLElement;
   }
   return node && (node.nodeType === 1 || node.nodeType === 11);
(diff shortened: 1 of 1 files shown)


zhangyuang:
@vltansky please see see

nolimits4web:
I would say checking `window.screen` is not right as it has nothing to do with the logic, I think better to check what we really need is `window.HTMLElement`

zhangyuang:
fine, has been modified

nolimits4web:
Merged, thanks!

PR #4835, opened 2021-08-18
fix(angular): demo

vltansky:


src/types/swiper-class.d.ts
```diff
@@ -304,10 +316,14 @@ interface Swiper extends SwiperClass<SwiperEvents> {
    */
   attachEvents(): void;
 
+  loopCreate(): void;
```
nolimits4web, line 319: Let's add `!INTERNAL` flag to loopedSlides, loopCreate and loopDestroy props. Don't want them to appear in docs

PR #4884, opened 2021-08-28
fix(a11y): fix argments for `initNavEl()` method

yuheiy:
When I using the a11y module, the `aria-label` and `aria-controls` values applied to the navigation buttons are incorrect.

Reproduction code:

```javascript
import Swiper, { A11y, Navigation } from "swiper";

new Swiper(".swiper", {
  modules: [A11y, Navigation],
  a11y: {},
  navigation: {
    nextEl: ".swiper-button-next",
    prevEl: ".swiper-button-prev",
  },
});
```

```html
<div class="swiper">
  <div class="swiper-wrapper">
    <div class="swiper-slide">Slide 1</div>
    <div class="swiper-slide">Slide 2</div>
    <div class="swiper-slide">Slide 3</div>
  </div>
  <button class="swiper-button-prev" type="button">Previous</button>
  <button class="swiper-button-next" type="button">Next</button>
</div>
```

Result:

```html
<button class="swiper-button-prev swiper-button-disabled" disabled="" tabindex="-1" aria-label="undefined" aria-controls="Previous slide" aria-disabled="true">Previous</button>
<button class="swiper-button-next" tabindex="0" aria-label="undefined" aria-controls="Next slide" aria-disabled="false">Next</button>
```

The reason for `aria-label="undefined"`, `aria-controls="Previous slide"` and `aria-controls="Next slide"` is that the arguments given to `initNavEl()` are wrong.

### DIFF
--- a/src/modules/a11y/a11y.js
+++ b/src/modules/a11y/a11y.js
@@ -209,10 +209,10 @@ export default function A11y({ swiper, extendParams, on }) {
     }
 
     if ($nextEl && $nextEl.length) {
-      initNavEl($nextEl, params.nextSlideMessage);
+      initNavEl($nextEl, wrapperId, params.nextSlideMessage);
     }
     if ($prevEl && $prevEl.length) {
-      initNavEl($prevEl, params.prevSlideMessage);
+      initNavEl($prevEl, wrapperId, params.prevSlideMessage);
     }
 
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thanks!

Issue #2875, opened 2018-11-13
Back button doesn't work properly when using loop, slidesPerView and slidesPerGroup


apriljunge:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Swiper Version: 4.4.2
* Platform/Target and Browser Versions: Windows (all major browsers)
* Live Link or JSFiddle/Codepen or website with isssue: https://codepen.io/anon/pen/pQRLEr

### What you did
I hit the back arrow multiple times, with the config you can see in the codepen.

### Expected Behavior
The back arrow should change between the groups ( Group 1, Group 2, Group 1, ...)

### Actual Behavior
The back arrow always shows group 2 (Slide 3, 4, 5) ( Group 2, Group 2, Group 2, ...)

AgamlaRage:
Same issue for me,  with **loop** and any combinaison of **slidesPerView** and **slidesPerGroup**, when **slidesPerGroup** > 1.

Going backward seems to be bugged.

Items are not displayed as expected.

tomaszmittlener:
I have the same issue with the newest version of swiper (4.5.0) . It seems that the problem appears when you try to go to previous group of slides when on the last view with **even** number of slides and below **settings**:

```
slidesPerView: 3,
slidesPerGroup: 3,
loop: true,
```

Codepen:

https://codepen.io/tmittlener/pen/BeOqBK

apriljunge:
Please do not close this issue.
This bug still exists with 5.2.0

tom-schneider-aw:
Is there any update on this issue?

apriljunge:
This bug still exists in 6.2.0
I updated [the codepen](https://unpkg.com/swiper/swiper-bundle.min.css)

vltansky:
updated demo to the latest version:
https://codesandbox.io/s/ecstatic-browser-4ofod?file=/index.html

Bug confirmed

johnson0903:
It happen when execute `slideTo` function in **loop/loopFix.js**. As far as I know, the purpose of execute slideTo when `activeIndex < loopedSlides` is to reset the track position from duplicate slide to real slide position.
But when total slides cannot devide by slidePerGroups, it will snap to wrong slide , and then it will slide one more time after execute `loopFix()` in **slide/slidePrev.js**. 
For example, in the situation @vltansky provideed, that is, 
[...]

PR #4904, opened 2021-09-01
fix slide prev bug when using loop with slidesPerGroup and slidesPerView

johnson0903:
Please see #2875 
I modify index.html in playground for demo usage.
Fix this bug through plus snapIndex 1 when total slides can't divide by slidePerGroup when call slideTo by loopFix

### DIFF
--- a/.github/ISSUE_TEMPLATE/swiper_angular_issue.yml
+++ b/.github/ISSUE_TEMPLATE/swiper_angular_issue.yml
@@ -48,7 +48,7 @@ body:
     attributes:
       label: Swiper version
       description: Exact release version or commit hash
-      placeholder: e.g 6.7.1
+      placeholder: e.g 7.0.0
     validations:
       required: true
   - type: input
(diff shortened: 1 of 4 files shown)


nolimits4web:
Hi! Can you please provide a demo with latest Swiper 7 with the issue this PR fixes?

johnson0903:
Hi, there is it. [https://johnson0903.github.io/swiper/playground/core/](url)

nolimits4web:
Thanks, but as I see with the fix it is still works kind of incorrect, I guess expected behavior is:
Initial: [1, 2, 3]
Press back: [8,9,10]
Press back: [5, 6, 7], but we see here [7,8,9]

Problem is somewhere deeper I think, and also there is a [`loopFillGroupWithBlank`](https://swiperjs.com/swiper-api#param-loopFillGroupWithBlank) which supposed to be used when number of slides not even to slidesPerGroup

johnson0903:
I think if you want to display [5,6,7], you have to update snap grids during fixloop, because the snap grids is set to [8,9,10][1,2,3][4,5,6][7,8,9][10,1,2][1,2,3] in loop mode when initialize swiper in this situation ,there is no [5,6,7] . But if update snap grids, you can't go back to [1,2,3] because it will become [5,6,7] -> [4,3,2]. 
I think maybe change [10,1,2] to [8,9,10] when initialize can be another solution.

PR #4927, opened 2021-09-08
Add pagination bullet CSS variables

Androlax2:
Added some CSS variables for pagination bullet to :

- Manage size (width & height) of a bullet
- Manage background color of inactive bullet
- Manage opacity of inactive color

I might add a variable to manage gap between bullet too.
Something like :

```scss
.swiper-pagination-bullet + .swiper-pagination-bullet {
  margin-left: var(--swiper-pagination-gap, 0px);
}
```

But it might cause issue for RTL

### DIFF
--- a/src/modules/pagination/pagination.less
+++ b/src/modules/pagination/pagination.less
@@ -1,6 +1,14 @@
 :root {
   /*
   --swiper-pagination-color: var(--swiper-theme-color);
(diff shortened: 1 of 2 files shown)


nolimits4web:
Thanks, I also added vertical & horizontal gaps

PR #4947, opened 2021-09-12
chore(ci): run on `Swiper6` branch

vltansky:


nolimits4web:
Why do we need to run CI on Swiper 6? It is supposed to be not maintained

vltansky:
I thought if ever Swiper 6 will need any security patches or similar better it will have CI.

PR #4949, opened 2021-09-12
feat(angular): `enabled` prop

### LINKED ISSUE #4948: Angular [enabled] property
### Discussed in https://github.com/nolimits4web/swiper/discussions/4756

<div type='discussions-op-text'>

<sup>Originally posted by **zygimantas** July 12, 2021</sup>
Is there a reason why [enabled] attribute is missing in Angular library?

In documentation, it is written:

> SwiperComponent props
> Swiper Angular component receive all Swiper parameters as component props

But not "enabled".</div>

vltansky:
closes #4948 #4756

### DIFF
--- a/src/angular/src/swiper.component.ts
+++ b/src/angular/src/swiper.component.ts
@@ -733,7 +733,6 @@ export class SwiperComponent implements OnInit {
           continue;
         }
         const newValue = changedParams[key]?.currentValue ?? changedParams[key];
-        console.log(newValue, key);
         this.updateParameter(key, newValue);
       }
 
@@ -812,7 +811,7 @@ export class SwiperComponent implements OnInit {
     if (_key === 'enabled') {
       if (value === true) {
         this.swiperRef.enable();
-      } else {
+      } else if (value === false) {
         this.swiperRef.disable();
       }
       return;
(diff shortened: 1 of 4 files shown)


src/angular/src/swiper.component.ts
```diff
@@ -732,6 +733,7 @@ export class SwiperComponent implements OnInit {
           continue;
         }
         const newValue = changedParams[key]?.currentValue ?? changedParams[key];
+        console.log(newValue, key);
```
nolimits4web, line 736: remove console.log here :)

src/angular/src/swiper.component.ts
```diff
@@ -807,7 +809,14 @@ export class SwiperComponent implements OnInit {
         this.swiperRef.params[_key] = defaultParams;
       }
     }
-
+    if (_key === 'enabled') {
+      if (value === true) {
+        this.swiperRef.enable();
+      } else {
```
nolimits4web, line 815: I think it should be `} else if (value === false) {` on this line, otherwise I guess it will disable it if it is `undefined` (not passed)

nolimits4web (changes requested):

src/angular/src/swiper.component.ts
vltansky: :)

src/angular/src/swiper.component.ts
vltansky: Done :)

nolimits4web (approved):

PR #4992, opened 2021-09-23
fix: make `swiperRef` prop not required in SwiperSlide vue component

rtabulov:
I believe this is just a bug that should be fixed since `swiperRef` props is *actually* not a required prop. 

it caught my eye thanks to Volar


nolimits4web:
Merged, thanks!

Issue #5006, opened 2021-09-27
Swiper shows slide 2 unexpetedly


PeterBenko:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/s/swiper-infinite-loop-angular-forked-y85mt?file=/src/app/app.component.ts

### Bug description

Under certain circumstances, slide no2 will be shown instead of the correct slide. This seems to be connected to the (now deprecated) [(index)] attribute.
This bug appears with the change from v7.0.5 -> v7.0.6

[This](https://github.com/nolimits4web/swiper/compare/v7.0.5...v7.0.6#diff-bfa627463b910d30122bd97c77f46167402610e987ac7922fb67145fb0104525R825) looks like the offending line. I'm happy to do a PR but am not sure if the line can just be removed or if there's a greater context I'm missing

### Expected Behavior

The slide at the given index should be displayed

### Actual Behavior

Slide no2 is displayed

### Swiper version

7.0.6

### Platform/Target and Browser Versions

tested on iPadOS Safari 15 & Chrome Version 93.0.4577.82 (Official Build) (64-bit)

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

nolimits4web:
Yeah, seems like that line came from local testing. PR to remove it is welcome 🙂

PR #5007, opened 2021-09-27
fix(angular): removed stray setIndex debug code (nolimits4web#5006)

PeterBenko:
Removed the offending code as discussed in #5006


nolimits4web:
Merged, thanks!

PR #5082, opened 2021-10-13
fix: playground demos depts path

linbuxiao:
The introduction method in the example needs to be updated, the one currently used is not declared in `export`, we need to follow the rules of `module` and introduce the package export part.

PS: I'm not familiar with the `angular` part, so I didn't change it. And I don't understand why I made changes to `yarn.lock`, I didn't change anything dependency related.


vltansky:
This is actually for development purposes and not for demos. Here are demos: https://swiperjs.com/demos 
But maybe adding some comments to avoid confusion would be helpful?

nolimits4web:
@vltansky probably just using local dep can work like what I have in TWM https://github.com/tailwind-mobile/tailwind-mobile/blob/master/package.json#L60

linbuxiao:
`file: src` is a good solution. Here I use `file:dist` for a number of reasons.
1. `bundle.css` is missing from `src` because `css` needs to be built.
2. If you use `src`, you need to redefine a `package.json`.
And the above two points are ready in `dist`. But using `dist` also brings new problems.
1. before running `playground`, we need to use `npm run build:prod`. Similarly, `npm run react` can only build the `react` part, which will miss the necessary `css` and cause errors.
2. After generating `dist`, we need to `npm install` again. Because `npm` needs to link `dist` to `npm` itself.

Translated with www.DeepL.com/Translator (free version)

linbuxiao:
@vltansky Please review

Issue #5202, opened 2021-11-16
Preloader CSS animation on Lazy module causing heavy CPU usage


aynzad:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://swiper-lazy-load-images-react-heavy-cpu-usage.vercel.app/

https://github.com/aynzad/swiper-lazy-load-images-react-heavy-cpu-usage

### Bug description

When Swiper with Lazy Images module is used, CPU usage is always high on the page even when there is no activity.

There will be a major issue when a page contains a lot of sliders using Swiper with lazy images that are using a lot of CPU power (in Chrome's Task Manager) and `Style recalcs / sec`, even when there is no activity.

### Expected Behavior

On basic usage like pagination and scrolling, CPU usage and `Style recalcs / sec` should not be noticeable heavy.
On idle (no activity) there should be minimum CPU usage with zero `Style recalcs / sec`

<img width="976" alt="Screen Shot 2021-11-16 at 4 28 01 PM" src="https://user-images.githubusercontent.com/3133497/141990302-66dd2910-7aa6-4c1d-9ed9-d4d0ffc8f9f8.png">

### Actual Behavior

On basic usage like pagination and scrolling, CPU usage and `Style recalcs / sec` are noticeable heavy.
On idle (no activity) there is heavy CPU usage with continuously changing `Style recalcs / sec`

<img width="975" alt="Screen Shot 2021-11-16 at 4 26 30 PM" src="https://user-images.githubusercontent.com/3133497/141989413-dc868578-2f2b-48e1-b3b3-8ae13c693bc5.png">

### Swiper version

7.2.0

### Platform/Target and Browser Versions

macOS 12.0.1, Chrome 92.0.45, Safari 15.1 

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

aynzad:
Possibility related to this old issue https://github.com/nolimits4web/Swiper/issues/1451

nolimits4web:
[...]

PR #5203, opened 2021-11-16
fix(lazy): only activate preloader-spin animation when slide is activ…

aynzad:
Related to the #5202

When Swiper with Lazy Images module is used, CPU usage is always high on the page even when there is no activity.

Solved by removing spinner animation from all non-activated lazy-preloader elements

### DIFF
--- a/src/modules/lazy/lazy.less
+++ b/src/modules/lazy/lazy.less
@@ -14,7 +14,6 @@
   margin-top: -21px;
   z-index: 10;
   transform-origin: 50%;
-  animation: none;
   box-sizing: border-box;
   border: 4px solid var(--swiper-preloader-color, var(--swiper-theme-color));
   border-radius: 50%;
--- a/src/modules/lazy/lazy.less
+++ b/src/modules/lazy/lazy.less
@@ -14,11 +14,15 @@
   margin-top: -21px;
   z-index: 10;
   transform-origin: 50%;
-  animation: swiper-preloader-spin 1s infinite linear;
+  animation: none;
   box-sizing: border-box;
   border: 4px solid var(--swiper-preloader-color, var(--swiper-theme-color));
   border-radius: 50%;
   border-top-color: transparent;
+}
+
+.swiper-slide-active .swiper-lazy-preloader {
+  animation: swiper-preloader-spin 1s infinite linear;
 }
 
 .swiper-lazy-preloader-white {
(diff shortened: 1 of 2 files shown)


src/modules/lazy/lazy.less
```diff
@@ -14,13 +14,17 @@
   margin-top: -21px;
   z-index: 10;
   transform-origin: 50%;
-  animation: swiper-preloader-spin 1s infinite linear;
+  animation: none;
```
vltansky, line 17: Wouldn't just removing `animation` completely work the same way?

vltansky:
Solving #5202 issue? 
What is the CPU usage after this fix?

aynzad:
> Solving #5202 issue? What is the CPU usage after this fix?

mentioned on "Expected Behavior", something near `0.1%`

src/modules/lazy/lazy.less
aynzad: Yes of course

vltansky:
1. Didn't get it, will this change close issue #5202 ?
2. I think animation should be applied to visible slides also, see 
https://swiperjs.com/swiper-api#lazy-loading

> If you use slidesPerView: 'auto' or slidesPerView > 1, then you should also enable watchSlidesProgress and Swiper will load images in currently visible slides

https://codesandbox.io/s/swiper-lazy-load-images-forked-cp29n?file=/index.html

aynzad:
> https://codesandbox.io/s/swiper-lazy-load-images-forked-cp29n?file=/index.html

‍1. Yes, will close this issue
2. Don't get it really, There should not be any animation for invisible slides in any condition.

vltansky:
2.
image

Only this slide is `.swiper-active` while others are `.swiper-visible` (when multiple slides are visible)

aynzad:
> image
> 
> Only this slide is `.swiper-active` while others are `.swiper-visible` (when multiple slides are visible)

Oh I see. I made a change. 

Thank you

vltansky (approved):

vltansky:
@aynzad thanks for your contribution! Your changes released in `v8` 🎉

aynzad:
> @aynzad thanks for your contribution! Your changes released in `v8` 🎉

I'm glad to help

ybiquitous:
Hi. I notice that animation isn't working on the following official demo:
https://swiperjs.com/demos/370-lazy-load-images/core.html

Is it necessary to add an extra configuration about `.swiper-visible`?

PR #5211, opened 2021-11-18
refactor(core/slide/slideTo): remove duplicate code

lucasrleandro:
Method src/core/slide/slideTo.js refactored.
No breaking changes, only conditional optimized and duplicate lines of code removed.

### DIFF
--- a/src/core/slide/slideTo.js
+++ b/src/core/slide/slideTo.js
@@ -172,9 +172,9 @@ export default function slideTo(
   swiper.emit('beforeTransitionStart', speed, internal);
   swiper.transitionStart(runCallbacks, direction);
 
-  if(speed === 0){
-      swiper.transitionEnd(runCallbacks, direction);
-  } else if (!swiper.animating){
+  if (speed === 0) {
+    swiper.transitionEnd(runCallbacks, direction);
+  } else if (!swiper.animating) {
     swiper.animating = true;
     if (!swiper.onSlideToWrapperTransitionEnd) {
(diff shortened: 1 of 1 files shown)


sanderkooger:
@lucasrleandro I have a quick question, I have noticed that if i do a slideTo or SlideToLoop(index, 200, false) somehow events from swipermovement are still fired. 

```
  const handleDividerVisible = (itemId: string) => {
    const dividerIndex = sidebarSwiperData.findIndex((divider) => divider.itemId === itemId);
    console.log(`dividerindex: ${dividerIndex}`);
    if (swiper) {
      // TODO: for this to work we need ti disable Waypoints when scrollTo is active ( LOOP)
      const speed = 400;
      const runCallbacks = false;
      swiper.slideToLoop(dividerIndex, speed, runCallbacks);
    }
  };
  ```
  
  On our swiper component we have something to track realIndex that triggers a .scrollTo liek so 
  ```
        <AppBar sx={classes.sidebarSwiperAppBar}>
          <SidebarSwiper
            sidebarSwiperData={sidebarSwiperData}
            onSwiper={setSwiper}
            onRealIndex={handleSwiperIndexChange}
          />
        </AppBar>
        ```
        
        Somehow even tho we do the .slideToLoop with callbacks disables it still triggers the event... in our case we have a horizontal swiper, and a vertical scroller with content and dividers that we try to connect bi. directional. So this causes some loopy behaviour. Do you have an idea how to hat at this ?

nolimits4web:
Merged, thanks!

PR #5223, opened 2021-11-23
Fixed can not get clicked slider in shadow root issue

yaoguoba:
Fixed can not get clicked slider in shadow root issue

### DIFF
--- a/src/core/events/onTouchEnd.js
+++ b/src/core/events/onTouchEnd.js
@@ -36,7 +36,8 @@ export default function onTouchEnd(event) {
 
   // Tap, doubleTap, Click
   if (swiper.allowClick) {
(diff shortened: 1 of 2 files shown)


yaoguoba:
#5223

nolimits4web:
- `event.path` is not standard property
- I guess `e.path[0]` you pass to updateClickedSlide is an Element, so accessing `e.target` in `updateClickedSlide` will be anyway undefined

yaoguoba:
Sorry, I forgot to change updateClickedSlide.
Do you mean we can't use e.path here, I saw e.path used in onTouchStart.

Anyway, I will check the other way and will update later.

yaoguoba:
Hi @nolimits4web, 

Event.composedPath is a standard equivalent which is a method.
Please check.

nolimits4web:
merged, thanks!

PR #5258, opened 2021-12-02
feat(a11y): add `aria-current` to current bullet

vltansky:
https://github.com/nolimits4web/swiper/issues/3149#issuecomment-941194570

### DIFF
--- a/src/modules/a11y/a11y.js
+++ b/src/modules/a11y/a11y.js
@@ -141,6 +141,11 @@ export default function A11y({ swiper, extendParams, on }) {
             $bulletEl,
             params.paginationBulletMessage.replace(/\{\{index\}\}/, $bulletEl.index() + 1),
           );
+          if ($bulletEl.is(`.${swiper.params.pagination.bulletActiveClass}`)) {
+            $bulletEl.attr('aria-current', 'true');
+          } else {
+            $bulletEl.removeAttr('aria-current');
+          }
         }
       });
     }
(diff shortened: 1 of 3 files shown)


nolimits4web:
There is already `updatePagination` https://github.com/nolimits4web/swiper/blob/master/src/modules/a11y/a11y.js#L132 function there which already executes in `paginationUpdate` handler. So this logic should be moved there

vltansky:
fixed

src/modules/a11y/a11y.js
```diff
@@ -131,9 +131,10 @@ export default function A11y({ swiper, extendParams, on }) {
 
   function updatePagination() {
     const params = swiper.params.a11y;
-    if (hasPagination()) {
-      swiper.pagination.bullets.each((bulletEl) => {
-        const $bulletEl = $(bulletEl);
+    const doHasPagination = hasPagination();
```
nolimits4web, line 134: Maybe there is a sense to replace this line with just 
```
if (!hasPagination()) return;
```
I mean what is the point to to loop over `swiper.pagination.bullets` and do checks for pagination inside if `hasPagination()` return false?

nolimits4web, line 134: Maybe there is a sense to replace this line with just 
```
if (!hasPagination()) return;
```
I mean what is the point to to loop over `swiper.pagination.bullets` and do checks for pagination inside if `hasPagination()` return false?

vltansky: `hasPagination()` checks for `clickable`, that not necessary for aria-current
```js
function hasPagination() {
    return (
      swiper.pagination &&
      swiper.params.pagination.clickable &&
      swiper.pagination.bullets &&
      swiper.pagination.bullets.length
    );
  }
```

nolimits4web: Then it should be tweaked, otherwise if pagination type is other than bullets then more likely `swiper.pagination.bullets.each` will throw an error

vltansky: Take a look on latest commit

Issue #5003, opened 2021-09-26
"Rewind" option


putnik-projects:
### Clear and concise description of the problem

I want to suggest a new way to loop slider, not with "loop" option. This way is rewind to the first/last slide when it's at the start/end. It would be nice, if swiper had such option. Some examples on other sliders:
- https://glidejs.com/docs/options/#rewind
- https://ganlanyuan.github.io/tiny-slider/demo/#rewind_wrapper

### Suggested solution

I think this should a check in file which defines that it is time to change the slide. And if this is the last slide, go to first slide, if the first slide, go to the last slide, else go to necessary slide.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this feature?

- [ ] I'm willing to open a PR

jaro-io:
would be so nice to have this!

PR #5270, opened 2021-12-06
Added option to allow next nav element at the end of slides, which returns back to the first slide

czraptor:
Hi, I just added small feature for "custom" loop. Recently we had issues with rendering performance of swiper in react and we find out many things in our code to improve and also our UX guys had proposal of making custom loop. This feature allows next navigation element to be shown also at last slide and it allows user to click on it and then it navigates back to first slide. This is much better for performance, since we won't duplicate any slides which would be also rendered and it does help user to navigate back even without loop turned on.

Do you think it would be possible to add this feature to swiper? I also updated app.jsx in react playground, so anyone could try it.
(This is a first time I'm trying to contribute to any GitHub project, so I'm sorry if there is something I should do different in this process.)
Thanks!

### DIFF
--- a/playground/react/App.jsx
+++ b/playground/react/App.jsx
@@ -14,7 +14,7 @@ const App = () => {
         slidesPerView={3}
         spaceBetween={50}
         navigation
-        loop
+        allowNextElementAtTheEnd
         scrollbar={{ draggable: true }}
         pagination={{ clickable: true }}
       >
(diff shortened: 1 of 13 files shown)


nolimits4web:
Hi, thanks for your PR. Looks good to me but I would propose couple of small changes:
- let's rename param name to `rewind`, as this is what was requested here (https://github.com/nolimits4web/swiper/issues/5003) and how it is called in other libs 
- let's not assign it to swiper instance, and do the checks for just `swiper.params.rewind`

nolimits4web:
and same functionality should be added on `prev` click, when we are at first slide and press back button, then it should go to the last slide

nolimits4web:
and same functionality should be added on `prev` click, when we are at first slide and press back button, then it should go to the last slide

PR #5296, opened 2021-12-16
fix(angular): setElement call updateInitSwiper bug

### LINKED ISSUE #5291: Error in swiper_angular.js in version 7.3.1 or 7.3.2

- [X] I confirm

### Reproduction link

no reproduction link

### Bug description

An error occurs at line 729 (version 7.3.0) when you try to run

if (changedParams.scrollbar) {
                if (this.scrollbar &&
                    typeof this.scrollbar !== 'boolean' &&
                    this.scrollbar.el &&
                    scrollbar &&
                    !scrollbar.el) {
                    this.updateParameter('scrollbar', this.scrollbar);
                    scrollbar.init();
                    scrollbar.updateSize();
                    scrollbar.setTranslate();
                }
                else {
                    scrollbar.destroy();   // -------->  here!
                    scrollbar.el = null;
                }
            }

because the scrollbar object is undefined, but changedParams.scrollbar is true.

This does not happen until version 7.3.0 .-

In my project there are 3 swipers, but none use scrollbar, I don't even read the Scrollbar object from swiper.

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Swiper version

7.3.1 or 7.3.2

Windows 10 Pro, Chrome Versión 96.0.4664.93 (Build oficial) (64 bits)

### Validations

vltansky:
fixes #5291

### DIFF
--- a/src/angular/src/swiper.component.ts
+++ b/src/angular/src/swiper.component.ts
@@ -674,7 +674,6 @@ export class SwiperComponent implements OnInit {
       }
 
       if (changedParams.scrollbar) {
-        console.log(changedParams.scrollbar);
         if (
           this.scrollbar &&
           typeof this.scrollbar !== 'boolean' &&
--- a/src/angular/src/swiper.component.ts
+++ b/src/angular/src/swiper.component.ts
@@ -459,7 +459,8 @@ export class SwiperComponent implements OnInit {
   ) {}
 
   private _setElement(el: ElementRef, ref: any, update: string, key = 'el') {
-    if (ref && el && el.nativeElement) {
+    if (!ref && !el) return;
+    if (el.nativeElement) {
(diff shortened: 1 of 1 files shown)


src/angular/src/swiper.component.ts
```diff
@@ -459,7 +459,8 @@ export class SwiperComponent implements OnInit {
   ) {}
 
   private _setElement(el: ElementRef, ref: any, update: string, key = 'el') {
-    if (ref && el && el.nativeElement) {
+    if (!ref && !el) return;
+    if (el.nativeElement) {
```
ha-akelius, line 463: I have error in this line 
`core.js:6486 ERROR TypeError: Cannot read properties of undefined (reading 'nativeElement')`
can we convert this line to 
`if(el?.nativeElement) {`

PR #5301, opened 2021-12-18
(fix)Added "svelte": "./swiper.esm.js" to package.json

slackers123:
Related to https://github.com/sveltejs/kit/issues/3015 and a few similar Issues

The "svelte" tag missing from package.json seems to cause some errors when building svelte kit applications.

In my testing adding this line to the package.json fixes the problem completely and there are no more errors.


quentincaffeino:
What are the blockers?

Issue #5418, opened 2022-02-02
Swiper does not respect changes to autoplay delay when changed in code


jayjetley:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/s/swiper-autoplay-react-forked-qifwv

### Bug description

When changing autoplay properties such as delay on the autoplay module at runtime the swiper does not respect this. This is due to it not being included in the params watch list.

### Expected Behavior

When changing the delay at runtime the swiper should respect the new delay value.

### Swiper version

8.0.1

### Platform/Target and Browser Versions

Chrome

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

PR #5419, opened 2022-02-02
Add autoplay to watched modules

jayjetley:
Swiper does not respect changes to autoplay module at runtime

fix https://github.com/nolimits4web/swiper/issues/5418


nolimits4web:
Merged, thanks!

Issue #5361, opened 2022-01-13
"RewindOnSwipe" option


putnik-projects:
### Clear and concise description of the problem

Thank you very much for adding #5003. I like this option, but it supports only navigation button clicks or calling `.slideNext()`/`.slidePrev()` functions. And it would be very nice if this also worked when you swipe (like here, https://glidejs.com/docs/options/#rewind).

### Suggested solution

As I understand swipes has **ratio** to trigger swipe to next/previous slide. But, if this is the last slide you can't go to next slide. But with new feature if going to next slide was triggered with swipe and this is the last slide, you go to start, as in "rewind" option (as in example given above). So I want implementing this feature either as a new "rewindOnSwipe" option (or something like that) or in old "rewind" option.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this feature?

- [ ] I'm willing to open a PR

PR #5422, opened 2022-02-02
feat(core): swiping in `rewind` mode

vltansky:
closes https://github.com/nolimits4web/swiper/issues/5361

not sure if the new parameter is needed or it just should work this way always in rewind mode.

### DIFF
--- a/playground/core/index.html
+++ b/playground/core/index.html
@@ -1,72 +1,69 @@
 <!DOCTYPE html>
 <html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Swiper Playground</title>
+    <meta name="viewport" content="width=device-width" />
+  </head>
 
-<head>
-  <meta charset="UTF-8" />
-  <title>Swiper Playground</title>
-  <meta name="viewport" content="width=device-width" />
-</head>
+  <body>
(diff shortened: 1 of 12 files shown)


nolimits4web:
Look at how it works in issue's Glide example link, it does rewind during swipe not on release, so it should happen within onTouchMove handler

vltansky:
Ok, will refactor it.
But should it be a parameter?
I think no..

nolimits4web:
It should consider touch direction and swipe ratio/duration. To avoid such cases:

https://user-images.githubusercontent.com/999588/152800276-42587612-baa7-4a09-ba8d-02a15ccedab7.mov

On first slide it should rewind to last on swipe if:
- swipe direction is `prev` + touch duration is more than `params.longSwipesMs` + swipe ratio `>= params.longSwipesRatio`
- swipe direction is `prev` + touch duration is less than `params.longSwipesMs`

same for rewind on last slide

vltansky:
How about this approach ?

nolimits4web:
Nothing changed as I see:

https://user-images.githubusercontent.com/999588/153372436-de11286d-a57e-4f10-bea3-b6e3eb203312.mov

PR #5424, opened 2022-02-02
feat(angular): events types

vltansky:
exported as `SwiperNgEvents`.
Have better naming suggestion?

### DIFF
--- a/playground/angular/src/app/home/home.component.ts
+++ b/playground/angular/src/app/home/home.component.ts
@@ -1,7 +1,7 @@
 import { SwiperEvents } from 'src/types/swiper-events.d';
 import { ChangeDetectorRef, Component, NgZone, ViewChild } from '@angular/core';
 import { BehaviorSubject } from 'rxjs';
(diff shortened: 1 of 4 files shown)


nolimits4web:
Let's rename it to just `Events`

vltansky:
Difficult for autoimport?

nolimits4web:
Then `SwiperEvents` ? :)

vltansky:
then it would be the same as Cores Swiper Events:)
Maybe `SwiperEventsParams` and export it from `swiper/types` instead of angular?

nolimits4web:
> then it would be the same as Cores Swiper Events:)

That is not a problem, as you anyway import it from `swiper/angular` and it is not needed anywhere else except in Angular

nolimits4web:
Maybe then just `EventsParams`

nolimits4web:
Can you also send a PR for Angular docs?

vltansky:
done
https://github.com/nolimits4web/swiper-website/pull/234

PR #5460, opened 2022-02-11
fix(vue): HMR error

indus:
this fixes `TypeError: originalSlots[slotName] is not a function` error during HMR in vue by ensuring the function exists.

I havn't opend a issue because it is quite some work to provide a repo and the change is just an assertment.
But I'm not the only one getting this error https://github.com/ionic-team/ionic-framework/issues/24134


vltansky:
@indus see build errors

nolimits4web:
Merged, thanks!

PR #5489, opened 2022-02-23
actually use verticalClass/horizontalClass in pagination

NikitoriusMinsk:
Pagination doesn't seem to use or apply horizontalClass and verticalClass props for some reason


nolimits4web:
Thanks, I have simplified it a bit in https://github.com/nolimits4web/swiper/commit/95c882c50a798ebe7adc20fcb61765aace0edbfc

Issue #2934, opened 2019-01-02
Thumbnail Navigation Issue With Click Events.


Vinsanity:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* Swiper Version: 4.4.6.
* Platform/Target and Browser Versions: macOS, Windows, IOS, ANDROID
* Live Link: http://idangero.us/swiper/demos/300-thumbs-gallery.html

### What you did
You can see my interaction pattern with your demo here: http://p.tri.be/l7Lv4d (this was performed in MacOS 10.14.2 and Chrome 71.0.3578.98).

Here's the demo currently on your site. http://idangero.us/swiper/demos/300-thumbs-gallery.html

This issue happens with and without centerSlides as well on my local version.

### Expected Behavior
When I click on a thumbnail, it should immediately change the slide of the main slider to the index of the clicked thumbnail.

### Actual Behavior
When I click on a thumbnail, it takes 2-3 times to trigger click event. It appears that there is something impeding the propagation of click events in thumbnail controlled sliders. However, when I use the next/prev buttons on the main slider, the thumbnails are working fine.

Vinsanity:
Thanks for the response @max-ci . Please see my comment above in "What you did". On the version I have locally, I am not using `centeredSlides` and there is no carousel for the thumbnails. The thumbs are static/stationary and this same issue happens. 

see: http://p.tri.be/h8fOtj

If you're saying the dragging is the issue, I cannot seem to find an API setting that allows me to disable that all together on thumbs.

Vinsanity:
@max-ci That did the trick! Thank you so much.
I was trying `allowTouchMove` set to false and cycling through the other touch events but as soon as I set centeredSlides to true, it would "break" again. 

It was a tad confusing going through those params, but it makes sense now. 

Kudos to you my friend!

tshamz:
I'm having this exact same problem (except my stationary thumbs are vertical) and I cannot for the life of me figure out how setting `threshold` fixed your problem. Do you set `threshold` on the thumb slider or the main slider? why would setting this value to anything higher than 0 (the default) help fix this issue?

Vinsanity:
Hi @tshamz,
[...]

PR #5525, opened 2022-03-07
fix: add threshold option to thumbs examples

zcuric:
### This PR
- [x] Adds `threshold` option to thumbs examples

### Context
Due to no `threshold` being set, clicking on thumbnail feels laggy and unresponsive. The intention is to [`threshold` option](https://swiperjs.com/swiper-api#param-threshold) to make the transition smoother and experience better. 

Value is taken from [this comment](https://github.com/nolimits4web/swiper/issues/2934#issuecomment-451178450) in [this issue.](https://github.com/nolimits4web/swiper/issues/2934)

### DIFF
--- a/demos/300-thumbs-gallery.html
+++ b/demos/300-thumbs-gallery.html
@@ -55,6 +55,7 @@
       slidesPerView: 4,
       freeMode: true,
       watchSlidesProgress: true,
+      threshold: 1000
     });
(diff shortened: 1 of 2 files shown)


nolimits4web:
Sorry, don't understand why is it for, plus threshold `1000` will make swiping impossible at all

PR #5582, opened 2022-03-26
fix(core) Don't process transitions or updates on Thumbs module if swiper is destroyed

JoniVanderheijden:
There is a weird situation in React / Next when you can switch between multiple instances of the Thumb swiper when switching between pages. It hits some edge cases of the Thumb module where it calls transitions and updates on a swiper which is destroyed, which obviously causes a ton of errors.

This MR aims to be a bit more defensive about these things to avoid trying to trigger transitions or updates on destroyed swipers.

Give me a shout if you want me to explain further or if I missed a critical step.

I struggled reproducing it in a small example, because it's linked to some pretty complex navigation behaviour in Next. But I have a production environment application I can point you to that triggers it an awful lot. I can always send a link to that one in a DM.

### DIFF
--- a/src/modules/thumbs/thumbs.js
+++ b/src/modules/thumbs/thumbs.js
@@ -21,7 +21,8 @@ export default function Thumb({ swiper, extendParams, on }) {
 
   function onThumbClick() {
     const thumbsSwiper = swiper.thumbs.swiper;
-    if (!thumbsSwiper) return;
+    if (!thumbsSwiper || thumbsSwiper.destroyed) return;
+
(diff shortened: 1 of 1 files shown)


src/modules/thumbs/thumbs.js
```diff
@@ -204,18 +205,17 @@ export default function Thumb({ swiper, extendParams, on }) {
     update(true);
   });
   on('slideChange update resize observerUpdate', () => {
-    if (!swiper.thumbs.swiper) return;
```
JoniVanderheijden, line 207: Redundant, it happens in the update too

PR #5665, opened 2022-04-24
History plugin fixes and new option adding

VahanMargaryan:
- [ADDED] (keepQuery) New option to keep the query string parameters when changing user when navigating
- [FIXED] History back via browser button not working
- [ADDED] typing for History plugin options
> - enabled?: boolean
> - keepQuery?: boolean

### DIFF
--- a/src/modules/history/history.js
+++ b/src/modules/history/history.js
@@ -7,7 +7,7 @@ export default function History({ swiper, extendParams, on }) {
       root: '',
       replaceState: false,
       key: 'slides',
-      keepQuery: true,
+      keepQuery: false,
     },
   });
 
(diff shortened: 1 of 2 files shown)


src/modules/history/history.js
```diff
@@ -58,6 +59,9 @@ export default function History({ swiper, extendParams, on }) {
     } else if (!location.pathname.includes(key)) {
       value = `${key}/${value}`;
     }
+    if (swiper.params.h.keepQuery) {
```
nolimits4web, line 62: should be `swiper.params.history.keepQuery` I guess

nolimits4web (changes requested):

PR #5751, opened 2022-05-30
chore: add react as a peerDependency

kevinji:
This makes the dependency on react explicit, which makes Yarn PnP happy. It's listed as an optional peerDependency so it's not automatically installed. Feel free to update the version ranges as needed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -101,6 +101,14 @@
       },
       "engines": {
         "node": ">= 4.7.0"
+      },
+      "peerDependencies": {
+        "react": "^17 || ^18"
+      },
+      "peerDependenciesMeta": {
+        "react": {
+          "optional": true
+        }
       }
     },
     "node_modules/@ampproject/remapping": {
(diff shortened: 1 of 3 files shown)


nolimits4web:
Hey @kevinji ! Thanks, but to be honest I am not sure it is needed, and I am sure it can produce/lead to errors for some other users who doesn't use React version, we have something similar with `svelte` in peer deps and users of other frameworks complained about it

kevinji:
Does the optional flag not solve the problem of a single package? Specifically, this only produces an error if a user explicitly uses React and it's not a supported version, but it's fine if they don't use React.

nolimits4web:
> if a user explicitly uses React and it's not a supported version

in this case user won't be able to use Swiper anyway i guess.

Plus as i can see `peerDependenciesMeta` was introduces in npm v7, so anyone using npm v6 will get warnings about React must be installed.

So i don't thing it is a way to go

PR #5755, opened 2022-05-31
fix: slide card click event would occasionally fail due to triggering…

JSMengLingqi:
I have found the PC click event occasionally failed. The reason is that the click event did not take effect because pointerDown/PointerMove  events(Touch Events) was triggered and the event was not blocked at the boundary condition, the current version only intercepts the TouchStart event on mobile. But Pointerdown also belongs to Touch Events


nolimits4web:
Pointerdown is not a touch only event, it is also produced by mouse events too, so this fix makes `isTouchEvent` obsolete

JSMengLingqi:
But it does solve the problem of click events occasionally failing

Issue #5756, opened 2022-05-31
swiper@8.1.6 autoplay "Maximum call stack size exceeded"


DarknessChaser:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://stackblitz.com/edit/vitejs-vite-6ep5ry?file=package.json,src%2FApp.vue&terminal=dev

### Bug description

"Maximum call stack size exceeded" error, when I use autoplay.
swiper@8.1.5 it works fine

![image](https://user-images.githubusercontent.com/22089140/171190735-5fec740a-1455-452d-9b30-80180cec1b3c.png)

### Swiper version

8.1.6 and 8.2.1 Maybe 8.16~8.2.1

### Platform/Target and Browser Versions

window10 chrome 101.0.4951.67

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

nolimits4web:
it is fixed by this commit https://github.com/nolimits4web/swiper/commit/56e050bcdbcc00db7af02ccb32e0725265dd242a

PR #5757, opened 2022-05-31
fix(autoplay): fix issues with https://github.com/nolimits4web/swiper…

DarknessChaser:
fix https://github.com/nolimits4web/swiper/issues/5756


nolimits4web:
Thanks, it is fixed by this commit https://github.com/nolimits4web/swiper/commit/56e050bcdbcc00db7af02ccb32e0725265dd242a

Issue #5668, opened 2022-04-25
ref not exposed to us as prop


ivanjeremic:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

nope

### Bug description

```js
<Swiper ref={someRef}>
 ...
</Swiper>
```
We can't do this but we should have the option.

### Swiper version

all

### Platform/Target and Browser Versions

all

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

nolimits4web:
what do you expect `ref.current` to return?

ivanjeremic:
It works now was an issue with an other hook I guess.

7iomka:
> ivanjeremic

How you found a workaround for this?

```tsx
 const swiperRef = useRef(null);

// ...
<Swiper ref={swiperRef}
   ... >
</Swiper>

```
gives me
```
Property 'ref' does not exist on type 'IntrinsicAttributes & SwiperProps & { children?: ReactNode; }'
```

Why? Swiper is exported as component with `forwardRef` and types needed to contain `ref`, for example with helper type `React.ForwardRefExoticComponent<SwiperProps & {...}>`

Another issue - I want to have a type of this ref. Currently swiper types not contain type of SwiperClass, but that is what we need.
What do you think? @nolimits4web 

This flow is required for control swiper instance outside.

Thanks.

ivanjeremic:
> > ivanjeremic
> 
> How you found a workaround for this?
> 
> ```tsx
>  const swiperRef = useRef(null);
> 
> // ...
> <Swiper ref={swiperRef}
>    ... >
> </Swiper>
> ```
> 
> gives me
> 
> ```
> Property 'ref' does not exist on type 'IntrinsicAttributes & SwiperProps & { children?: ReactNode; }'
> ```
> 
> Why? Swiper is exported as component with `forwardRef` and types needed to contain `ref`, for example with helper type `React.ForwardRefExoticComponent<SwiperProps & {...}>`
> 
[...]

Issue #5500, opened 2022-02-27
[React] Typescript errors in `ref` and `onSwiper` props


vicasas:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/s/swiper-typescript-types-errors-437w68

### Bug description

Using Swiper with Typescript it throws errors in the `ref` and `onSwiper` props.

### Expected Behavior

It should not throw an error even if types are assigned to it.

### Swiper version

8.0.6

### Platform/Target and Browser Versions

macOs bigsur Chrome lastest

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

rtabulov:
hey, this is not a problem with Swiper or Typescript, i would say this is not a problem at all

[`ref` is not supported for function components in react](https://reactjs.org/docs/refs-and-the-dom.html#accessing-refs)
> You may not use the ref attribute on function components because they don’t have instances.

`onSwiper` shows error becuase types null and Swiper are not compatible, that's because you did not properly set the type on `useState`

to do that you should write

```ts
import type { Swiper } from 'swiper';
const [swiperInstance, setSwiperInstance] = useState<Swiper | null>(null);
```

youngkyo0504:
Do you want access swiper container?? 
then do this.

```typescript
function App() {
    const ref = useRef<SwiperCore>();
   
   <Swiper
      onInit={(swiper) => {
        ref.current = swiper.el; 
        // "swiper.el" is what you want
      }}
   ></Swiper>
}
```

vicasas:
> hey, this is not a problem with Swiper or Typescript, i would say this is not a problem at all

@rtabulov The `ref` makes sense. But this means that Swiper should still be able to receive the ref without Typescript throwing an error. Perhaps using `fordwarRef`.

> to do that you should write

Thinking out loud, you should consider adding last name `Props` to types. e.g. `SwiperProps`

Tends to be confusing without using `import type`

vicasas:
@youngkyo0504 Taking that same code to my example in codesandbox how would it be? Using `fordwarRef`?

[...]

PR #5807, opened 2022-06-18
Fix Typescript errors with ref

7iomka:
Since swiper for react uses forward ref, we need to fix exported types. Also, we need type as alias for `SwiperCore`

With this fix we can write code like this:

```tsx
import type { SwiperInstance } from 'swiper/react/swiper-react';
import { Swiper, SwiperSlide } from 'swiper/react/swiper-react';

// ...
const swiperRef = useRef<SwiperInstance>(null);

// ...
<button
  className="button-outside-of-swiper-area"
  type="button"
  onClick={() => swiperRef.current?.slideNext()}
/>
<Swiper
  ref={swiperRef}
></Swiper>
// ...

```

Related issues: 
- #5500
- #5668

### DIFF
--- a/src/react/swiper-react.d.ts
+++ b/src/react/swiper-react.d.ts
@@ -78,8 +78,9 @@ interface SwiperProps
     | 'onScroll'
   > {}
 interface SwiperSlideProps extends React.HTMLAttributes<HTMLElement> {}
+interface SwiperInstance = typeof SwiperClass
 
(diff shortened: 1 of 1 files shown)


nolimits4web:
`Swiper` component ref is the HTML element, and not Swiper instance

ivanjeremic:
> `Swiper` component ref is the HTML element, and not Swiper instance

but has `Swiper` the ability to take `ref`?

PR #5863, opened 2022-07-07
Rename swiper-solid.js and swiper-react.js to index.js

TiagoCavalcante:
This way the following code is valid:
```js
// With SolidJS
import { Swiper } from "swiper/solid"
// With React
import { Swiper } from "swiper/react"
```


nolimits4web:
Merged, thanks!

PR #5860, opened 2022-07-07
Stop using cloneNode in sliders without loops in SolidJS

TiagoCavalcante:
This allows events to be assigned to the slides as usually

Before it didn't work:
```html
<Swiper>
  <SwiperSlide>
    <button onClick={() => window.alert("clicked")}>
      Click me!
    </button>
  </SwiperSlide>
</Swiper>
```

Now it works

PR #5867, opened 2022-07-09
SolidJS: allow event handlers in slides of a looping Swiper

TiagoCavalcante:
Same as #5860 but in sliders with loops.
Sadly, before this (and before #5860), clicking buttons in looping slides was already broken.


nolimits4web:
Not really, with this fix loop mode just doesn't work at all, as it is just shuffle slides rather than creating cloned ones which are necessary for loop mode to work properly

TiagoCavalcante:
Ok, I think that buttons won't work in looping slides so

PR #5873, opened 2022-07-11
Removed editor specific folders

IgorShayderov:


vltansky:
vscode settings force prettier formatting on save. Otherwise, it will fail on the Github action that runs `npm run check-format`.
auto-formatting on save pretty useful feature as for me.

Issue #5791, opened 2022-06-11
Swiper´s version 8.2.3 broken on Angular 13


JMJHOX:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/s/heuristic-yalow-0k5bi9?file=/package.json

### Bug description

Using swiper´s version  8.2.3 on Angular 13 crashes a clean project displaying "Error: Module build failed" . 

> It says depends upon a library published using Angular version 14.0.0, which requires Angular version 14.0.0 or newer to work correctly. Consider upgrading your application to use a more recent version of Angular.
I tried to replicate the error on codesandbox, but it seems to not be replicating it correctly.

### Expected Behavior

On a fresh-clean install of Angular project, installing swiper´s library and importing  SwiperModule on app.module.ts, it should be working correctly as expected.

### Actual Behavior

When I create a clean project on Angular 13 and install only the swiper´s library through npm  with version  "^8.2.3" and try to import it on the app.module.ts, it crashes my project and displays an error and angular fails to compile it.

Here's a pastebin with the error logs:
[Pastebin of the errors log on terminal](https://pastebin.com/nptp51sp)

There´s not issue when you make a rollback with the previous version like 8.2.2

### Swiper version

8.2.3

### Platform/Target and Browser Versions

Web

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

Andre-O:
I have the same issue with 8.1.5. Tried to downgrade and to upgrade to other versions - no success. Even those of our app versions which worked well before can no longer be built! 
[This issue](https://github.com/nolimits4web/swiper/issues/5792) describes the same problem for version 8.2.2.

jjamaral:
Same here. Any help on this please?

Andre-O:
> Same here. Any help on this please?

Did you try to install version 8.2.4? It works for me now.

vltansky:
@jjamaral 
https://codesandbox.io/s/swiper-navigation-angular-forked-6ion4f?file=/package.json

jjamaral:
[...]

PR #5777, opened 2022-06-06
feat(angular): update to angular v14

yharaskrik:
Update to angular v14

PR #5926, opened 2022-08-01
chore(deps): upgrade angular to v14

vltansky:


nolimits4web:
We had this PR before https://github.com/nolimits4web/swiper/pull/5777 but it produced breaking changes

nolimits4web:
Will it still work with Angular 13? https://github.com/nolimits4web/swiper/issues/5791

PR #5927, opened 2022-08-01
feat(build): upgrade deps & cjs to esm

vltansky:


nolimits4web:
Thanks! Seems like everything is fine except Cypress, seems like it doesn't work/like ES modules? :) What should we do with it?

vltansky:
@nolimits4web all green now :) 🎉

Issue #4413, opened 2021-04-07
Children does not render properly (after wrapper component)


tobiaszciesielski:
Hello, I have already posted on stackoverflow about rendering children in swiperjs using React.

https://stackoverflow.com/questions/66993033/swiper-js-for-react-does-not-render-children-properly
I use: 
```
"gatsby": "^3.0.1",
"react": "^17.0.1",
"swiper": "^6.5.0",
```

## What You Did

I want to render children like this: 
`<Swiper>{children}<Swiper/>`
But children are rendered after swiper wrapper.

## Expected Behavior

I want to pass all children and render them inside Swiper wrapper because I fetch data from external source (headless cms)

## Actual Behavior

Swiper Slides are rendered after Swiper component what makes cards invisible in page content.

The code you can see in stack overflow and this commit of my project
https://github.com/tobiaszciesielski/tciesielski.pl/pull/1/commits/5500865d7eef1a99facb03e3dedd987ee1ce6033

alicerocheman:
Same issue.

## react
> Slider.tsx
```tsx
import { FC, useMemo } from 'react';
import { Swiper } from 'swiper/react';
import { Navigation, Scrollbar, A11y, Keyboard } from 'swiper';

const breakpoints = {
  1100: {
    slidesPerView: 4,
  },
  768: {
    slidesPerView: 3,
  },
  464: {
    slidesPerView: 2,
  },
};

const Slider: FC = ({ children }) => {
  const buttonClassName = useMemo(
    () =>
      `group rounded-full w-9 h-9 flex flex-col justify-center items-center border border-gray-50 bg-white text-xl font-bold ml-2 hover:bg-primary hover:border-primary`,
    []
  );
  const prevClassName = useMemo(() => `prev ${buttonClassName}`, [buttonClassName]);
  const nextClassName = useMemo(() => `next ${buttonClassName}`, [buttonClassName]);
  return (
    <>
      <Swiper
        className="w-full"
        tag="ul"
        modules={[Navigation, Scrollbar, A11y, Keyboard]}
        navigation={{ prevEl: '.prev', nextEl: '.next' }}
        slidesPerView={1}
        spaceBetween={16}
        breakpoints={breakpoints}
      >
        {children}
      </Swiper>
      <div className="hidden absolute right-0 top-0 flex-row sm:flex w-24 h-8">
        <button className={prevClassName}>
[...]

PR #5954, opened 2022-08-11
Allow React SwiperSlide children as long as displayName includes SwiperSlide

crtl:
Fixes [#4413](https://github.com/nolimits4web/swiper/issues/4413)

Replaces strict `displayName` check with `includes` check to allow different `displayNames` for components that act as `SwiperSlide` as long as their `displayName` includes the phrase `SwiperSlide`.


nolimits4web:
merged, thanks!

PR #5956, opened 2022-08-12
Types: add missing AutoplayMethods members

TiGR:
Autoplay typing misses some members that might be handy to have (such as `paused` and `pause()`). This change fixes it.


nolimits4web:
Merged, thanks!

PR #6120, opened 2022-10-05
Update outdated LICENSE year

Serjobas:
Updated the outdated copyright year to the present. 
 So there will not be a need for any further license year updates


nolimits4web:
This seems to be is not valid https://opensource.stackexchange.com/a/4890

Issue #5962, opened 2022-08-16
Accessibility issue: Focusing on input forces slider to jump


wisniowski:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/embed/swiper-slides-per-view-forked-n8pht9?fontsize=14&hidenavigation=1&theme=dark

### Bug description

Hello,

everything works just great in 99% of scenarios, although I found an odd bug that causes the slider to roughly jump once the input field gets focused - it doesn't even swipe, it simply instantly jumps to another section.

### Expected Behavior

When input/textarea gets focused, swiper should stay.

### Actual Behavior

When input/textarea gets focused, swiper jumps oddly (it doesn't even swipe).

### Swiper version

8.3.2

### Platform/Target and Browser Versions

All

### Validations

- [x] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [x] I'm willing to open a PR

conniesmits:
I have also encountered this problem so hopefully this will be fixed soon, because now we are forced to use an old version of swiper in our project.

ntbrown:
@conniesmits @wisniowski @nolimits4web Likely relevant: https://allyjs.io/tutorials/focusing-in-animated-ui.html

mo3az96:
to workaround disable accessibility (if you don't need accessibility feature) by adding
a11y: {
    enabled: false,
},
to your swiper options

Blacklightdesign:
I have a solution that mobile touch on ui elements like links or select fields no longer makes the slide jump to first place. Touch events also trigger a focus event. I added a condition to check that the focus event was not triggered by touch.
#6139

davidmarquis:
The more appropriate workaround if you don't want to completely disable all accessibility options is to set the `scrollOnFocus` option specifically:

```
a11y: {
scrollOnFocus: false,
},
```

Issue #5878, opened 2022-07-13
focusableElements with allowTouchMove doesn't work


yasurev86:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/s/brave-kilby-rfulow

### Bug description

focusableElements: false with allowTouchMove: false
do not work together

### Expected Behavior

when you click on the button in the slide, the prescribed handler is triggered

### Actual Behavior

the slide is changed, after which the button becomes available for clicking

### Swiper version

8.3.0

### Platform/Target and Browser Versions

Windows Chrome, Firefox

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [x] I'm willing to open a PR

Blacklightdesign:
I have a solution that mobile touch on ui elements like links or select fields no longer makes the slide jump to first place. Touch events also trigger a focus event. I added a condition to check that the focus event was not triggered by touch.
#6139

Sefibrah:
This is how I solved this issue in my Angular project, hopefully, it's replicable in other scenarios. Instead of providing the swiper container ```allowTouchMove``` directly in the template as such:

```
// this DOESN'T WORK (for me at least)
<swiper-container [allowTouchMove]="false"> 
  // stuff
</swiper-container>
```

Instead, I did the following. In my home.component.ts I added an options variable:
```
@Component({ .. })
export class HomeComponent {
  mySwiperOptions: SwiperOptions = {
      allowTouchMove: false,
      // other swiper options
  }
}
```

And provided this variable to the swiper container in my home.component.html like so:
```
<swiper-container [options]="mySwiperOptions"> 
  // stuff
</swiper-container>
```

Hopefully, this helps someone 😁

Issue #5814, opened 2022-06-22
Multiple slides per view - Clicking links on mobile (or right click/scroll wheel click on desktop) jumps the clicked element to first position instead of opening the link


IonutG22:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codepen.io/Ionut2022/pen/rNJEOgx

### Bug description

In sliders with mutliple elements visible that contain a link, clicking on a slide on a mobile device (simulating touch with browser DevTools also works - refresh page after resizing the window), or right clicking/scroll wheel clicking an slide, jumps the clicked slide to the first position instead of opening the link.

The link opens correctly only when the slider is at the last slide, or clicking the first slide of the current view.

Edit: The problem appeared since version 8.0.0
Version 7.4.1 is the last version that works correctly: https://codepen.io/Ionut2022/pen/ZErdJEb

### Expected Behavior

Clicking a slide that contains a link on a mobile device opens the link.
On desktop, scroll wheel clicking on a slide with a link, opens the link in new tab.

### Actual Behavior

Clicking a slide with a link on a mobile device (other then the first slide in the current view), jumps the clicked slide to the first position and does not open the link.
Scroll wheel clicking a slide with a link, does not open the link and jumps the clicked slide to the first position.

### Swiper version

8.2.4

### Platform/Target and Browser Versions

Android Chrome / Windows Chrome

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [x] I'm willing to open a PR

romain25:
Hi everyone, I have the same issue.

voran:
+1

MasterProvider:
add target="_blank" in the inside of your a tag to open a new tab

dijkermans:
"or right clicking/scroll wheel clicking an slide, jumps the clicked slide to the first position instead of opening the link."

seems to be related with: https://github.com/nolimits4web/swiper/issues/5625

laurenskling:
Thanks @nolimits4web , just tested in and fixes the bug! Much appreciated

Superhands89:
[...]

Issue #5625, opened 2022-04-11
Unexpected behavior when right clicking on a link


Grumvol:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://swiperjs.com/demos#slides-per-view

### Bug description

Right-clicking on a link in a slide will move to the next slide. To reproduce the error, the slide must not be the first in a row.

![QIP Shot - Screen 1287](https://user-images.githubusercontent.com/20611667/162740983-024636c5-2c88-4da8-b237-054b1619241f.png)

### Expected Behavior

Context menu open

### Actual Behavior

Move to the next slide and context menu open

### Swiper version

8.1.0

### Platform/Target and Browser Versions

Windows 10, Chrome

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

robizzt:
Can confirm, same happening to me. If slide is not on the first position, right click will move that slide to the first position.

ThatWebCo:
Same happening here. Only seems to happen when slidesPerGroup is a different number to slidesPerView.

Grumvol:
Same behavior on smartphones after touch. That is, it is impossible to follow the link if it is inside the slide. This is a very annoying bug, only for this reason I can't use version 8.

mohamadsinane:
Hello, i got the same issue.
It seams that adding `watchSlidesProgress: true,` to my default config fixes the issue.

Grumvol:
> watchSlidesProgress: true

Thanks

dijkermans:
> Hello, i got the same issue. It seams that adding `watchSlidesProgress: true,` to my default config fixes the issue.

Yes but I noticed it's CPU intensive. I hope there's a better solution.

Superhands89:
This is now fixed for desktop but this is still an issue on mobile/touch devices in version 8.4.2

Blacklightdesign:
I have a solution that mobile touch on ui elements like links or select fields no longer makes the slide jump to first place. Touch events also trigger a focus event. I added a condition to check that the focus event was not triggered by touch.
#6139

PR #6139, opened 2022-10-13
fix: mobile touch on focusable elements jumps element to first position

Blacklightdesign:
Fix:
- check if focus event triggered by touch
- don't jump the slide element

The problem has been discussed several times and a solution has already been implemented for desktop click events.

On mobile devices with touch gestures, however, the error is still present. If a UI element like select field or button is clicked in a slide element, the slide jumps to the first position.

Exactly as it is desired for the focus events. (if you navigate with tabulator key through the slider).

--> The problem was, with a touch event, the focus event is always triggered. For this I have now implemented a query.

Similar problems:
* #5962 
* #5878 
* #5814 
* #5513 
* #5625 

Always follow the [contribution guidelines](https://github.com/nolimits4web/swiper/blob/master/CONTRIBUTING.md) when submitting a pull request.

### DIFF
--- a/src/modules/a11y/a11y.js
+++ b/src/modules/a11y/a11y.js
@@ -186,6 +186,7 @@ export default function A11y({ swiper, extendParams, on }) {
       swiper.visibleSlides &&
       swiper.visibleSlides.includes(slideEl);
     if (isActive || isVisible) return;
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thanks!

Issue #5241, opened 2021-11-29
Problem on ref in typescript


farhadf246:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/s/swiper-default-react-forked-scddn?file=/src/App.tsx

### Bug description

i want to use ref in typescript but i got this error on IDE and build time

**this is my code:** 
`const swiperRef = useRef() as React.RefObject<any>
<Swiper ref={swiperRef} className="mySwiper">`

**this is error message:**
`Type '{ children: Element[]; ref: RefObject<any>; className: string; }' is not assignable to type 'IntrinsicAttributes & SwiperProps & { children?: ReactNode; }'.
  Property 'ref' does not exist on type 'IntrinsicAttributes & SwiperProps & { children?: ReactNode; }'`

i'm not sure i'm doing something wrong or this is a bug.

### Swiper version

7.3.1

### Platform/Target and Browser Versions

chrome

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

mattvb91:
Just came across the same issue and my current hack is to do the following:

```ts
  const swiper = useRef() as any;
```

```ts
<Swiper
          onInit={(core: SwiperCore) => {
            swiper.current = core.el
          }}
/>
```

farhadf246:
> Just came across the same issue and my current hack is to do the following:
> 
> ```ts
>   const swiper = useRef() as any;
> ```
> 
> ```ts
> <Swiper
>           onInit={(core: SwiperCore) => {
>             swiper.current = core.el
>           }}
> />
> ```

Thanks
I'm doing this right now but I think this is not the right way

mattvb91:
Yup its for sure just a workaround hack to get the ref. I haven't been able to find a solution for working with the ref in another way

wibed:
you could create your own union type using:
```js
const RefSwiper: React.FunctionComponent<
  SwiperProps & RefAttributes<SwiperCore>
> = Swiper;
```

wibed:
@farhadf246 
@nolimits4web 

[...]

PR #6186, opened 2022-11-03
Fix: Add type support for Swiper/React ref object

dasilvaluis:
**Problem:**

When using Swiper/React with Typescript, and when one attempts to assign a ref to the Swiper component, TS gives an error similar to this:

```
Type '{ children: Element[]; ref: RefObject<any>; className: string; }' is not assignable to type 'IntrinsicAttributes & SwiperProps & { children?: ReactNode; }'. Property 'ref' does not exist on type 'IntrinsicAttributes & SwiperProps & { children?: ReactNode; }'
```

This was also discussed in this issue: https://github.com/nolimits4web/swiper/issues/5241

**Solution:**

This happens because, although the Swiper component uses forwardRef, it is not defined in the .d.ts file.

The proposal is to create an interface that represents the element injected in the ref, such as
```ts
interface SwiperRef extends React.HTMLAttributes<HTMLElement> {
  swiper: SwiperClass;
}
```

And use it in the Swiper `declare` with `React.RefAttributes`, this way:

```ts
declare const Swiper: React.FunctionComponent<React.RefAttributes<SwiperRef> & SwiperProps>;
```
**Usage:**

With this solution in place, devs could then use the `SwiperRef` in such a way:

```ts
import React, { useRef } from 'react';
import { Swiper, SwiperRef, SwiperSlide } from 'swiper/react';

() => {
 const swiperRef = useRef<SwiperRef>(null);
 
 const swiperApi = swiperRef.current?.swiper;

  return (
    <Swiper ref={swiperRef}>
      <SwiperSlide>Hello1</SwiperSlide>
      <SwiperSlide>Hello2</SwiperSlide>
    </Swiper>
  );
};
```


nolimits4web:
This is wrong as Swiper ref is a HTML Element, not an instance of Swiper

nolimits4web:
Yes, missed that. Merged, thanks!

ibrvimv:
Hi, I cannot find SwiperRef in swiper-react.d.ts. Should I install it? thanks

omaxlive:
> 

Probably because it's still not released @ibrvimv 

Thank you @dasilvaluis for this PR. Do you know when is it expected to be released?

dasilvaluis:
> > 
> 
> Probably because it's still not released @ibrvimv
> 
> Thank you @dasilvaluis for this PR. Do you know when is it expected to be released?

I don't know man, I'm just some other guy

waleedcj:
swiperRef doesn't exist

ilik666:
SwiperRef doesn't exist yet

yusufcode:
Is there a solution on this case? I tried almost all the solutions but non of them worked. I'm working on create-react-app --template typescript. When I was working without --template typescript there was no problem like this.

dasilvaluis:
@yusufcode hard to help without seeing your code.

yusufcode:
@dasilvaluis hey man, thank you for return. Actually, there is no difference between typescripted and not typescripted codes. These are my codes:
```
<Swiper
        spaceBetween={24} // <- problem accours here
      >
        {
          swiperItems.map((swiperItem,i)=>
            <SwiperSlide key={i}>
              <Box display="flex" flexDir="column" bg="#fff" border="1px solid #E0D8D7" borderBottom="4px solid #ED1C24" borderRadius="2px">
                <Text>{swiperItem.title}</Text>
                <Text>{swiperItem.description}</Text>
                <Box>
                  <Button variant="second">Learn More</Button>
                  <Button variant="first">Find A Dealer</Button>
                </Box>
              </Box>
            </SwiperSlide>
          )
        }
      </Swiper>
```

By the way I SOLVED the problem. I downgraded the swiper from v10.0.4 to v9.4.1 and type problems gone. higher versions from v9.4.1  has these type problems. So I advise you guys to use lower versions of swiper.

PR #6259, opened 2022-12-06
bump: angular peerDependencies to support 13 to 15

steve3d:
update the angular peerDependencies, so swiper can work with angular 15.

Also, because the angular component use template, and current angular in packages.json version is 13.x, so the built package can not work with any version below 13, but it should work with angular 15.

Again thanks for this great package.


juanmarin96:
When will you release this?

PR #6346, opened 2023-02-03
fix(element): css importing paths 

michael-hein:
When trying to work with the swiper as a webcomponent and the needs to call the register function with false paramenter (https://swiperjs.com/element#injecting-styles).
And then trying to import the swiper elements styles 'swiper/element/css' or some other swiper element css styles (not modules) it fails with the reason that the styles are not found.

So correcting some exports in copy package.json should fix this.


nolimits4web:
Oh, that was a silly typo. Merged, thank you!

Issue #6261, opened 2022-12-07
TypeError : Cannot read properties of undefined (reading 'autoplay')


hackftz:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/s/green-thunder-19u5fv?file=/src/App.js

### Bug description

this is a typescript error

<img width="1984" alt="image" src="https://user-images.githubusercontent.com/33424860/206129571-ad1eedc9-ba56-41d6-818d-e245402bd362.png">

### Swiper version

8.4.5

### Platform/Target and Browser Versions

Version 106.0.5249.119 (Official Build) (arm64)

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

IlanKro:
I am getting the same issue, tried to track it, it seems the swiper object is "destroyed" when it arrives at the "clearTimeout" function
The demo doesn't seem to reproduce the issue, I don't see the error there.
it falls on a null check since a destroyed swiper has no "paras". or "auto play"
maybe adding a null check before the error will help?
 if(!swiper.params) return;

jeroenhoebe:
I'm having the same issue. With angular though.

oleghalin:
Getting same error here, when i destroy swiper in Vue destroy hook after timeout pass it throws same error, swiper instance already destroyed, but it tries to execute swiper params

Issue #5244, opened 2021-11-30
TypeError : Cannot read properties of undefined (reading 'autoplay')


muneebrana414:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

its personal

### Bug description

![Screenshot 2021-11-30 200022](https://user-images.githubusercontent.com/36512513/144071490-b0db3d19-5436-4559-a08c-dbfb7545212d.jpg)

### Swiper version

7.3.1

### Platform/Target and Browser Versions

chrome

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

muneebrana414:
Here I, apply swiper


import { Swiper, SwiperSlide } from "swiper/react";
import "swiper/css";
import SwiperCore, { Autoplay } from 'swiper'
SwiperCore.use([ Autoplay ]);

export default function SideSwipe(){
    return(
        <div>
            <Swiper slidesPerView={5} centeredSlides={false} slidesPerGroup={1} spaceBetween={1} loop={true} direction={'vertical'} 
            // autoplay={true} autoplay={{
            //     'delay':2000,
            //     'disableOnInteraction': false
            // }} 
            className="mySwiper mt-2" style={{height:700}}>
                <SwiperSlide><img src='./img/logo' /></SwiperSlide>
                <SwiperSlide><img src='./img/logo' /></SwiperSlide>
                <SwiperSlide><img src='./img/logo' /></SwiperSlide>
                <SwiperSlide><img src='./img/logo' /></SwiperSlide>
                <SwiperSlide><img src='./img/logo' /></SwiperSlide>
                <SwiperSlide><img src='./img/logo' /></SwiperSlide>
                <SwiperSlide><img src='./img/logo' /></SwiperSlide>
                <SwiperSlide><img src='./img/logo' /></SwiperSlide>
                <SwiperSlide><img src='./img/logo' /></SwiperSlide>
            </Swiper>
        </div>
    );
}

muneebrana414:
@nolimits4web i post my code

hackftz:
[...]

PR #6366, opened 2023-02-08
Fixing autoplay undefined error

IlanKro:
Added a null check to "proceed" function so it doesn't run when the swiper doesn't exist used the logic in "onTransitionEnd" for consistancy purposes, tested the fix on my project and it works correctly now...
The problem is described in this issue: https://github.com/nolimits4web/swiper/issues/6261
and this: https://github.com/nolimits4web/swiper/issues/5244
It probably has some specific trigger conditions but it doesn't matter on this fix because it's adding null checks which are never a bad thing.


nolimits4web:
Merged, thank you!

PR #6431, opened 2023-02-24
Fix a crash with resize when the autoplay has gone away during timeout

aub:
We had a crash this morning that you can see below:

Screenshot 2023-02-24 at 8 49 24 AM

I think what's happening is that:

* A resize occurs
* This schedules a timeout to resume the autoplay, checking that it should do so before scheduling the timeout
* Before the timeout's callback fires, the autoplay goes away, presumably because the user has moved on to a different page
* The timeout fires and does not check again for the existence of the autoplay, which causes the crash

This just duplicates the check for needing to resume the autoplay within the callback. I debated about whether all three checks were needed but went for the maximal approach just to be sure... happy to tweak that if needed.

### DIFF
--- a/src/core/events/onResize.js
+++ b/src/core/events/onResize.js
@@ -44,7 +44,9 @@ export default function onResize() {
   if (swiper.autoplay && swiper.autoplay.running && swiper.autoplay.paused) {
     clearTimeout(timeout);
     timeout = setTimeout(() => {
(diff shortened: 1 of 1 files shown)


nolimits4web:
Thanks!

dovudbek7:
👍

Issue #6473, opened 2023-03-08
swiper element does not reset initialized flag on destroy


liamdebeasi:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/p/sandbox/swiper-default-forked-p9ll7i

Steps to reproduce:

1. Open code reproduction.
2. Observe that you can swipe the Swiper container.
3. Click "Move". This will move the Swiper element inside of `.test-container`.
4. Observe that you cannot swipe the Swiper container after the move.

### Bug description

Swiper Element has an `initialized` flag that it sets in the `initialize` function: https://github.com/nolimits4web/swiper/blob/0c8ae8bda504f94c0fd00d9d34920f7ff0d53adc/src/element/swiper-element.js#L110

If `this.initialized` is `true` then the `initialize` function returns early. When `disconnectedCallback` is fired, Swiper Element destroys the Swiper class instance: https://github.com/nolimits4web/swiper/blob/0c8ae8bda504f94c0fd00d9d34920f7ff0d53adc/src/element/swiper-element.js#L141-L145

However, Swiper Element never resets its own `initialized` flag. When `connectedCallback` is fired and `initialize` runs, the function returns early because `this.initialized` is still `true`. As a result, the slides component does not respond to user interaction when the component is re-added to the DOM. This is because a new Swiper class instance is not created.

This impacts developers using Ionic Angular because page components will be moved around in the DOM when using `ion-router-outlet`, causing `disconnectedCallback` to fire on Swiper Element (see https://github.com/nolimits4web/swiper/discussions/6384)

### Expected Behavior

I expect Swiper Element to re-initialize whenever it is re-added to the DOM.

### Actual Behavior

Swiper Element does not re-initialize whenever it is re-added to the DOM.

### Swiper version

9.1.0

### Platform/Target and Browser Versions

macOS Chrome 110

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

PR #6474, opened 2023-03-08
fix(swiper): reset initialized flag on disconnectedCallback

liamdebeasi:
resolves https://github.com/nolimits4web/swiper/issues/6473

This PR updates the `disconnectedCallback` function to set `this.initialized = false`. That way the `initialize` function will not return early when it is called from `connectedCallback`.

I could not find where the E2E are in `master`. I see there are Cypress E2E tests in the `v8` branch: https://github.com/nolimits4web/swiper/tree/v8/cypress/e2e/modules But it looks like they were moved/removed for v9. Happy to write a test if someone could point me in the direction of the tests.

### DIFF
--- a/src/element/swiper-element.js
+++ b/src/element/swiper-element.js
@@ -142,6 +142,7 @@ class SwiperContainer extends ClassToExtend {
     if (this.swiper && this.swiper.destroy) {
       this.swiper.destroy();
     }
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thank you!

PR #6475, opened 2023-03-10
fix #6460 for margin-right

### LINKED ISSUE #6460: Pagination bug in loop mode

- [X] I confirm

### Reproduction link

https://codesandbox.io/embed/festive-browser-f1ebos?fontsize=14&hidenavigation=1&theme=dark

### Bug description

Hey! Here is a bug with pagination in loop mode, when slidesPerView: "auto".
If you click the last bullet and then the third one, which index is 2, swiper don't make loopFix correctly and slide to index 1.

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Swiper version

9.1.0

Chrome 110.0.5481.178

### Validations

drdah123:
when swiperSlide has a margin-right with spaceBetween happen problem so I make sure the margin-right of swiperSlide  = 0

### DIFF
--- a/src/core/update/updateSlides.js
+++ b/src/core/update/updateSlides.js
@@ -67,12 +67,12 @@ export default function updateSlides() {
   // reset margins
   slides.forEach((slideEl) => {
     if (rtl) {
-      slideEl.style.marginLeft = '';
+      slideEl.style.marginLeft = '0';
     } else {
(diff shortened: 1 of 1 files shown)


drdah123:
added commit to fix #6460 pagination last slides with loop

PR #6476, opened 2023-03-10
Add pagination bullet border-radius CSS variable

yuliyasitnikova:
Added the feature to manage the border radius value for pagination bullet.

You can change the value in the same way as the bullet size or color:

```scss
.my-swiper {
  --swiper-pagination-bullet-width: 48px;
  --swiper-pagination-bullet-height: 4px;
  --swiper-pagination-bullet-border-radius: 0;
}
```


nolimits4web:
Merged, thank you!

PR #6495, opened 2023-03-15
Fixed loop usage

flytestb:
Fixed loop when using custom slideActiveClass


nolimits4web:
Merged, thank you!

flytestb:
you are welcome
when can we expect the new release?

Issue #6496, opened 2023-03-16
Bad default value for lastClickTime


MRamonLeon:
In a recent Cypress test we faced an issue because of this:
https://github.com/nolimits4web/swiper/blob/8f48d7f3df9a70801527afaf1df73fc6402391f1/src/core/core.js#L185

My test tap the active image. The test runs very quickly and Swiper interprets that this `tap` event is actually a `double tap` because the time passed between this tap and the _previous_ click stored at `lastClickTime` is less than 300 ms.

That's not true, there was only a single tap, but since the previous tap is stored there and is initialized to now(), Swiper thinks there was two taps.

It could be resolved by initializing the `lastClickTime` to 0 for example.

I created a simple PR to tackle that. Don't hesitate to come up with a different solution. #6497

PR #6497, opened 2023-03-16
Initialize lastClickTime to very old time

MRamonLeon:
Related to: https://github.com/nolimits4web/swiper/issues/6496

This way we avoid Swiper to think a single tap is a double tap when we do it very fast, for example at Cypress tests.


nolimits4web:
Merged, thank you!

Issue #6530, opened 2023-03-29
Virtual slides severely limited by max CSS px values


dylan1951:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/p/sandbox/swiper-virtual-slides-forked-ro9ds1

### Bug description

Virtual slides stop displaying after a certain number even though the slides are in the DOM.

If you go to slide 100,000 in my example it may or may not display the slides depending on the size of the window.

Pretty sure this is because the 'left' CSS property of the swiper-slide's gets into the 10's of millions and eventually reaches the browser limit and so can no longer be displayed.

It's pretty annoying because I choose this library because of the virtual slides feature and I spent a decent while getting my slider to look nice :(

### Expected Behavior

Visible slides.

### Actual Behavior

Invisible slides.

### Swiper version

9.1.1

### Platform/Target and Browser Versions

macOS Firefox

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [x] I'm willing to open a PR

PR #6533, opened 2023-03-29
patch for very large sliders using virtual slides

dylan1951:
Rudimentary patch for issue #6530 

Patched it enough for my purposes but unlikely to be best practices and won't cover all usages. I'm new to web dev and open source so feedback is welcome.


dylan1951:
OK thanks I put it into a swiper method, also tested with RTL and vertical direction.

nolimits4web:
Merged, thank you!

PR #6544, opened 2023-03-31
feat: lazyPreloadPrevNext option to preload prev/next images

Carnoisseur1979:
Add option to preload a number of next and previous images, based on the new method of lazy loading.

This replaces the functionality lost when the lazy module was removed in v9.

Implemented in the core, based on my inital attempt in #6416.


nolimits4web:
Merged, thank you!

Issue #6552, opened 2023-04-02
Issues when the controller module is used when the Swiper Controller is destroyed


TakhyunKim:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/s/swiper-testing-bcm44v

### Bug description

I believe this is fixed in version 9.2.0, though,
I'm raising this as an issue because I think it could be done differently.

When using the Controller module with React 18 Strict mode enabled,
an error will be thrown if the sweeper control is destroyed on unmount.
(If you don't use StrictMode, you won't have this issue.)

I think there is a way to do something similar to the workaround for [that issue](https://github.com/nolimits4web/swiper/issues/6491).

In the conditional statements 182 and 186 of controller.js, add the condition `swiper.controller.control.destroyed`.
The example code is shown below.
```js
  on('setTranslate', (_s, translate, byController) => {
    if (!swiper.controller.control || swiper.controller.control.destroyed) return; // check
    swiper.controller.setTranslate(translate, byController);
  });
  on('setTransition', (_s, duration, byController) => {
    if (!swiper.controller.control || swiper.controller.control.destroyed) return; // check
    swiper.controller.setTransition(duration, byController);
  });
```

The existing conditional also handles early return when control doesn't exist, 
so it would be nice to do the same when it is destroyed.

### Expected Behavior

No errors are thrown

### Actual Behavior

An error occurs

### Swiper version

9.1.1

### Platform/Target and Browser Versions

macOS with chrome

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

PR #6555, opened 2023-04-03
add a conditional to early return if the swiper controller is destroyed

TakhyunKim:
Fixes to address this issue [#6552](https://github.com/nolimits4web/swiper/issues/6552)
I did some debugging to fix this bug and found a solution.

Check if the swiper controller is destroyed before the `setTranslate`, `setTransition` functions run.

[In other PR](https://github.com/nolimits4web/swiper/pull/6501), the above two functions are executed and checked for destruction within the setControlledTranslate and setControlledTransition functions, but I think it would be nice to check for this in advance to avoid running unnecessary logic.

This is my first open source PR. Feedback is always welcome.

### DIFF
--- a/src/modules/controller/controller.js
+++ b/src/modules/controller/controller.js
@@ -179,11 +179,11 @@ export default function Controller({ swiper, extendParams, on }) {
     removeSpline();
   });
   on('setTranslate', (_s, translate, byController) => {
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thank you!

TakhyunKim:
Thanks!

PR #6586, opened 2023-04-14
fix: made defineProperty configurable as it broke when minified.

PedroPovedaQ:
To reproduce the bug: Embed into a website.

Fixes the below issue where the build crashes when minified:

image
Minified Code Below: 

```
paramsList.forEach((paramName) => {
  if (paramName === 'init') return;
  paramName = paramName.replace('_', '');
  Object.defineProperty(SwiperContainer.prototype, paramName, {
    / / added configurable:true  HERE to fix crash
    get() {
      return (this.passedParams || {})[paramName];
    },
    set(value) {
      if (!this.passedParams) this.passedParams = {};
      this.passedParams[paramName] = value;
      if (!this.initialized) return;
      this.updateSwiperOnPropChange(paramName, value);
    },
  });
});
```

It seems that the property is redefined in a later loop but as it is read only it crashes. Making it configurable fixes this issue.

### DIFF
--- a/src/element/swiper-element.js
+++ b/src/element/swiper-element.js
@@ -198,6 +198,7 @@ paramsList.forEach((paramName) => {
   if (paramName === 'init') return;
   paramName = paramName.replace('_', '');
   Object.defineProperty(SwiperContainer.prototype, paramName, {
+    configurable: true,
     get() {
       return (this.passedParams || {})[paramName];
     },


nolimits4web:
There shouldn't be such issue in today's release, or is it still there?

PedroPovedaQ:
Hi,

I was using the following cdn include in my html:

`<script src="https://cdn.jsdelivr.net/npm/swiper@9/swiper-element-bundle.min.js"></script> `

PedroPovedaQ:
> There shouldn't be such issue in today's release, or is it still there?

I was using the following cdn include in my html:

`<script src="https://cdn.jsdelivr.net/npm/swiper@9/swiper-element-bundle.min.js"></script> `

Issue #6452, opened 2023-03-02
Mousewheel: Add passive listeners for mousewheel


Davi-T:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://swiperjs.com/demos/270-mousewheel-control/core

### Bug description

Chrome shows passive events violation when mouse wheel is enabled and notice it's not there in the source (but looks like it's added to other events)

Can we add in pretty please?

Thanks so much!

### Expected Behavior

No violation warning

### Actual Behavior

CHROME > [Violation] Added non-passive event listener to a scroll-blocking 'wheel' event. Consider marking event handler as 'passive' to make the page more responsive

### Swiper version

9.1.0

### Platform/Target and Browser Versions

Chrome 110.0.5481.100 

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

nolimits4web:
No, we can not. Passive listeners don't allow preventDefault

frzsombor:
I've found this issue after receiving the _"Does not use passive listeners to improve scrolling performance"_ warning in Google PageSpeed Insights / Lighthouse. I also found a few related issues: #3854, #4609, and #4610. Leaving this comment here in case someone else finds this issue through the same warning and is looking for more information.

The code part that triggers this warning (at least for me) is [this](https://github.com/nolimits4web/swiper/blob/v11.2.8/src/modules/mousewheel/mousewheel.mjs#L412) `wheel` event listener.
However, according to the [documentation](https://developer.chrome.com/docs/lighthouse/best-practices/uses-passive-event-listeners) of Lighthouse:

> Lighthouse uses the following process to identify event listeners that may affect scrolling performance:
> - Collect all event listeners on the page.
> - Filter out non-touch and non-wheel listeners.
> - **Filter out listeners that call preventDefault().** <= ❗❗❗
> - Filter out listeners that are from a different host than the page.

[...]

PR #6646, opened 2023-05-04
fix(wheel): Adding an active listener (#6452)

TakNePoidet:

Always follow the [contribution guidelines](https://github.com/nolimits4web/swiper/blob/master/CONTRIBUTING.md) when submitting a pull request.


nolimits4web:
What about the description? What does it fix?

TakNePoidet:
Removes the Chrome passive event warning from the mouse wheel

gerbenvandijk:
it would be great to have this implemented :) the passive event warning in chrome is influencing our lighthouse performance scores.

PR #6743, opened 2023-06-08
Parse breakpoints JSON if passed in `breakpoints` attribute

ryanm-bw:
Provides support for a JSON string to be passed as the `breakpoints` attribute on the Swiper element:

```
<swiper-container navigation="true" pagination="true" scrollbar="true" breakpoints='{"600": {"slidesPerView": 3},"900": {
"slidesPerView": 4}}'>
```

### DIFF
--- a/playground/element/index.html
+++ b/playground/element/index.html
@@ -3,12 +3,24 @@
 
 <head>
   <meta charset="UTF-8" />
-  <title>Swiper Playground</title>
+  <title>Swiper Playground - Element</title>
   <meta name="viewport" content="width=device-width" />
 </head>
 
 <body>
-  <swiper-container>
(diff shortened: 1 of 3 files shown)


PouryaMontakhab:
it does not work,
the following error generate:
```
Cannot use 'in' operator to search for '97' in {"320": {"slidesPerView": 1},"640": {"slidesPerView": 2},"992": {
            "slidesPerView": 3}}
```

ryanm-bw:
> it does not work, the following error generate:
> 
> ```
> Cannot use 'in' operator to search for '97' in {"320": {"slidesPerView": 1},"640": {"slidesPerView": 2},"992": {
>             "slidesPerView": 3}}
> ```

@PouryaMontakhab  please confirm you're using the updated `getBreakpoint`/`setBreakpoint` files from the PR?

nolimits4web:
It is now supported in v10

PR #6744, opened 2023-06-09
Update core.scss

catboy1006:
fix ios flicker

When I use video and Swiper in the webpage, there is a flicker of Swiper when I slide the page. I noticed that only the Android case is handled in the code. Please add the iOS part.


hkongm:
These phenomena only appear on the iPhone Pro & ProMax.
I suspect it has something to do with the LTPO mechanism.

catboy1006:
> 



I found this issue on iPhone 13, but strangely the code did not have any handling for iOS.

fzmorales (approved):

PR #6807, opened 2023-07-03
added overflow:hidden for fallback if clip is not supported in target…

XEngine:
Added fallback if clip is not supported


nolimits4web:
Thanks, I'm sure hidden rule mist be before the one with clip

PR #6852, opened 2023-07-12
Make VirtualOptions generic

Magda98:
I made `VirtualOptions` interface generic to be able to specify what is the type of item in `slides` array. By default item type is set to `any` to maintain compatibility.


nolimits4web:
Merged, thank you!

PR #7000, opened 2023-08-30
fix(react): fix react components props type

d4v1d82:
This PR fix an issue with typescript react components props definitions. 
In the original definition both SwiperProps and SwiperSlideProps are defined 2 times generating issue during the type check process.

### DIFF
--- a/src/swiper-react.d.ts
+++ b/src/swiper-react.d.ts
@@ -2,29 +2,455 @@ import * as React from 'react';
 
 import type { SwiperOptions, Swiper as SwiperClass } from './types/index.d.ts';
 
-interface SwiperProps extends SwiperOptions {
-  /**
-   * Swiper container tag
-   *
-   * @default 'div'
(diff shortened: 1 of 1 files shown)


arodik:
Hi, @d4v1d82 . Which TS version do you use? Looks like I have a similar problem, but your approach doesn't help me by some reason.
I created another fix which works for my case.
https://github.com/nolimits4web/swiper/pull/7004

Probably make sense to merge them both :D

d4v1d82:
Hi @arodik, I'm on typescript v4.9.5.
My fix worked on my project and I don't think I've issue with importing from index.
Btw.. yes, the types have some issue right now :P

arodik:
TS v5 fixes all these bugs btw, I suggest you to upgrade if you can 😅

nolimits4web:
Thank you, had to merge manually as there is some not needed extra in your file

PR #7004, opened 2023-09-01
Fix #7003. SwiperProps interface doesn't include SwiperOptions fields on TS4

### LINKED ISSUE #7003: Swiper/React - SwiperProps interface doesn't include SwiperOptions (Typescript 4.8)

- [X] I confirm

### Reproduction link

https://codesandbox.io/p/sandbox/swiper-10-2-interface-problem-vwn9g5

### Bug description

I've updated Swiper to the latest (10.2) version and have a problem. My typechecker fails because it doesn't see some properties such as `modules` in Swiper component's props. We use Typescript 4. Everything is ok on TS5, but unfortunately, we can't upgrade right now.

For some reason SwiperProps interface doesn't include fields from SwiperOptions, despite it extends from it.
https://github.com/nolimits4web/swiper/blob/master/src/swiper-react.d.ts

You can reproduce the issue in the linked codesanbox. Run `npm run tsc` and you'll see the error.

### Expected Behavior

All fields from `SwiperProps` are accessible, including fields from `SwiperOptions`

### Actual Behavior

Only fields from `SwiperProps` interface are available, `SwiperOptions` fields are missing

### Swiper version

10.2

Typescript 4

### Validations

arodik:
Fix for #7003 

Looks like Typescript 4 has some bugs related to imports/exports in .d.ts files. If we make direct imports instead of importing from index.d.ts - everything works. Checked on my project

Any idea why it happens?

### DIFF
--- a/src/swiper-react.d.ts
+++ b/src/swiper-react.d.ts
@@ -1,6 +1,8 @@
 import * as React from 'react';
 
-import type { SwiperOptions, Swiper as SwiperClass } from './types/index.d.ts';
+// Importing from index file causes errors in Typescript 4
+import type { SwiperOptions } from './types/swiper-options';
+import type SwiperClass from './types/swiper-class';
 
 interface SwiperProps extends SwiperOptions {
   /**


nolimits4web:
Thank you, but Swiper TS definitions are done to support TS v5, this PR can break `esnext` module resolution. Consider upgrading to TS v5

Issue #7048, opened 2023-09-19
Remove d.ts from your import paths!


infacto:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

not necessary or not possible

### Bug description

TypeScript is used incorrectly. The build fails.

```
TS2691: An import path cannot end with a '.d.ts' extension.
```

You can simply fix this by removing all `d.ts` suffix from path string.

**Before** 👎 

```ts
import type Swiper from './swiper-class.d.ts';
```

**After** 👍 

```ts
import type Swiper from './swiper-class';
```

(tested)

Othwesie it's not possible to use it in TypeScript anymore. It's a blocker. Workaround: Use previous versions like 9.

Affected files: 
- node_modules/swiper/types/swiper-options.d.ts
- node_modules/swiper/types/swiper-events.d.ts
- node_modules/swiper/types/swiper-class.d.ts
- node_modules/swiper/types/modules/zoom.d.ts
- node_modules/swiper/types/modules/thumbs.d.ts
- node_modules/swiper/types/modules/scrollbar.d.ts
- node_modules/swiper/types/modules/pagination.d.ts
- node_modules/swiper/types/modules/navigation.d.ts
- node_modules/swiper/types/modules/mousewheel.d.ts
- node_modules/swiper/types/modules/keyboard.d.ts
- node_modules/swiper/types/modules/hash-navigation.d.ts
- node_modules/swiper/types/modules/controller.d.ts
- node_modules/swiper/types/modules/autoplay.d.ts
- All of your type files.

### Expected Behavior

No errors during compiling.

### Actual Behavior

The build fails due to incorrect use of typescript.

### Swiper version

10.2.0

### Platform/Target and Browser Versions

Win 10

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

nolimits4web:
Or update to latest TS v5

infacto:
@nolimits4web Can you please explain why you insist on the file extensions when importing? I see no problems or disadvantages. Removing the d.ts suffix is better in any way.

Kbdq:
[...]

PR #7052, opened 2023-09-20
Fix type imports.

infacto:
TS2691: An import path cannot end with a '.d.ts' extension.

Fixes: #7048


nolimits4web:
Sorry, but removing extensions makes it broken for me in TS v5

infacto:
Now I tested in TypeScript v5 too. And removing the `d.ts` works in v5 and below. It's also not listed as [breaking change](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#breaking-changes-and-deprecations). I doubt it, otherwise 99.99% of packages with types won't work. Sorry, I just want to understand... Can you provide an example please?

nolimits4web:
When setting moduleResolution to nodenext and disabling skipLibCheck

Issue #6466, opened 2023-03-07
Provide Typescript definition for Element


chrisgwgreen:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

N/A

### Bug description

According to https://swiperjs.com/react, it is recommended to work with Swiper Element vs Swiper React as it's likely to be removed in the future. However, there are no typescript definitions causing web components e.g. `<swiper-container>` to fail.  

### Swiper version

9.1.0

### Platform/Target and Browser Versions

All

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

SanderCokart:
Yes we need this!

hu-tu-yi-shi:
same here

markcnunes:
I see there is a TS definition but the Element is a simple HTMLElement.

https://github.com/nolimits4web/swiper/blob/master/src/element/swiper-element.d.ts

Will it be extended with the Swiper kebab-case attributes?

See below the workaround for React if anyone is blocked because of this missing type definition, however, have in mind I don't fully know these types and I could be wrong about this type declaration but it has worked fine for me so far.

```tsx
import { ReactNode, RefObject, useRef } from 'react';
import Swiper, { SwiperOptions } from 'swiper';
import { register } from 'swiper/element/bundle';
import { SwiperSlideProps } from 'swiper/react';

/**
 * When you import Swiper custom elements from node modules, we need to manually register them.
 * It should be done only once and it registers Swiper custom elements globally.
 */
register();

type Kebab<
  T extends string,
  A extends string = '',
> = T extends `${infer F}${infer R}`
  ? Kebab<R, `${A}${F extends Lowercase<F> ? '' : '-'}${Lowercase<F>}`>
  : A;

/**
 * Helper for converting object keys to kebab case because Swiper web components parameters are available as kebab-case attributes.
 * @link https://swiperjs.com/element#parameters-as-attributes
 */
type KebabObjectKeys<T> = {
  [key in keyof T as Kebab<key & string>]: T[key] extends Object
    ? KebabObjectKeys<T[key]>
    : T[key];
};

[...]

PR #7055, opened 2023-09-21
fix(types): detection of custom html tags

GNURub:
**Before**: Detects custom tags as HTMLElement.

image

**After**: Detects tags as Swiper's custom element.
image

https://github.com/nolimits4web/swiper/issues/6466


nolimits4web:
Thank you!

PR #7072, opened 2023-09-29
fix: 6915

gablittledevshop:
The issue is that `swiper.loopFix()` doesn't run when the `slideToLoop` is called, so it gets stuck and doesn't loop. I added `swiper.loopFix()` to the underlying `slideTo` function after the transition to fix this.

### DIFF
--- a/src/core/slide/slideTo.mjs
+++ b/src/core/slide/slideTo.mjs
@@ -166,6 +166,11 @@ export default function slideTo(
         swiper.onSlideToWrapperTransitionEnd = null;
         delete swiper.onSlideToWrapperTransitionEnd;
         swiper.transitionEnd(runCallbacks, direction);
+
+        // Fix the loop if we're in loop mode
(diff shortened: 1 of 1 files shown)


gablittledevshop:
Changed status to WIP. It doesn't work when you jump more than 1 slide.

gablittledevshop:
Additional change: `slidesAppended` would return a weird value when going right so slides would be missing so I updated how it's calculated. Works with 3, 5, and 7 `slidesPerView`.

gablittledevshop:
This breaks autoplay.

gablittledevshop:
Autoplay fixed.

nolimits4web:
Hi! Thank you but it i think we should avoid fixing it in current version of Swiper. As it is for sure can break something else. This is already fixed in `next` branch for upcoming Swiper v11, you can check it there

gablittledevshop:
Ah ok thank you!

PR #7097, opened 2023-10-09
Split classes in createElement util by space

### LINKED ISSUE #7096: Failed to execute 'add' on 'DOMTokenList' for Scrollbar dragClass option

- [X] I confirm

### Reproduction link

https://codesandbox.io/s/swiper-dragclass-issue-nhyjzt

### Bug description

When using Tailwind with class fields on e.g. Scrollbar's dragClass option, Swiper is assuming a single classname. This should be split on spaces and passed to classList.add as an array.

### Expected Behavior

We should be able to pass multiple classes to this, as per #6344 and other related issues.

### Actual Behavior

See codesandbox:
```
Failed to execute 'add' on 'DOMTokenList': The token provided ('left-0 top-0 relative w-full h-full rounded-xl bg-font-tertiary') contains HTML space characters, which are not valid in tokens.
```

### Swiper version

10.3.1

Chrome 117.0.5938.149

### Validations

themagickoala:
I am using this change locally to fix #7096 via patch-package. It will ensure multiple space-separated classes are supported when creating elements.

### DIFF
--- a/src/shared/utils.mjs
+++ b/src/shared/utils.mjs
@@ -207,7 +207,7 @@ function elementChildren(element, selector = '') {
 
 function createElement(tag, classes = []) {
   const el = document.createElement(tag);
-  el.classList.add(...(Array.isArray(classes) ? classes : [classes]));
+  el.classList.add(...(Array.isArray(classes) ? classes : classes.split(' '));
   return el;
 }
 function elementOffset(el) {


nolimits4web:
Hi! Thank you, but this already have been fixed here https://github.com/nolimits4web/swiper/commit/89a6f7192f7da120487c224dc1b1a7668422321a

PR #7152, opened 2023-11-01
dont let autoplay module slide when its time over

momo-hash1:
Hello, when I was working on a project, I encountered a problem. Swiper doesn't provide a method for disabling slide movement when the time is over. In my project, I needed to animate elements on the slide when it disappears and appears, and it must also slide to the next after a certain period of time. Here, a problem arises: Swiper's Autoplay module moves slides automatically and doesn't provide any methods for disabling this. This is a problem because I needed to hide elements on the slide before the next slide appears. When i added disableAutoSlide to autoplay params, this parameter makes autoplay frezee then its time over, and unfreeze then slide changed by hand or programaticaly. I will be very glad, if you find this feature useful and merge this pull request


nolimits4web:
But that is the point of autoplay. You can then just enable/disable it manually within `slideChange` event

Issue #7173, opened 2023-11-09
when  direction  is  'vertical' , the  modlue  grid Parameter fill not  work 


hxzAugustinus:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/p/sandbox/bold-paper-q76jc9

### Bug description

when direction  is 'vertical' ,  if i change grid.fill from  'column'  to  'row' , it's not  work .

### Expected Behavior

change grid.fill from  'column'  to  'row' , it's sort work fine  and  calculate right   with spaceBetween

### Actual Behavior

when direction  is 'vertical' ,  if i change grid.fill from  'column'  to  'row' , it's not  work .

### Swiper version

11.0.3

### Platform/Target and Browser Versions

chrome  119.0.6045.124

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

PR #7175, opened 2023-11-09
fix(grid): coordination between 'direction  #7173

hxzAugustinus:
Just fixed modifications related to the initialization phase classe ， The other collaborative parts have not yet been solved.


nolimits4web:
Originally grid module is designed to be used with horizontal direction only, otherwise `rows` and `cols` will have a diferrent meaning in vertical direction, and it will require a lot of fixes across whole codebase

hxzAugustinus:
ok ,i will  close the request

PR #7183, opened 2023-11-13
fix(core): swipe to last with slidesPerView: auto

gzzhanghao:
Can't swipe to the last slide with **long touches** when `slidesPerView` is set to `auto`.

### Reproduction link

https://codesandbox.io/s/swiper-slides-per-view-auto-forked-d4mdyk?file=/index.html


nolimits4web:
Merged, thank you!

PR #7188, opened 2023-11-14
Only update context when values change to avoid extra re-renders

cvbuelow:
Currently the context is changing whenever the SwiperSlide is re-rendered since it isn't memoized. This fixes that to reduce re-renders of child components.


nolimits4web:
Would be good to see the CodeSandbox illustrating both issue and fix

nolimits4web:

PR #7201, opened 2023-11-22
fix(modules/a11y): filter out falsy pagination elems

jordanrastrick:
Similar to #6823 I am noticing intermittent exceptions generated in the `destroy()` method of the `a11y` module, in this case where the `el` argument to the callback passed to `paginationEl.forEach` is `undefined` and so the attempt to `.removeEventListener` is a `TypeError`. As in the previous case, because these exceptions are thrown from within an event handler they are uncatchable and crash the entire page.

It looks like the use of `makeElementsArray`, which filters out such falsy elements (and should otherwise behave identically to the existing code), is also required with the pagination element.


nolimits4web:
Merged, thank you

Issue #7216, opened 2023-11-27
Swiper incorrectly calculates last slide as activeSlide on load.


learyjk:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/s/swiper-bug-loop-centered-auto-width-kwj8ls?file=/index.html

### Bug description

a slider with slidesPerView set to 'auto' in loop mode and centeredSlides set to true calculates the last slide as the active slide on load. In the demo you'll see that the first slide shown is slide 7, not slide 1.

- changing the slide width in CSS from 20% to 21% causes the first slide to show.

- can also play with initialSlide value to correct it.

My guess is some sort of math error when calculating the first slide when combo of slidesPerView 'auto', centeredSlides true, and loop mode are all in play.

### Expected Behavior

- I expect Slide 1 to be the first to show in the provided example.
- I expect changing slide width in CSS to not change which slide is calculated to be first (i.e. I expect when initalSlide is set to 0, the first slide is shown.)

### Actual Behavior

The last slide in the html is calculated to be the active slide on load.

### Swiper version

11.0.5

### Platform/Target and Browser Versions

Chrome, Safari

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

huggie1984:
+1. 
Having a similar issue when component is mounted -> unmounted -> mounted. 
In some cases the last slide is the initial slide. 

We using the following modules and the following props.

`<Swiper
          slidesPerView={'auto'}
          spaceBetween={32}
          centeredSlides
          initialSlide={1}
          pagination={{
            clickable: true,
            dynamicBullets: true,
          }}
          navigation={{
            clickable: true,
            nextEl: '.button-next',
            prevEl: '.button-prev',
          }}
          modules={[Navigation, Pagination]}
        >`

learyjk:
@huggie1984 yea I've been doing more testing and getting the result at odd ranges of slide widths.

I'm setting slide width in % and spaceBetween is 16. The following percent values cause last slide to render first.
50-100%
24-33%
17-19%
13-14%

[...]

PR #7228, opened 2023-12-03
fix: initialSlide #7216

yasuhiro-yamamoto:
Fixed behavior where the combination of 'initialSlide:0' and 'slidesPerView:auto' would shift the first slide position

- getPropertyValue has an error when getting a decimal point value.
- 'slidesPerView:auto' caused an error in the actual number of slides displayed and the calculated number of slides displayed due to the error.

### DIFF
--- a/src/core/core.mjs
+++ b/src/core/core.mjs
@@ -337,7 +337,6 @@ class Swiper {
   }
 
   slidesPerViewDynamic(view = 'current', exact = false) {
-    console.log('debug');
     const swiper = this;
     const { params, slides, slidesGrid, slidesSizesGrid, size: swiperSize, activeIndex } = swiper;
     let spv = 1;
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thanks!

JarvisChao:
When setting slide width to a larger value, it still show the last slide on load.

https://codepen.io/jarvis73045/pen/MWxxWjR

yasuhiro-yamamoto:
@JarvisChao @nolimits4web 

I appreciate you taking the time to report this problem. 

Through your detailed bug report, I was able to reproduce and diagnose the cause of the slide index shifting incorrectly. 

Based on the investigation, I have implemented a solution by modifying the slide index calculation logic when moving to the initial displayed slide.

I opened a new pull request with the fix so we can get this resolved. (#7319)

PR #7284, opened 2024-01-29
Fix: Add Optional `swiperElementNodeName` Prop for (non swiper-element) Web Component Usage

sghoweri:
Our design system team recently ran into some major rendering issues using Swiper internally with some of our LitElement based web components (slotted children slides aren't detected). 

After doing some digging, I stumbled across the change added in #6870 which uses a hard-coded nodeName to ultimately determine if a particular swiper instance is rendering as a web component or not.

This PR should allow continued Swiper usage (as-is) via the `swiper-container` custom element, yet also help other folks with customized web component usage by using this new `swiperElementNodeName` parameter to define the custom element tag name being used in their implementation.

### DIFF
--- a/src/components-shared/params-list.mjs
+++ b/src/components-shared/params-list.mjs
@@ -7,6 +7,7 @@ const paramsList = [
   'init',
   '_direction',
   'oneWayMovement',
+  'swiperElementNodeName',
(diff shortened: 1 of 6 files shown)


nolimits4web:
Merged, thank you!

PR #7293, opened 2024-02-02
Fix for React strict mode double initialization errors

VolkerLieber:
React strict mode creates everything two times and destroys one instance again to check for potential leaks in dev move. 
This behaviour causes swiper to potentially handle an already destroyed instance.


nolimits4web:
Merged, thanks!

Issue #7309, opened 2024-02-10
Double tap to zoom does not reliably work on mobile


broox:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://swiperjs.com/demos/400-zoom/core

### Bug description

In the Zoom module, double tap to zoom does not reliably work in Chrome on my iOS device. It will sometimes zoom if I tap very quickly 3+ times in a row, but it does not work as expected, or like it does in any other browser.

It works as expected on MacOS Chrome, MacOS Safari, and iOS Safari but not iOS Chrome.

### Expected Behavior

When the `toggle` parameter is set to `true`, double tap should zoom into the coordinates that I tap on in all browsers.

### Actual Behavior

Double tapping an image does not zoom in/out of the image on Chrome in iOS.

### Swiper version

11.0.6

### Platform/Target and Browser Versions

iOS 17.2.1, Chrome 121.0.6167.171

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

broox:
the bug seems to be that the `doubleTap` event it not being emitted on chrome.
`touchEndTime - data.lastClickTime` is much greater than 300, even if `timeDiff` is well under 300.

broox:
PR incoming.

matiasmenker:
Any update on this ?

Wankishh:
Any update here?

broox:
FWIW - The `toggle` option seems to work with the version of swiper that I'm running now `11.1.14` with mobile chrome (`131.0.6778.134`) and safari (`18.2`on iOS. Can anyone else confirm whether this is an issue for them still?

PR #7310, opened 2024-02-10
Move click/tap events to onClick handler to fix bug with double tapping on mobile Chrome

broox:
There was an issue in which double tap events were not being emitted on mobile Chrome.

On mobile Chrome, the second touch event was not being emitted appropriately, which resulted in touch times being much longer than 300ms.

The `click` event is being emitted on all browsers, as expected. This change moves tap/click events to the `onClick` handler, which I think makes more sense. This change fixes the bug for me.

Closes: #7309


m345054:
Why hasn't this been merged yet? Is there a workaround?

m345054 (approved):

broox:
hi @nolimits4web, do you have concerns with this implementation? anything i can do to help?

siimsams (approved):

broox:
FWIW - this solution does not seem to work after rebasing the latest `master`. I'll try to sort out what's going on as time allows...

PR #7311, opened 2024-02-11
Add ability to constrain max zoom to 100% of original image size

### LINKED ISSUE #7303: Constrain max zoom to 100% of a slide's image size on the Zoom module

A numeric `maxRatio` multiplier can be specified as an option on the Zoom module, which is great, however the image can scale past a reasonable resolution if the image is small. It would be nice to be able to specify a `maxRatio` of 100% of the original image.

### Suggested solution

Allow a percentage to be passed into `maxRatio`, introduce a `maxRatioPercentage` property, or an optional boolean field to prevent zooming past 100% of the original image size.

### Alternative

_No response_

### Additional context

_No response_

### Validations

broox:
A numeric `maxRatio` multiplier can be specified as an option on the Zoom module, which is great, however the image can scale past the image's original resolution. This adds a new boolean parameter called `limitToOriginalSize` that will prevent zooming past the image's natural size.

Closes #7303

### DIFF
--- a/src/modules/zoom/zoom.mjs
+++ b/src/modules/zoom/zoom.mjs
@@ -11,6 +11,7 @@ export default function Zoom({ swiper, extendParams, on, emit }) {
   extendParams({
     zoom: {
       enabled: false,
+      limitToOriginalSize: false,
       maxRatio: 3,
       minRatio: 1,
       toggle: true,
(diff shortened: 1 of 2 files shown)


nolimits4web:
Merged, thanks!

PR #7317, opened 2024-02-14
Declare enabled option in mousewheel.d.ts

lifecoderua:
Declare `enabled` option for the Mousewheel options, to prevent the type check errors.

Currently TS build raise a (non-blocking) error on build:
error TS2353: Object literal may only specify known properties, and 'enabled' does not exist in type 'MousewheelOptions', using the following declaration:

```
<swiper
    :mousewheel="{
      enabled: true
    }"
  >
  ...
```


nolimits4web:
Merged, thanks!

Issue #7427, opened 2024-03-28
Typing of effectInit is wrong


9romise:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

[...]

Issue #7205, opened 2023-11-23
Typing of slideNext and slidePrev is wrong


sha237gh:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://stackblitz.com/edit/stackblitz-starters-igjhxg?file=src%2Fmain.ts

### Bug description

The typings for the functions `slideNext` and `slidePrev` are typed as voids in `node_modules/swiper/types/swiper-class.d.ts`, however when logging their return values it appears that they are indeed returning boolean values. In the stackblitz use the buttons to slide around and check the console.

### Expected Behavior

`slideNext` and `slidePrev` typings return booleans.

### Actual Behavior

`slideNext` and `slidePrev` typings return voids.

### Swiper version

11.0.5

### Platform/Target and Browser Versions

macOs Chrome 119.0.6045.159

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

Meshavegas:
i have the same issue,
next js 14

PR #7428, opened 2024-03-28
fix(types): fixed SwiperModule, slideTo, effectInit types

9romise:
Fixed some wrong types based on source code.

Closes: #7427 , #7205


nolimits4web:
t0ggles-create swiper

Issue #5436, opened 2022-02-08
centerInsufficientSlides doesn't count slidesOffsetBefore/slidesOffsetAfter params


vsavaryn-lohika:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

see attached screenshot from file updateSlides.js

### Bug description

`centerInsufficientSlides` doesn't count `slidesOffsetBefore/slidesOffsetAfter` params
As a result slides **NOT CENTERED** correctly.
When I remove slidesOffset params - it centers correctly.

<img width="740" alt="image" src="https://user-images.githubusercontent.com/91122565/153017033-49c14dda-9e54-4630-91b7-acef68930dea.png">

Please fix it.

### Swiper version

8.0.3

### Platform/Target and Browser Versions

all

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

Hanskrogh:
Would love for this to be implemented!

lennartdeknikker:
I'm bumping into this issue as well. I think this could be solved by not applying the slidesOffsetBefore & slidesOffsetAfter when the centerInsufficientSlides gets triggered or better yet by subtracting the offset from the x value in the translate3D function.

ovelilla:
I have the same problem

benwallis:
When using any SlidesOffset the wrapper gets position wrong (translate3D xpos is double/triple what it should be)

nolimits4web:
Issue is closed because of outdated/irrelevant/not actual/needed more information/inactivity.

If this issue is still actual and reproducible for latest version of Swiper, please create new issue and fill the issue template correctly:
* Clearly describe the issue including steps to reproduce when it is a bug.
* Make sure you fill in the earliest version that you know has the issue.
* Provide live link or JSFiddle/Codepen or website with issue

PR #7437, opened 2024-04-04
fix(core): centerInsuffientSlides takes offsets into account

slemling:
When using centerInsufficientSlides together with slidesOffsetBefore and slidesOffsetAfter, the slides won’t be centered correctly. Apparently slidesOffsetBefore and  slidesOffsetAfter are not taken into account in the calculation of the offset for centerInsufficientSlides.

[Codesandbox](https://codesandbox.io/p/sandbox/swiper-center-insufficient-slides-bug-98rqzr?file=%2Fsrc%2Findex.js)

Resolves https://github.com/nolimits4web/swiper/issues/5436 (was closed already due to lacking information / activity, but issue still remains)

### DIFF
--- a/src/core/update/updateSlides.mjs
+++ b/src/core/update/updateSlides.mjs
@@ -269,8 +269,9 @@ export default function updateSlides() {
       allSlidesSize += slideSizeValue + (spaceBetween || 0);
     });
     allSlidesSize -= spaceBetween;
-    if (allSlidesSize < swiperSize) {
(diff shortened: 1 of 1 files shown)


nolimits4web:
t0ggles-create swiper

sascha-eisenmann:
Hi @nolimits4web, I also ran into this problem. Any chance to get this released in the next few weeks? 

@slemling Thanks for the fix. This saves me from building a messy workaround.

PR #7516, opened 2024-05-03
Update updateSlides.mjs

personaljs:
The error is that the loop is trying to iterate over an array with zero length. `slideLength` takes the length from virtual slides, but `slides` are obtained from DOM


nolimits4web:
What issue does it fix?

nolimits4web:

PR #7556, opened 2024-05-26
Avoid rendering same slide vnode twice for small amount of slides in loop + virtual mode

artur-pipchenko:
This fix adds one extra logical check in order to avoid pushing more elements to `slidesToRender` that we have slides, i.e. we avoid duplicates. 

Refers to #7557 

Cases with 4+ slides work as before


nolimits4web:
Merged, thanks!

Issue #7609, opened 2024-07-10
SASS Deprecation Warning


stijns96:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

/

### Bug description

I'm using SCSS in my projects and use SASS to compile it. However, they now giving me this nesting warnings. [More info here](https://sass-lang.com/documentation/breaking-changes/mixed-decls/).

**The warnings only appear with sass v1.77.7**

```bash
Deprecation Warning: Sass's behavior for declarations that appear after nested
rules will be changing to match the behavior specified by CSS in an upcoming
version. To keep the existing behavior, move the declaration above the nested
rule. To opt into the new behavior, wrap the declaration in `& {}`.

More info: https://sass-lang.com/d/mixed-decls

   ╷
81 │ ┌   &:after {
82 │ │     content: 'next';
83 │ │   }
   │ └─── nested rule
84 │     right: var(--swiper-navigation-sides-offset, 10px);
   │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ declaration
   ╵
    ~swiper/modules/navigation.scss 84:3                @use
    src/assets/scss/modules/swiper/navigation.scss 1:1  root stylesheet

Deprecation Warning: Sass's behavior for declarations that appear after nested
rules will be changing to match the behavior specified by CSS in an upcoming
version. To keep the existing behavior, move the declaration above the nested
rule. To opt into the new behavior, wrap the declaration in `& {}`.

More info: https://sass-lang.com/d/mixed-decls

    ╷
81  │ ┌   &:after {
82  │ │     content: 'next';
83  │ │   }
    │ └─── nested rule
... │
85  │     left: auto;
    │     ^^^^^^^^^^ declaration
    ╵
    ~swiper/modules/navigation.scss 85:3                @use
    src/assets/scss/modules/swiper/navigation.scss 1:1  root stylesheet
```

### Expected Behavior

No warnings

### Actual Behavior

SASS warnings

### Swiper version

11.1.4

### Platform/Target and Browser Versions

macOS

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

[...]

PR #7612, opened 2024-07-11
Update navigation.scss to remove SASS Deprecation Warning

guillem-gelabert:
https://github.com/nolimits4web/swiper/issues/7609

Change order of properties to avoid SASS Deprecation Warning


nolimits4web:
Merged, thank you!

Issue #7623, opened 2024-07-18
Support using slots as swiper wrappers for web components


andreyoganesyan:
### Clear and concise description of the problem

As a developer using Swiper and developing web components, I want to be able to make swiper-wrapper a slot, so that the slides can be dynamically added to the wrapper when web components are used.

### Suggested solution

The current implementation uses wrapper's children property to detect slides, but slot's content is not included in its children, there is a separate [assignedElements()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedElements) method to access it.

So, I suggest making the following changes to the [elementChildren()](https://github.com/nolimits4web/swiper/blob/7d104dedd9014d08a6acdc8bbf7ce307acb0ff09/src/shared/utils.mjs#L205) util function:

```
function elementChildren(element, selector = '') {
  const children = [...element.children];
  if(element instanceof HTMLSlotElement) {
    children.push(...element.assignedElements())
  }

  if(!selector) {
    return children;
  }
  return children.filter((el) => el.matches(selector));
}
```

I also suggest adding the following utility function to be used wherever `wrapperEl.contains()` was used:

```
function elementIsChildOf(el, parent) {
  const children = elementChildren(parent);
  return children.includes(el);
}
```

To avoid duplication, I reused the elementChildren() function in elemtIsChildOf(), that's why I added a check for empty selector, since `el.matches('')` throws a DOMException.

### Additional context

I have tested this solution in a component library I'm working on, and changing these two functions does indeed enable support for using slots as wrappers. Here is a barebones example of what it looks like in a `custom-carousel` component:

```
<div class="swiper carousel">
    <slot class="swiper-wrapper"></slot>
</div>
```

Then, when using the `custom-carousel` element:

```
<custom-carousel>
    <div>Slide 1</div>
    <div>Slide 2</div>
    <div>Slide 3</div>
    <div>Slide 4</div>
</custom-carousel>
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this feature?

- [X] I'm willing to open a PR

defmech:
[...]

PR #7624, opened 2024-07-18
#7623: Added support for using slots as swiper wrappers

andreyoganesyan:
Implementation for issue #7623.

I extended the elementsChildren() function in utils.mjs so that it doesn't only check the children of the element, but also assignedElements in case it's a slot.

Please see the original issue for more context.


nolimits4web:
t0ggles-create swiper

nolimits4web:
Merged, thanks!

PR #7632, opened 2024-07-22
added new prop for a11y module - scrollOnFocus

SlDo:
Added the 'scrollOnFocus' property with a zero value, which will prevent scrolling to the element that was focused. At the same time, the default value will be 'true', which will avoid backward compatibility issues.


nolimits4web:
t0ggles-create swiper

nolimits4web:
Merged, thanks!

OSMANCAKlR:
This broke the swipe touch on mobile!

SlDo:
> This broke the swipe touch on mobile!

Can you provide more info about that?

Issue #7570, opened 2024-06-14
Ability to disable image lazy loading listeners


Xenossolitarius:
### Clear and concise description of the problem

I am running a multi row (swiper) image gallery which is actionable and I am lazy loading the images with my custom logic. Because I need the performance I didn't want to relay on the swiper but unfortunately it attaches listeners anyway. Which makes it 100% unusable for my usecase. Currently I am patching the swiper to stop that behavior.

### Suggested solution

```
diff --git a/shared/swiper-core.mjs b/shared/swiper-core.mjs
index 619c42a4ee03ac543db7869eb80aa7e782bccab2..06880c6f982f6d2d62dad27700a3ea13a5652a20 100644
--- a/shared/swiper-core.mjs
+++ b/shared/swiper-core.mjs
@@ -2860,9 +2860,9 @@ const events = (swiper, method) => {
   }
 
   // Images loader
-  el[domMethod]('load', swiper.onLoad, {
-    capture: true
-  });
+  // el[domMethod]('load', swiper.onLoad, {
+  //   capture: true
+  // });
 };
 function attachEvents() {
   const swiper = this;
@@ -3614,11 +3614,11 @@ class Swiper {
     if (params.breakpoints) {
       swiper.setBreakpoint();
     }
-    [...swiper.el.querySelectorAll('[loading="lazy"]')].forEach(imageEl => {
-      if (imageEl.complete) {
-        processLazyPreloader(swiper, imageEl);
-      }
-    });
+    // [...swiper.el.querySelectorAll('[loading="lazy"]')].forEach(imageEl => {
+    //   if (imageEl.complete) {
+    //     processLazyPreloader(swiper, imageEl);
+    //   }
+    // });
     swiper.updateSize();
     swiper.updateSlides();
     swiper.updateProgress();
@@ -3786,19 +3786,19 @@ class Swiper {
 
     // Attach events
     swiper.attachEvents();
-    const lazyElements = [...swiper.el.querySelectorAll('[loading="lazy"]')];
-    if (swiper.isElement) {
-      lazyElements.push(...swiper.hostEl.querySelectorAll('[loading="lazy"]'));
-    }
-    lazyElements.forEach(imageEl => {
-      if (imageEl.complete) {
-        processLazyPreloader(swiper, imageEl);
-      } else {
-        imageEl.addEventListener('load', e => {
-          processLazyPreloader(swiper, e.target);
-        });
-      }
-    });
[...]

PR #7638, opened 2024-07-24
feat(toggle-preloading): add param to disable preloading

Xenossolitarius:
Add ability to toggle preloading of images. If the user doesn't want to rely on the swiper implementation the user is unable to opt out of it and infers huge performance penalties.
More in https://github.com/nolimits4web/swiper/issues/7570

### DIFF
--- a/src/components-shared/params-list.mjs
+++ b/src/components-shared/params-list.mjs
@@ -90,6 +90,7 @@ const paramsList = [
   'slidePrevClass',
   'slideBlankClass',
   'wrapperClass',
+  'lazyPreload',
   'lazyPreloaderClass',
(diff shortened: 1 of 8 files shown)


nolimits4web:
Thanks a lot for the PR and for pushing this! The repo has since migrated to TypeScript, so this couldn't be merged as-is — the same approach landed reimplemented on the current codebase in 81345db82559f3dda08da1e3bcb575026e913e24 as the new `lazyPreload` parameter (default `true`, set to `false` to disable all built-in lazy image handling). Will be available in the next release. See https://github.com/nolimits4web/swiper/issues/7570 for details.

PR #7647, opened 2024-07-26
add types for minTranslate and maxTranslate functions

mwoess:
I only extended the types for the `maxTranslate` and `minTranslate` functions.


nolimits4web:
Thanks!

PR #7649, opened 2024-07-26
Bugfix: Fixed typo in utils/elementIsChildOf

andreyoganesyan:
The current implementation of elementIsChildOf function references a variable 'element', which is not defined in its scope:
```js
function elementIsChildOf(el, parent) {
  const isChild = parent.contains(el);
  if (!isChild && parent instanceof HTMLSlotElement) {
    const children = [...element.assignedElements()];
    return children.includes(el);
  }
  return isChild;
}
```

I believe it should've been 'parent', the current implementation just throws an error if parent is an HTMLSlotElement.


nolimits4web:
Thanks!

PR #7655, opened 2024-07-29
fix types of swiper-effect-utils.d.ts

kudohamu:
This PR fix two type issues of `swiper-effect-utils.d.ts`.

#### 1. change `import` of .d.ts file to `import type`

related: https://github.com/nolimits4web/Swiper/commit/30ce8e0c9c7ce32025275469e9480ca8c81f30a7

> A declaration file cannot be imported without 'import type'. Did you mean to import an implementation file './types/index.ts' instead?

fix_types_of_effect_utils_01

#### 2. Fix invalid optional parameter declaration of `createShadow`.

In making this correction, I checked that `suffix` parameter is specified at all places in this repo where `createShadow` is called.

> A required parameter cannot follow an optional parameter.

fix_types_of_effect_utils_02

### DIFF
--- a/src/swiper-effect-utils.d.ts
+++ b/src/swiper-effect-utils.d.ts
@@ -1,6 +1,6 @@
-import { Swiper, SwiperOptions } from './types/index.d.ts';
+import type { Swiper, SwiperOptions } from './types/index.d.ts';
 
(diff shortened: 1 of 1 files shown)


nolimits4web:
t0ggles-create swiper

nolimits4web:
Merged, thanks!

Issue #7300, opened 2024-02-09
zoom.in function with a custom ratio causes the position to jump if we have previously clicked on the photo


dkacperski97:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/p/sandbox/swiper-zoom-forked-87sdhv

### Bug description

1. call zoom.in with a custom ratio
2. double-click the photo to zoom out
3. then, when the initial zoom scale is set to 1, the first call of zoom.in with a custom ratio, zooms to the center of the photo
4. the second call of zoom.in with a custom ratio zooms to the position of the last click, which causes the position to jump, because the previous call zooms to different position

The bug is caused by these lines: https://github.com/nolimits4web/swiper/blob/fc308b377deee8cbbeb3c5cd41b1da710b82da8e/src/modules/zoom/zoom.mjs#L474-L477 
Swiper only zooms to the center of photo if currentScale === 1. Instead, swiper should clear image.touchesStart after currentScale is set to 1.

I couldn't reproduce the bug without calling zoom.in before clicking on the photo, not sure why.

### Expected Behavior

zoom.in should zoom to the same spot as the previous call

### Actual Behavior

The second call of zoom.in zooms to the position of the last click/touch.

### Swiper version

11.0.6

### Platform/Target and Browser Versions

macOS Firefox 122.0

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

PR #7663, opened 2024-08-01
Fix zoomIn after currentScale is set to 1

dkacperski97:
Fixes https://github.com/nolimits4web/swiper/issues/7300 

Now, `image.touchesStart.x` and `image.touchesStart.y` are correctly set to `undefined` every time we change the scale to 1.


nolimits4web:
A bit tweaked it and merged, thanks!

Issue #7270, opened 2023-12-21
Swiper Infinite loop scroll jumping


deividisss:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/p/sandbox/yl6ddq

### Bug description

As an example, I will use the Slider Infinite Loop demo with 9 slides: https://swiperjs.com/demos/200-infinite-loop/core. 
When you press the back button arrow to go from the first slide to the last slide and then press pagination dot to go from slide 9 to slide 8, you will notice a crazy jump through multiple slides, and it keeps happening when you switch between slides 8 and 9 using pagination dots.

### Expected Behavior

Slider should scroll one slide

### Actual Behavior

It scrolls multiple slides
https://github.com/nolimits4web/swiper/assets/48391144/0392a2c5-9847-44dd-9076-e4aed34486f5

### Swiper version

 v11.0.5

### Platform/Target and Browser Versions

Windows /Linux Version 120.0.6099.111 (Official Build) (64-bit)

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

hajirasyafi:
Yeah I can see it

hamzaejaz787:
I have same issue, only happens when using navigation to go from last to first then i clicking on any pagination in between and it jumps like crazy. Any help would be appriciated

and-t:
I wrote a quick fix to this issue, after dealing with it on production. For reference, my swiper configuration:

```
const swiper = new Swiper(".swiper", {
    modules: [Navigation, Pagination],
    slidesPerView: "auto",
    centeredSlides: true,
    spaceBetween: 15,
    grabCursor: true,
    loop: true,
    slidesPerGroup: 1,
    loopFillGroupWithBlank: false,
    loopAdditionalSlides: 2,
    lazy: {
        loadPrevNext: true,
        loadPrevNextAmount: 2,
    },
    keyboard: {
        enabled: true,
    },
    navigation: {
        nextEl: ".swiper-button-next",
        prevEl: ".swiper-button-prev",
    },
    pagination: {
[...]

PR #7690, opened 2024-08-19
fix(pagination) Fixed swiper Infinite loop scroll jumping

StarsLeopard:
fixed the BUG at
https://github.com/nolimits4web/swiper/issues/7270


nolimits4web:
Merged with a few tweaks, thanks!

PR #7708, opened 2024-08-28
Add containerRole parameter to a11y module

### LINKED ISSUE #7707: Feature Request for A11y: Add containerRole parameter

The Accessibility module currently allows for a aria-roledescription attribute on the slider container via the [containerRoleDescriptionMessage](https://swiperjs.com/swiper-api#param-containerRoleDescriptionMessage) option, but there is no support for setting a role to the same container.

As per https://www.w3.org/TR/wai-aria-1.1/#aria-roledescription, this is a violation of best practices, and my a11y auditors have flagged it as an issue to be fixed.

**Ensure aria-roledescription is only used on elements with an implicit or explicit role. Success criteria 4.1.2**

### Suggested solution

Add a containerRole parameter to the Accessibility module so we can set the proper role (region) when setting the containerRoleDescriptionMessage option.

### Additional context

Answers for @nolimits4web 

#### What is the issue? 

Providing users an ability to set an aria-roledescription without an aria role is an accessibility error.

#### What does it do? 

Adding this would allow Swiper to be more accessible.

#### What is it for? 

Better accessibility.

#### What does it fix? 

The accessibility error of having an aria-roledescription without a proper role.

#### Why you can't just add aria-role with "hands" to HTML element?

For the same reason you provide `containerRoleDescriptionMessage` in the first place: we don't always have direct control over the markup.

> Clearly describe the issue including steps to reproduce when it is a bug.
> Make sure you fill in the earliest version that you know has the issue.
> Provide live link or JSFiddle/Codepen or website with isssue

This is on all versions, it's a new feature being added, not a code bug to reproduce. The w3 guidelines for this are linked above.

### Validations

michaelbourne:
To fix accessibility error related to aria-roledescription by adding an additional role parameter to the a11y module.

Fixes #7707

### DIFF
--- a/src/modules/a11y/a11y.mjs
+++ b/src/modules/a11y/a11y.mjs
@@ -15,6 +15,7 @@ export default function A11y({ swiper, extendParams, on }) {
       slideLabelMessage: '{{index}} / {{slidesLength}}',
       containerMessage: null,
       containerRoleDescriptionMessage: null,
+      containerRole: null,
       itemRoleDescriptionMessage: null,
       slideRole: 'group',
       id: null,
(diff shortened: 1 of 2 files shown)


nolimits4web:
Merged, thanks!

Issue #7714, opened 2024-09-02
Search for the custom navigation control buttons in the swiper-container


Ashkanph:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/p/sandbox/sn7mfn

### Bug description

If there are more than one Swiper custom element with custom navigation buttons (for example in the `container-end` slot), if they use the same CSS selectors for their navigation buttons, clicking on one of the buttons swipes both sliders.

This can be solved by using different CSS selectors for the buttons, but in version 9 there was no need to have different CSS selectors; because the custom buttons were part of `swiper.el`, but in versions >= 10 the custom buttons are no longer descendants of `swiper.el`, they are descendants of `swiper.hostEl`.

### Expected Behavior

There is no need to have different CSS selectors for the custom navigation buttons of the Swiper custom elements on a page.

### Swiper version

11.1.12

### Platform/Target and Browser Versions

All of the platforms 

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

Ashkanph:
Is the following change acceptable to the maintainers as a solution?

https://github.com/nolimits4web/swiper/compare/master...Ashkanph:swiper:fix_navigation_getel

nolimits4web:
> Is the following change acceptable to the maintainers as a solution?
> 
> [master...Ashkanph:swiper:fix_navigation_getel](https://github.com/nolimits4web/swiper/compare/master...Ashkanph:swiper:fix_navigation_getel)

Yes, PR is welcome

nolimits4web:
Fixed by #7716

PR #7716, opened 2024-09-02
fix(element): Query the swiper.hostEl for the navigation buttons (#7714)

Ashkanph:
To be able to have more than one Swiper custom element on a page without having to have different CSS selectors for their custom navigation buttons.

A PR for the issue #7714


nolimits4web:
Merged, thank you!

PR #7739, opened 2024-09-16
Remove 'Lazy' declaration

diego-betto:
`Lazy` is no longer exported since [9.0.0](https://github.com/nolimits4web/Swiper/compare/v8.4.7...v9.0.0), 2023-02-01
But it still exists in the d.ts file. This can be confusing.


nolimits4web:
Merged, thanks!

diego-betto:
You're welcome, thank you!

Issue #7565, opened 2024-05-30
Thumbs Module not working on Strict Mode


JavoByte:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

[...]

PR #7789, opened 2024-11-08
Fix React strict mode double initialization errors for thumbs

wuhaochen:
The same issue we fixed in #7293 is causing #7565.

This patch would handle thumbs inits more gracefully in strict mode.


nolimits4web:
Merged, thanks!

Marvelousoneofall (approved):

PR #7813, opened 2024-12-04
feat: add sideEffects to package.json

denisx:
Add sideEffects for better code split

It will help bundlers cut unused code then you use named imports


denisx:
@swiper @tomjn hello here!

tomjn:
@denisx was there a reason to ping me? I'm not affiliated with this project, I just had a bug a few years ago I had issues with

PR #7815, opened 2024-12-04
fix: use find where possible instead of filter

siimsams:
I love swiper! This project is super fast. I'm using this in a calendar project which has multiple swipers on one page. I'm facing some performance issues on slow mobile phones and I'm trying to figure out ways to help out and make this library even more performant.

I found this previous [similar PR](https://github.com/nolimits4web/swiper/pull/7235) which was not merged. I guess it had too many changes. I tried to keep this small and concise as possible so it's easy to see what changed.

I think swapping [filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) out for [find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) should bring some performance gains and is quite safe thing to do. 

I will keep looking for ways to increase performance of this awesome project.

### DIFF
--- a/src/components-shared/get-element-params.mjs
+++ b/src/components-shared/get-element-params.mjs
@@ -69,9 +69,7 @@ function getParams(element, propName, propValue) {
     attrsList.push({ name: propName, value: isObject(propValue) ? { ...propValue } : propValue });
   }
   attrsList.forEach((attr) => {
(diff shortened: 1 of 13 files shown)


nolimits4web:
Looks good, thank you! Merged

PR #7818, opened 2024-12-05
Allow using a CSS selector as breakpointsBase

### LINKED ISSUE #7816: Allow specifying the breakpoint container element

When using `breakpointsBase: 'container'`, the container element will be used as the size reference.

In our situation, we need the swipes per slide to change based on the size of a parent higher up the DOM tree.

### Suggested solution

I suggest allowing the setting of `breakpointsBase` as a DOM element or a CSS selector. This would make it much more flexible. Or we can have a separate parameter like `breakpointBaseContainer` when `breakpointsBase` is `container`.

### Alternative

_No response_

### Additional context

_No response_

### Validations

kilobyte2007:
Fixes #7816 

Allows specifying a CSS selector for the breakpoints base so that not only the container can be referenced but also elements higher up the DOM tree.

I've tested the solution and it works well. Let me know if the solution looks good or whether any changes are needed.

### DIFF
--- a/src/core/breakpoints/setBreakpoint.mjs
+++ b/src/core/breakpoints/setBreakpoint.mjs
@@ -10,8 +10,10 @@ export default function setBreakpoint() {
   const breakpoints = params.breakpoints;
   if (!breakpoints || (breakpoints && Object.keys(breakpoints).length === 0)) return;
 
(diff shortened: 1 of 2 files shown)


nolimits4web:
Merged, thank you!

PR #7830, opened 2024-12-22
Fix: Ensure the Zoom module's  array is cleared onTouchEnd

### LINKED ISSUE #7304: Long pressing on a zoomable slide breaks zoom interactions for that slide

- [X] I confirm

### Reproduction link

https://swiperjs.com/demos/400-zoom/core

### Bug description

On mobile, if you long press an image on a given slide (e.g. to copy or save the image), you can no longer zoom in and out of that slide until you leave the slide and come back.

### Expected Behavior

You should be able to long press on an image to copy/save it, and continue to interact with the slide.

### Actual Behavior

Zoom gets stuck in the last state.

### Swiper version

11.0.6

iOS Chrome

### Validations

broox:
This fixes an issue in which long pressing a slide's image left the zoom module in a bad state.

Closes #7304

### DIFF
--- a/src/modules/zoom/zoom.mjs
+++ b/src/modules/zoom/zoom.mjs
@@ -28,7 +28,6 @@ export default function Zoom({ swiper, extendParams, on, emit }) {
   let isScaling = false;
   let fakeGestureTouched;
   let fakeGestureMoved;
-  let preventZoomOut;
   const evCache = [];
   const gesture = {
     originX: 0,
@@ -378,6 +377,7 @@ export default function Zoom({ swiper, extendParams, on, emit }) {
   }
   function onTouchEnd() {
     const zoom = swiper.zoom;
+    evCache.length = 0;
     if (!gesture.imageEl) return;
     if (!image.isTouched || !image.isMoved) {
       image.isTouched = false;


src/modules/zoom/zoom.mjs
```diff
@@ -28,7 +28,6 @@ export default function Zoom({ swiper, extendParams, on, emit }) {
   let isScaling = false;
   let fakeGestureTouched;
   let fakeGestureMoved;
-  let preventZoomOut;
```
broox, line 31: 🧹 this was simply unused.

src/modules/zoom/zoom.mjs
```diff
@@ -378,6 +377,7 @@ export default function Zoom({ swiper, extendParams, on, emit }) {
   }
   function onTouchEnd() {
     const zoom = swiper.zoom;
+    evCache.length = 0;
```
broox, line 380: On Chrome for iOS, long pressing a slide's image would leave the previous touch event in the `evCache` array.
This ensure that `evCache` is cleared `onTouchEnd`.

nolimits4web:
Merged, thank you!

PR #7848, opened 2025-01-09
Fix: Typo in autoplay param check

selimhex:
There seems to be a typo in Mousewheel module, where it checks for an autoplay param
```js
 // Stop autoplay
        if (swiper.params.autoplay && swiper.params.autoplayDisableOnInteraction)
```
by which i believe
```js
swiper.params.autoplay.disableOnInteraction
```
is meant.


nolimits4web:
Merged, thank you!

PR #7883, opened 2025-02-06
Export swiper-vars.scss for overwriteability

### LINKED ISSUE #7882: Export swiper-vars.scss for overwriteability

Currently, the [`src/swiper.scss`](https://github.com/nolimits4web/swiper/blob/master/src/swiper.scss) file features following line of code:
```scss
@use 'swiper-vars.scss' as vars;
```
This is so the variables inside [`src/swiper-vars.scss`](https://github.com/nolimits4web/swiper/blob/master/src/swiper-vars.scss) (which is only `$themeColor` at the time of writing) is included into the namespace of `vars`.
This included variable is used once in [`src/core/core.scss`](https://github.com/nolimits4web/swiper/blob/master/src/core/core.scss) via `vars.$themeColor`.

However, as a developer using swiper, this now complicates overwriting said variable in my own code.
There is **_no way_** of overwriting `$themeColor` for example by:
```scss
@use "~swiper/scss" with (
    $themeColor: /* some value */,
);
```
because the variable cannot be overwritten through two levels of `@use`.
This will generally hold true for any variable that might be added to the file in the future.

### Suggested solution

In the [`src/copy/package.json`](https://github.com/nolimits4web/swiper/blob/master/src/copy/package.json) we could export the [`src/swiper-vars.scss`](https://github.com/nolimits4web/swiper/blob/master/src/swiper-vars.scss) file to `./scss/vars`, for example:

```diff
{
  /* ... */
  "exports": {
    /* ... */
    "./scss/thumbs": "./modules/thumbs.scss",
+   "./scss/vars": "./swiper-vars.scss",
    "./scss/virtual": "./modules/virtual.scss",
    /* ... */
  },
}
```

This would allow for following code to work:
```scss
@use "~swiper/scss/vars" with (
    $themeColor: /* some value */,
);
@use "~swiper/scss";
```

### Alternative

There is an alternative to use `@import` instead of `@use` as was the case in older versions of Swiper, but this is not recommended since the use of the old `@import` is meant to be replaced by `@use` (see also https://sass-lang.com/documentation/at-rules/use/#differences-from-import).

### Additional context

_No response_

### Validations

entoncode:
Fixes #7882

The [`src/swiper-vars.scss`](https://github.com/nolimits4web/swiper/blob/master/src/swiper-vars.scss) file is now exported in package.json so that developers using Swiper can override the variables it contains.

### DIFF
--- a/src/copy/package.json
+++ b/src/copy/package.json
@@ -95,6 +95,7 @@
     "./scss/parallax": "./modules/parallax.scss",
     "./scss/scrollbar": "./modules/scrollbar.scss",
     "./scss/thumbs": "./modules/thumbs.scss",
+    "./scss/vars": "./swiper-vars.scss",
     "./scss/virtual": "./modules/virtual.scss",
     "./scss/zoom": "./modules/zoom.scss",
     "./element": {


nolimits4web:
Thank you! Merged. In meanwhole i recommend you to switch to Swiper's CSS files instead. Sass/Less files will be removed in next update. There is no any point in them anymore here

Issue #7233, opened 2023-12-04
Swiper Element: lazyload preloader is not working as expected 


nikitasol:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codesandbox.io/p/sandbox/swiper-lazy-load-images-forked-gt7jrr

### Bug description

After going through the list of slides and looping back to the first slide the lazy-preloader becomes visible and never disappears.

### Swiper version

@11

### Platform/Target and Browser Versions

macOS Chrome 119, macOS Safari 17.1

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

ayuminy:
In my case, the same phenomenon as above occurred in both versions 9 and 10 so that I did down grade to version 8.
macOS Chrome and Safari, windows Chrome.

Is this trouble cause I'm using a CDN?

niktor76:
Same problem here. 
As a workaround the preloader-icon can be hidden with:
`swiper-slide::part(preloader) {
  display: none;
}`

bibullo:
same issue, a new preloader is added every time we loop through the list

![image](https://github.com/nolimits4web/swiper/assets/45494960/869319aa-9913-43a0-8180-07d02581068d)

salasmx:
v11 and the issue still persists. Is there any timeline for resolving this?

FuTangKPS:
Any update?

sinjuangajan:
Still not fixed?

eduardobattisti:
I just opened a PR fixing the problem. Waiting for reviews.

PR #7886, opened 2025-02-10
fix: remove infinite preloader in loop

eduardobattisti:
If approved, this PR:

- Add a condition to check if the swiper-slide flag `swiperLoopMoveDOM`, avoiding calling the render method again, which was generating the infinite pre-loader injection.

Fixes: #7233


nolimits4web:
Merged, thank you!

Issue #7891, opened 2025-02-14
Swiper V2 Incompatible With ES2019


SteveCorbett:
### Clear and concise description of the problem

Due to the use of an optional chaining operator in V11.2.0 Swiper is no longer compatible with any JavaScript version prior to ES2020.

### Suggested solution

Swiper should remain compatible with JavaScript versions prior to ES2020 due to the number of projects that are unable to be updated to later versions.

### Alternative

Currently, the only option for a project using an earlier JavaScript version is to install V11.1.15

### Validations

- [x] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://swiperjs.com/swiper-api).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this feature?

- [x] I'm willing to open a PR

PR #7892, opened 2025-02-14
Restore compatibility with ES2019 #7891

SteveCorbett:
This PR will enable projects that are compiled with a version of JavaScript prior to ES2020 to use Swiper V11.2.0 or above.


nolimits4web:
Merged, thanks!

Issue #7896, opened 2025-02-17
❌ Wrong example code on https://swiperjs.com/swiper-api#param-renderBullet on Pagination


regepan:
### Check that this is really a bug

- [x] I confirm

### Reproduction link

https://swiperjs.com/demos/070-pagination-custom/core

### Bug description

```js
// ❌ Wrong renderBullet position

const swiper = new Swiper('.swiper', {
  //...
  renderBullet: function (index, className) {
    return '<span class="' + className + '">' + (index + 1) + '</span>';
  }
});
```

### Expected Behavior

```js
// ✅ Correct renderBullet position

var swiper = new Swiper(".mySwiper", {
  pagination: {
    el: ".swiper-pagination",
    clickable: true,
    renderBullet: function (index, className) {
      return '<span class="' + className + '">' + (index + 1) + "</span>";
    },
  },
});
```

### Swiper version

11.2.4

### Platform/Target and Browser Versions

macOS Chrome 133.0.6943.98

### Validations

- [x] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://swiperjs.com/swiper-api).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [x] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [x] I'm willing to open a PR

PR #7897, opened 2025-02-17
[#7896] Fix Pagination renderBullet wrong example code

regepan:
https://github.com/nolimits4web/swiper/issues/7896


nolimits4web:
Hi, thanks, but it is also wrong. Pushed correct fix

Issue #7881, opened 2025-02-05
exception for early unmounting if prop loop="true"


DaniilIsupov:
### Check that this is really a bug

- [x] I confirm

### Reproduction link

https://github.com/DaniilIsupov/vue-swiper-reproduction-unmounted

### Bug description

exception for early (until end of mounting) unmounting
it's relevant only when prop `:loop="true"`

error in dev mode
![Image](https://github.com/user-attachments/assets/ff16b823-68f7-41f8-9fe1-8684a678880e)

error in prod mode
![Image](https://github.com/user-attachments/assets/4f5b189d-c4c1-42a9-8f47-617bc1ccdab2)

### Expected Behavior

behavior without unhandled exceptions, such as what happens when prop `:loop="false"`

### Swiper version

11.2.2

### Platform/Target and Browser Versions

macOS Chrome 132.0.6834.160

### Validations

- [x] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://swiperjs.com/swiper-api).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [x] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

nolimits4web:
t0ggles-create swiper

eduardobattisti:
Hey!

I did some tests regarding this case, and it is specific to when Vue Router is used. Since the component uses render functions, the unmount happens before the children are fully rendered. In this case, I added verification to check if `slidesEl` is defined during the loopDestroy process.

When testing without Vue Router, the component renders as expected.

If the approach is approved, I have already created a PR that solves the problem.

PR #7906, opened 2025-02-22
feat: check if slidesEl is defined in loopDestroy

eduardobattisti:
If approved, this PR:

- Checks if `slidesEl` is defined on `loopDestroy`, avoiding cases where components can be unmounted too soon to complete render.

This is a specific case, more information is provided in the related issue.

Fixes: #7881


nolimits4web:
t0ggles-create swiper

nolimits4web:
Merged, thanks!

Issue #7666, opened 2024-08-04
Swiper Element - Unable to Reinitialize After Destroying


millertchris:
### Check that this is really a bug

- [X] I confirm

### Reproduction link

https://codepen.io/millertchris/pen/abgWGKW?editors=1011

### Bug description

# Swiper Element - Unable to Reinitialize After Destroying

## Description

When using the Swiper Web Component (Swiper Element) with the `init="false"` attribute, there is an issue when attempting to reinitialize the element after destroying it. The initial setup and destruction work as expected, but reinitializing the element after it has been destroyed does not work.

## Steps to Reproduce

1. Set up the Swiper Element with `init="false"` attribute.
2. Manually initialize the Swiper using the `initialize()` method.
3. Destroy the Swiper using the `destroy()` method.
4. Attempt to reinitialize the Swiper using the `initialize()` method again.

## Expected Behavior

The Swiper Element should be reinitialized and function as expected after being destroyed and reinitialized.

## Actual Behavior

The Swiper Element does not reinitialize after being destroyed and reinitialized. It remains non-functional upon attempting to initialize it again.

## Code Example

```html
<swiper-container init="false">
  <!-- Slides here -->
</swiper-container>

<script>
  const swiperEl = document.querySelector('swiper-container');
  
  // Initial initialization
  swiperEl.initialize();
  
  // Destroy the swiper
  swiperEl.destroy();
  
  // Attempt to reinitialize
  swiperEl.initialize();
</script>
```

## Environment

- Swiper version: 11.1.9

## Additional Information

This issue impacts dynamic content scenarios where Swiper instances may need to be reinitialized based on user interactions or other dynamic changes in the application.

### Expected Behavior

The Swiper Element should be reinitialized and function as expected after being destroyed and reinitialized.

### Actual Behavior

The Swiper Element does not reinitialize after being destroyed and reinitialized. It remains non-functional upon attempting to initialize it again.

### Swiper version

11.1.9

### Platform/Target and Browser Versions

macOS Chrome 125.0.6422.176

### Validations

- [X] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://swiperjs.com/swiper-api).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [X] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

[...]

PR #7907, opened 2025-02-23
refactor(element): use swiper state for init/destroy tracking

eduardobattisti:
If approved, this PR:

- Uses swiper state to define the initialization/destroy behavior. Using the state defined in the web component was not reflected, because the `initialize` and `destroy` methods only update the `swiper` instance state.

In the case of `disconnectedCallback`. Only will it be triggered if the component is removed from the DOM.

Fixes: #7666

### DIFF
--- a/src/swiper-element.mjs
+++ b/src/swiper-element.mjs
@@ -147,7 +147,7 @@ class SwiperContainer extends ClassToExtend {
   }
 
   initialize() {
-    if (this.swiper?.initialized) return;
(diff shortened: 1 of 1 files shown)


eduardobattisti:
I noticed that most of the codebase is not using the `optional chaining`. If necessary, I can update the code.

nolimits4web:
t0ggles-create swiper

nolimits4web:
Merged, thank you!

PR #7936, opened 2025-03-28
fix(element): Add existance Check for params on lazyPreloaderClass

### LINKED ISSUE #7935: Uncaught TypeError: Cannot read properties of undefined (reading 'lazyPreloaderClass')

- [x] I confirm

### Reproduction link

https://codesandbox.io/p/devbox/swiper-default-forked-6c96g3?workspaceId=ws_NcfeSzfz6N96RJW7VKTnaL

### Bug description

We are getting an error `Uncaught TypeError: Cannot read properties of undefined (reading 'lazyPreloaderClass')`.

Due to our complex case (combining VueJS with other technologies) we cannot reproduce the issue.

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Swiper version

11.2.6

macOS Arc Browser, same issue on Windows/Linux/iOS/Android and on other browsers too.

### Validations

BerkeAras:
Added Check, if params does not exist

Closes #7935

### DIFF
--- a/src/shared/process-lazy-preloader.mjs
+++ b/src/shared/process-lazy-preloader.mjs
@@ -1,17 +1,17 @@
 export const processLazyPreloader = (swiper, imageEl) => {
   if (!swiper || swiper.destroyed || !swiper.params) return;
-  const slideSelector = () => (swiper.isElement ? `swiper-slide` : `.${swiper.params.slideClass}`);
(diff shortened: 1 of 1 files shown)


nolimits4web:
Don't get why it is needed? We have this line before:
```
if (!swiper || swiper.destroyed || !swiper.params) return;
```
which already will stop code execution if there is no `swiper.params`

PR #7945, opened 2025-04-02
fix(types): fix `effect` param type

### LINKED ISSUE #7944: Type IntelliSense doesn't work properly on the `effect` param

- [x] I confirm

### Reproduction link

https://codesandbox.io/p/sandbox/swiper-effect-param-type-3f4565

### Bug description

The expected type of the `effect` param is `'slide' | 'fade' | 'cube' | 'coverflow' | 'flip' | 'creative' | 'cards' | string | undefined`, but the resulting type is `string | undefined`, so that code completion on this param doesn't work as expected.

### Expected Behavior

Code completion works properly on the `effect` parameter while allowing an arbitrary `string` value to be passed.

Image

### Actual Behavior

Code completion doesn't work as expected because the resulting type of the `effect` param  is `string | undefined`.

Image

### Swiper version

11.2.6

mac OS

### Validations

m-kawafuji:
Fixes #7944

### Expected Behavior
Code completion works properly on the `effect` parameter while allowing an arbitrary `string` value to be passed.
Screenshot 2025-04-02 at 22 43 04

### Actual Behavior
Code completion doesn't work as expected because the resulting type of the effect param is `string | undefined`.
Screenshot 2025-04-02 at 22 47 20

### DIFF
--- a/src/types/swiper-options.d.ts
+++ b/src/types/swiper-options.d.ts
@@ -210,7 +210,7 @@ export interface SwiperOptions {
    *
    * @default 'slide'
    */
-  effect?: 'slide' | 'fade' | 'cube' | 'coverflow' | 'flip' | 'creative' | 'cards' | string;
+  effect?: 'slide' | 'fade' | 'cube' | 'coverflow' | 'flip' | 'creative' | 'cards' | (string & {});
 
   /**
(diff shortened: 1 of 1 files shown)


nolimits4web:
Didn't know about such hack. Merged, thank you!

PR #7966, opened 2025-04-22
fix(vue): added missing props

ExileofAranei:
Added missing `loopAdditionalSlides` and `loopAddBlankSlides` props to Vue `Swiper` component.
They are listed in `swiper-vue.d.ts`  typings, but not in props: https://github.com/nolimits4web/swiper/blob/10084e3c605dad488ff2d517f872d3b23763a581/src/swiper-vue.d.ts#L259-L263


nolimits4web:
Merged, thank you!

PR #7987, opened 2025-05-13
Add template for CMS

marcospassos:
I've helped so many teams integrate React Swiper with a CMS that I decided to turn it into a [template](https://croct.com/templates/interface/section/react-swiper-carousel).

This one shows an example using a slideshow for e-commerce, where content can be updated without touching code:
image

You get a fully working setup with just one command:

```bash
npx croct@latest use npm://swiper
```

It uses Croct, which has a forever-free plan and works great for managing this kind of dynamic content.

Hopefully it saves someone the time I've spent wiring this up in the past.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1041,6 +1041,15 @@ Swiper is not compatible with all platforms, it is a modern touch slider which i
 - **Accessibility (A11y)**
 - **And many more ...**
 
+### CMS integration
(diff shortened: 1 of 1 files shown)


nolimits4web:
Seems like an ad of some specific CMS

PR #7989, opened 2025-05-17
Bug fix for slideResetTransition events

igierard:
slideResetTransitionStart and slideResetTransitionEnd events never trigger b/c the reset event emit is inside a conditional that would never trigger. This moves the reset events to a location that will actually emit.


nolimits4web:
t0ggles-create swiper

nolimits4web:
Merged, thank you!

Issue #7983, opened 2025-05-08
Having `init="false"` leads to layout shifts


krzksz:
### Check that this is really a bug

- [x] I confirm

### Reproduction link

https://colorful-grizzled-oboe.glitch.me/

### Bug description

I'm working on a website that uses Swiper Element. There are multiple components that wrap it to provide custom options so the `init="false"` is used in that cases. Because there can be multiple carousels inserted into the HTML at once, I wanted to split their initialisation into separate tasks to not block the main thread for too long.

The issue is that any delay in calling the `.initialize()` leads to a layout shift. This is caused by the fact that both `<swiper-container>` and `<swiper-slide>` call `this.attachShadow()` in their constructors. It makes the browser hide all of their content until `initialize()` moves all elements into slots.

### Expected Behavior

I would expect the carousel to retain its appearance until the `.initialize()` function is called, especially when `init="false"` option is provided.

### Actual Behavior

If the `<swiper-container>` and `<swiper-slide>` elements are already registered, they will hide their whole content until the `.initialize()` function is called.

### Swiper version

v11.2.6

### Platform/Target and Browser Versions

macOS, Chrome

### Validations

- [x] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://swiperjs.com/swiper-api).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [x] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [x] I'm willing to open a PR

nolimits4web:
This is expected behavior

PR #7992, opened 2025-05-20
fix(element): attach container shadow DOM on init

krzksz:
This PR addresses the issue I described in https://github.com/nolimits4web/swiper/issues/7983.

It moves the shadow DOM attachment for the `SwiperContainer` element from the `constructor` to the `initialize` function. This makes sure that the elements stay visible all the time as the `render` function is triggered synchronously right after.

The change is not needed for the `SwiperSlide` as it moves its contents immediately anyway.

### DIFF
--- a/src/swiper-element.mjs
+++ b/src/swiper-element.mjs
@@ -40,8 +40,6 @@ const addStyle = (shadowRoot, styles) => {
 class SwiperContainer extends ClassToExtend {
   constructor() {
     super();
-
-    this.attachShadow({ mode: 'open' });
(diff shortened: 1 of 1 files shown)


krzksz:
@nolimits4web Thank you for looking into the issue. I can see you closed it as "expected behaviour". The fix seems quite simple, but feel free to close this PR if you don't like it.

Thanks again!

adityapradhan-lf:
Hey @nolimits4web, just following up on this. Thank you.

PR #8006, opened 2025-05-27
fix: detect contenteditable element in keyboard module

### LINKED ISSUE #7665: Keyboard control active while contenteditable element is focused

- [X] I confirm

### Reproduction link

https://jsfiddle.net/yb5s2kot/

### Bug description

When using SwiperJS with keyboard control enabled, the slider incorrectly responds to left and right arrow key inputs when a contenteditable element is focused. Instead of moving the input caret within the contenteditable element, the SwiperJS slides are navigated. Just like on an input element, when a contenteditable element is focused, keyboard controls should be disabled.

### Expected Behavior

Keyboard controls should be disabled when a contenteditable element is focused

### Actual Behavior

Keyboard controls remain active when a contenteditable element is focused, preventing the user from using the arrow keys to move the text input caret

### Swiper version

11.1.9

Windows 10 / Firefox

### Validations

ahmadkzx:
Fixes an issue where Swiper responds to arrow key presses even when a `contenteditable` element is focused. This prevents users from moving the caret inside editable text.

This PR adds a check to ignore keyboard navigation when a `contenteditable` element is active.

Closes #7665

### DIFF
--- a/src/modules/keyboard/keyboard.mjs
+++ b/src/modules/keyboard/keyboard.mjs
@@ -50,9 +50,10 @@ export default function Keyboard({ swiper, extendParams, on, emit }) {
     }
     if (
       document.activeElement &&
-      document.activeElement.nodeName &&
-      (document.activeElement.nodeName.toLowerCase() === 'input' ||
-        document.activeElement.nodeName.toLowerCase() === 'textarea')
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thank you!

PR #8021, opened 2025-06-16
feat(core): add `alignSlides` option to keep minimum full slides visi…

mofman:
Fixes: https://github.com/nolimits4web/swiper/discussions/4780

Currently it's possible for the swiper to fall out of alignment if a fractional value is set for `slidesPerView` and the user reaches the end of the swiper (as show below). The PR Introduces a new boolean param `alignSlides` (default `false`) in core defaults and when set to `true` ensures all sides are always aligned in their original place.

Screenshot 2025-06-16 at 01 02 18

- Introduce new boolean param `alignSlides` (default `false`) in core defaults
- Trim `snapGrid` so that only snaps leaving ≥ floor(slidesPerView) slides fully visible remain
- Preserve original “right-flush” fallback only when `alignSlides` is disabled

### DIFF
--- a/src/core/defaults.mjs
+++ b/src/core/defaults.mjs
@@ -58,6 +58,7 @@ export default {
   slidesOffsetAfter: 0, // in px
   normalizeSlideIndex: true,
   centerInsufficientSlides: false,
+  alignSlides: false,
 
   // Disable swiper and hide navigation when container not overflow
   watchOverflow: true,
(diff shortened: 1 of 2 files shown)


mofman:
Any chance of getting this merged reviewed/merged soon?

nolimits4web:
Hi, thanks, but was merged kinda manually as a there are still a bunch of use cases + added demo + types

PR #8030, opened 2025-06-26
fix: slideTo function does not work when grid.rows > 1

danielkorte:
The `swiper.slideTo()` function does not respect grid.rows settings greater than one. I first noticed this when using keyboard navigation on a two row grid with three slides per view. Once the tabbing gets to the last item in the first row it fails to navigate to the first focused slide on the second row with `fill: 'row'`.


src/core/slide/slideToClickedSlide.mjs
```diff
@@ -29,7 +31,11 @@ export default function slideToClickedSlide() {
       } else {
         swiper.slideTo(slideToIndex);
       }
-    } else if (slideToIndex > swiper.slides.length - slidesPerView) {
+    } else if (
+      slideToIndex > isGrid
```
danielkorte, line 35: This is comparing an integer to a boolean?

PR #8061, opened 2025-08-08
feat: add param to disable wrapper live region in a11y module

### LINKED ISSUE #8060: A11y module causes screen readers to announce every text change made inside slides

- [x] I confirm

### Reproduction link

https://codesandbox.io/p/sandbox/swiper-a11y-reproduction-sygtg4

### Bug description

When the A11y module is used and enabled, content in slides that has dynamically updated text (such as the current video progress in video players) is always being announced to screen readers. This persists even if the user is not currently on the slide with dynamically updated text. Especially for carousels that have videos, this makes it impossible for those using screen readers to hear the video as the current progress of the video will consistently be read every second

Currently, there is no way to disable the wrapper from being a live region. While enabling autoplay will set the attribute to "off", this does not disable the announcements as "off" only indicates that announcements should occur when focus is inside the element

The reproduction has a basic carousel with videos using the `video.js` package. Here is a video using the reproduction and VoiceOver on mac where the video progress is announced every second (due to how video.js updates the video progress element) https://github.com/user-attachments/assets/5ae7107b-178f-4a9f-b503-496e4c1050c7

### Expected Behavior

Using the A11y module should not result in dynamic text updates always being announced to screen readers

### Actual Behavior

Using the A11y module results in dynamic text updates always being announced to screen readers

### Swiper version

11.2.10

macOS, Windows

### Validations

bwnn:
closes #8060

This adds a `wrapperLiveRegion` option to the A11y module params to enable/disable if the `aria-live` attribute should be added to the swiper-wrapper element. The default is `true` so as to prevent breaking existing behaviour

### DIFF
--- a/src/modules/a11y/a11y.mjs
+++ b/src/modules/a11y/a11y.mjs
@@ -25,6 +25,7 @@ export default function A11y({ swiper, extendParams, on }) {
       slideRole: 'group',
       id: null,
       scrollOnFocus: true,
+      wrapperLiveRegion: true,
     },
   });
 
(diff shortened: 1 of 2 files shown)


nolimits4web:
Merged, thank you!

PR #8143, opened 2026-01-14
fix: prevent blank slide when loop enabled with next navigation

### LINKED ISSUE #8137: next slide is blank when loop is set `true`, slidesPerView is `3` and actual slides size is 4

- [x] I confirm

### Reproduction link

https://codesandbox.io/p/sandbox/exciting-wozniak-mvtc47

### Bug description

When the total size of the slides is smaller than slidesPerView + slidesPerGroup, calling the next method may result in a blank space appearing on the right side—this behavior is by design.

My question is: why is this only designed to happen when sliding to the right? From my observation, sliding to the left in loop mode does not produce a similar issue.

I’ve briefly looked into the source code for this:

```javascript
/loopfix.mjs#L98

// prepend last slides before start
  if (activeColIndexWithShift < loopedSlides) {
    slidesPrepended = Math.max(loopedSlides - activeColIndexWithShift, slidesPerGroup);
    for (let i = 0; i < loopedSlides - activeColIndexWithShift; i += 1) {
      const index = i - Math.floor(i / cols) * cols;
...
```
I noticed that when scrolling to a position where activeColIndexWithShift < loopedSlides, a prepend operation is triggered—this is the root cause of the blank space issue.

It seems this part of the code is designed to reorganize the DOM order when approaching the boundary, in order to prevent blank areas. However, this logic appears more appropriate for the prev (backward) operation; it’s unnecessary for the next (forward) direction.

### Expected Behavior

/loopfix.mjs#L98

// prepend last slides before start
  if (activeColIndexWithShift < loopedSlides && !isNext) {
    slidesPrepended = Math.max(loopedSlides - activeColIndexWithShift, slidesPerGroup);
    for (let i = 0; i < loopedSlides - activeColIndexWithShift; i += 1) {
      const index = i - Math.floor(i / cols) * cols;
...

### Actual Behavior

_No response_

### Swiper version

11.2.10

Windows 11

### Validations

0xsatoshi99:
## Description

Fixes #8137

When loop is enabled with `slidesPerView: 3` and 4 total slides, navigating to the next slide causes a blank space to appear.

## Root Cause

The prepend operation in `loopFix.mjs` was triggered regardless of navigation direction. This caused slides to be prepended even when moving forward (`isNext`), which is unnecessary and creates blank spaces.

## Solution

Added `!isNext` condition to line 100 to ensure prepend only occurs when moving backward (`isPrev`). This prevents the blank slide issue while maintaining proper loop behavior for backward navigation.

## Changes

- Modified `src/core/loop/loopFix.mjs` line 100
- Changed: `if (activeColIndexWithShift < loopedSlides)`
- To: `if (activeColIndexWithShift < loopedSlides && !isNext)`

## Testing

Tested with the reproduction case from #8137:
- `loop: true`
- `slidesPerView: 3`
- 4 total slides

Result: No blank slide appears when navigating forward.

---

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=162055292

### DIFF
--- a/src/core/loop/loopFix.mjs
+++ b/src/core/loop/loopFix.mjs
@@ -97,7 +97,7 @@ export default function loopFix({
     activeColIndex +
     (bothDirections && typeof setTranslate === 'undefined' ? -slidesPerView / 2 + 0.5 : 0);
   // prepend last slides before start
-  if (activeColIndexWithShift < loopedSlides) {
+  if (activeColIndexWithShift < loopedSlides && !isNext) {
     slidesPrepended = Math.max(loopedSlides - activeColIndexWithShift, slidesPerGroup);
     for (let i = 0; i < loopedSlides - activeColIndexWithShift; i += 1) {
       const index = i - Math.floor(i / cols) * cols;


0xsatoshi99:
@nolimits4web Can you please check this PR and give me feedback if you have any?

PR #8144, opened 2026-01-14
fix(types): fix Vue build errors by updating exports map and relative imports (#8128)

### LINKED ISSUE #8128: "swiper/types" has build errors

- [x] I confirm

### Reproduction link

i dont have one

### Bug description

when i import something from "swiper/types" it throws err on build
env : Nuxt 4 , Vue 3 , TS 
swiper v : 12.0.3
this is the line 
export { default as Swiper } from './swiper-class.d.ts';

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Swiper version

12.0.3

Firefox

### Validations

rami0617:
### Summary
This PR addresses the TypeScript build errors encountered in Vue environments (reported in #8128). The issue was primarily caused by the missing types field in the exports map and absolute-like module resolution conflicts in the Vue type definitions.

**Key Changes:**
src/copy/package.json: Added explicit types declarations for ./types and ./vue endpoints in the exports map. This ensures that modern TypeScript moduleResolution (like bundler or node16) can correctly locate the type definition files.

swiper-vue.d.ts: Updated internal type imports ( SwiperOptions, SwiperClass) to use relative paths (../types/index) instead of bare specifiers. This prevents potential circular or failed resolutions when the package is consumed as a dependency.

**Related Issue**
[Fixes #8128
](https://github.com/nolimits4web/swiper/issues/8128)

**How to test**
Created a fresh Vue + TypeScript project using Vite.
Linked the modified Swiper package using the file protocol.
Verified that the project builds successfully without the "Cannot find module 'swiper/types'" error.
Confirmed that IDE (VS Code) correctly provides IntelliSense for Swiper components and options in .vue files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/nolimits4web/Swiper.git"
+    "url": "git+https://github.com/nolimits4web/Swiper.git"
   },
   "keywords": [
     "swiper",
--- a/src/copy/package.json
+++ b/src/copy/package.json
@@ -90,14 +90,17 @@
       "default": "./swiper-react.mjs"
     },
     "./vue": {
-      "types": "./swiper-vue.d.ts",
-      "default": "./swiper-vue.mjs"
+      "types": "./vue/swiper-vue.d.ts",
+      "import": "./vue/swiper-vue.mjs",
+      "default": "./vue/swiper-vue.mjs"
     },
     "./modules": {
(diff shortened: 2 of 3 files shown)


rami0617:
Thank you for the detailed feedback. 
I clearly see where I missed the mark regarding the build structure and the pathing for the Vue integration. I’d like to correct these points and update the PR to align with the actual build output. Could I go ahead and push the fix to this PR?

PR #8141, opened 2026-01-12
fix(thumbs): Update slide classes in `virtualUpdate` event

nWacky:
Swiper has virtual and thumbnail modules.

When using both modules:

1. slide 1 is highlighted
2. scroll far to the left.
   virtual module removes slide 1 dom node
3. scroll back
   virtual module recreates slide 1 dom node

Currently: slide 1 stops being highlighted as the selected slide.

Expected: slide 1 is highlighted as the selected slide after scrolling.

Also, clicking "Next slide" does not highlight the next slide as current.
Expected for it to be highlighted.

https://github.com/user-attachments/assets/2c646b7c-cb22-45e5-a642-5136880e29ca

<details>

<summary>Test html page</summary>

Copy to `playground/core/index.html`

```html<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <title>Swiper Playground</title>
    <meta name="viewport" content="width=device-width" />
    <style>
      html,
      body {
        position: relative;
        height: 100%;
      }

      body {
        background: #eee;
        font-family: Helvetica Neue, Helvetica, Arial, sans-serif;
        font-size: 16px;
        color: #000;
        margin: 0;
        padding: 0;

        box-sizing: border-box;
        display: flex;
        flex-flow: column;
        padding: 1rem;
        height: 100dvh;
        gap: 2rem;
      }

      .swiper {
        width: 100%;
        height: 100%;
      }

      .swiper.thumbs {
        box-sizing: border-box;
        padding: 0 calc(var(--swiper-navigation-size) + 1rem);
        height: 30%;
      }

      .image-slide {
        display: flex !important;
        align-items: center;
        justify-content: center;
        background: #fff;
        text-align: center;
        font-size: 18px;
      }

      .thumb-slide.swiper-slide {
        display: flex;
        align-items: center;
        justify-content: center;
        background: #d3d3d3;
      }

      .thumb-slide.swiper-slide-thumb-active {
        background: #76ff8d;
      }
    </style>
  </head>

  <body>
    <div class="images swiper">
      <div class="swiper-wrapper">
        <!-- <div class="swiper-slide image-slide">Slide 1</div> -->
      </div>

      <div class="swiper-button-next"></div>
      <div class="swiper-button-prev"></div>
    </div>

    <div class="thumbs swiper">
      <div class="swiper-wrapper">
        <!-- <div class="swiper-slide thumb-slide">Thumbnail 1</div>  -->
      </div>
[...]

PR #8146, opened 2026-01-15
fix(types): Add `autoScroll` to `thumbs.update` type signature

nWacky:
After #8141 `update()` supports 2 parameters:

https://github.com/nolimits4web/swiper/blob/975277111b73f389043cb0ed19feee0244a80f57/src/modules/thumbs/thumbs.mjs#L96

`update()` is exposed publicly in `swiper.thumbs.update`.

**In this PR:**

Add `p: { autoScroll: boolean }` to `update()`'s type signature.

When calling `update(false, { otherOption: true })`, default `autoScroll` to `true`

### DIFF
--- a/src/modules/thumbs/thumbs.mjs
+++ b/src/modules/thumbs/thumbs.mjs
@@ -93,7 +93,7 @@ export default function Thumb({ swiper, extendParams, on }) {
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thanks!

PR #8186, opened 2026-05-06
fix: avoid duplicate css type resolutions

Stanzilla:
## Summary
- point each CSS export at a declaration file matching its emitted CSS file
- generate side-effect-only CSS declaration files during the type build
- preserve shared declarations only for true CSS aliases

## Problem

Currently all imports are duplicated via side effects on the consumer side, causing lint issues

## Verification
- npm run check-format -- scripts/build-types.js src/copy/package.json
- npm run lint
- npm run build:prod
- reproduced and verified import-x/no-duplicates with eslint-plugin-import-x + eslint-import-resolver-typescript

### DIFF
--- a/scripts/build-types.js
+++ b/scripts/build-types.js
@@ -130,5 +130,27 @@ export default async function buildTypes() {
       return fs.writeFile(destPath, fileContent);
     }),
   );
+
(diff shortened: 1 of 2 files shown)


nolimits4web:
Merged, thank you!

PR #8201, opened 2026-07-03
fix(thumbs): bind Thumbs tap handler when thumbs swiper is set after init

### LINKED ISSUE #8198: v14: clicking a thumbnail no longer navigates the main Swiper (Thumbs tap handler not attached)

- [x] I confirm

### Reproduction link

https://stackblitz.com/edit/vitejs-vite-ddcn4c8w?file=package.json,src%2FApp.jsx&terminal=dev

### Bug description

With `swiper/react`, clicking a thumbnail no longer navigates the main Swiper in v14. This uses the standard pattern where the thumbs Swiper is stored in state (`useState` + `onSwiper`) and passed to the main Swiper via `thumbs={{ swiper }}`, so the thumbs instance only becomes available _after_ the main Swiper has initialized.

Thumbnails still render, but clicking one does nothing, and the active-thumbnail highlight is never applied. Arrow navigation and programmatic `slideTo` / `slideNext` keep working. This worked in v12.2.0 and broke in v14.0.0; it is still present in v14.0.1.

### Root cause

Because the thumbs instance is supplied after init, `beforeInit` runs while `thumbs.swiper` is still `null`, so the Thumbs module is initialized later through `updateSwiper()` → `swiper.thumbs.init()` when the `thumbs` param changes.

In v14, [`src/components-shared/update-swiper.ts`](https://github.com/nolimits4web/swiper/blob/v14.0.1/src/components-shared/update-swiper.ts) gates this late initialization with `isObject(passedThumbs.swiper)`. That helper, from [`src/components-shared/utils.ts`](https://github.com/nolimits4web/swiper/blob/v14.0.1/src/components-shared/utils.ts), explicitly excludes Swiper instances via the `__swiper__` marker:

```js
// isObject(...) — trailing condition
&& !o.__swiper__
```

A Swiper instance is created with `__swiper__ = true` (`swiper.__swiper__ = true` in the core), so a valid thumbs Swiper instance fails this check and `isObject(passedThumbs.swiper)` is always `false`. As a result `needThumbsInit` is never set and `swiper.thumbs.init()` is never called, so the `tap` handler registered by the [Thumbs module](https://github.com/nolimits4web/swiper/blob/v14.0.1/src/modules/thumbs/thumbs.ts) (which calls `slideTo` / `slideToLoop` on the main Swiper) is never attached. Since `swiper.thumbs.swiper` also stays unset, the module's `update()` early-returns, which is why the active-thumbnail highlight is not applied either. Programmatic navigation is independent of Thumbs, so it keeps working.

In [v12.2.0](https://github.com/nolimits4web/swiper/blob/v12.2.0/src/components-shared/update-swiper.mjs) the same gate used a truthiness check on the instance (`passedParams.thumbs.swiper`), which a Swiper instance satisfies, so late initialization worked.

### Suggested fix

`thumbs.swiper` holds a Swiper instance (intentionally marked `__swiper__`), so it should not be validated with `isObject`, which is designed to reject Swiper instances. Gating on `passedThumbs.swiper` being truthy and not destroyed — as the adjacent `controller` branch already does for `passedController.control` — restores the v12 behavior. The `controller` wiring in the same function is unaffected precisely because it checks the instance for truthiness rather than with `isObject`.

https://github.com/nolimits4web/swiper/blob/master/src/components-shared/update-swiper.ts#L52
```
[before]:
isObject(passedThumbs.swiper) &&
↓
[after]
passedThumbs.swiper &&
```

I verified this change locally and it resolves the bug.

### Expected Behavior

Clicking a thumbnail navigates the main Swiper to the corresponding slide, including when the thumbs instance is provided after initialization (the documented React `useState` + `onSwiper` pattern).

### Actual Behavior

Clicking a thumbnail has no effect and the active-thumbnail highlight is never applied. Arrow and programmatic navigation still work.

### Swiper version

14.0.1 (also reproduced on 14.0.0). Last working version: 12.2.0.

Web, desktop. Reproduced in Chrome and Safari.

### Validations

xlgorbylx:
## What

Fixes #8198.

With `swiper/react`, clicking a thumbnail does not navigate the main Swiper in v14 when the thumbs Swiper is provided **after** the main Swiper has initialized (the standard `useState` + `onSwiper` pattern). Thumbnails render, but clicking one does nothing and the active-thumbnail highlight is never applied. Arrow navigation and programmatic `slideTo` / `slideNext` keep working. This worked in v12.2.0 and regressed in v14.0.0 (still present in v14.0.1).

## Root cause

Because the thumbs instance is supplied after init, `beforeInit` runs while `thumbs.swiper` is still `null`, so `swiper.thumbs.init()` has to run later from `updateSwiper()` when the `thumbs` param changes. In `src/components-shared/update-swiper.ts`, that late-init gate validates the instance with `isObject(passedThumbs.swiper)`:

```ts
isObject(passedThumbs) &&
isObject(passedThumbs.swiper) &&   // <-- instance checked with isObject
!(passedThumbs.swiper as Swiper).destroyed &&
```

But `isObject` (`src/components-shared/utils.ts`) intentionally **rejects Swiper instances** via the `__swiper__` marker (it ends with `&& !obj.__swiper__`), and instances are created with `__swiper__ = true`. So `isObject(passedThumbs.swiper)` is always `false`: `needThumbsInit` is never set, `swiper.thumbs.init()` never runs, and the thumbnail `tap` handler is never attached. Since `swiper.thumbs.swiper` also stays unset, the module's `update()` early-returns, which is why the active-thumbnail highlight is not applied either.

The adjacent `controller` branch validates its instance with a plain truthiness check (`passedController.control`), which is why controllers are unaffected.

In v12.2.0 this thumbs gate also used a truthiness check on the instance, so late initialization worked.

## Change

Validate `passedThumbs.swiper` with a truthiness check instead of `isObject`, matching the `controller` branch and restoring the v12.2.0 behavior:

```diff
     changedParams.includes('thumbs') &&
     isObject(passedThumbs) &&
-    isObject(passedThumbs.swiper) &&
+    passedThumbs.swiper &&
     !(passedThumbs.swiper as Swiper).destroyed &&
```

## Verification

- Applied this change locally and confirmed the thumbnail click navigates the main Swiper again; reverting the line reproduces the bug.

This only relaxes the gate to accept a valid, non-destroyed Swiper instance (restoring prior behavior). The change is in the shared `update-swiper` used by all framework wrappers; I verified the React wrapper but have not tested Vue / Svelte / Solid — please double-check those.

### DIFF
--- a/src/components-shared/update-swiper.ts
+++ b/src/components-shared/update-swiper.ts
@@ -49,7 +49,7 @@ export function updateSwiper(args: UpdateSwiperArgs): void {
   if (
     changedParams.includes('thumbs') &&
     isObject(passedThumbs) &&
-    isObject(passedThumbs.swiper) &&
+    passedThumbs.swiper &&
     !(passedThumbs.swiper as Swiper).destroyed &&
     isObject(currentThumbs) &&
     (!currentThumbs.swiper || (currentThumbs.swiper as Swiper).destroyed)


nolimits4web:
Merged, thank you!

PR #8204, opened 2026-07-18
fix(loop): stop false "not enough slides" warning with auto + centeredSlides

### LINKED ISSUE #7586: Swiper incorrectly warns about a wrong number of slides for loop mode

- [X] I confirm

### Reproduction link

https://codesandbox.io/p/devbox/swiper-react-infinite-loop-vfz433?file=%2Fsrc%2FApp.jsx

### Bug description

If you attempt to render a basic loop swiper with any number of slides within jest tests the swiper warns about the incorrect number of slides when it shouldn't

The Swiper
```tsx
import { Autoplay, Navigation, Pagination } from 'swiper/modules';
import { Swiper, SwiperSlide } from 'swiper/react';

const modules = [Navigation, Pagination, Autoplay];

export const App = () => {
  return (
    <Swiper
      loop
      slidesPerView={1}
      slidesPerGroup={1}
      loopAddBlankSlides
      modules={modules}
    >
      <SwiperSlide>test</SwiperSlide>
      <SwiperSlide>test</SwiperSlide>
      <SwiperSlide>test</SwiperSlide>
      <SwiperSlide>test</SwiperSlide>
      <SwiperSlide>test</SwiperSlide>
      <SwiperSlide>test</SwiperSlide>
    </Swiper>
  );
};
```

the test

```tsx
import { render } from '@testing-library/react';

import { App } from './abc';

describe('App', () => {
  it('App', () => {
    render(<App />);
  });
});

```

> [!CAUTION]
> Swiper Loop Warning: The number of slides is not enough for loop mode, it will be disabled and not function properly. You need to add more slides (or make duplicates) or lower the values of slidesPerView and slidesPerGroup parameters

### Expected Behavior

No warning

### Actual Behavior

Warning

### Swiper version

11.1.4

macOS 14.5

### Validations

kirthi-b:
slidesPerViewDynamic() is used to work out how many slides are visible when slidesPerView is 'auto', and loopFix() uses that number to decide whether there are enough slides to run loop mode. When centeredSlides is also on, the dynamic calculation sums slide sizes outward from the active slide until the running total passes the container size (swiper.size).

The container size is only set once updateSize() measures a non-zero clientWidth/clientHeight. If the container is hidden, not yet laid out, or running in a DOM environment with no real layout, swiper.size stays undefined. Comparing any slide size against undefined is always false, so the centered branch never stops and walks every slide, returning a count close to the total slide count instead of what actually fits. loopFix then sees that inflated count and reports "not enough slides for loop mode" even with plenty of slides. This matches what people in the thread found: the warning shows up specifically with slidesPerView 'auto' plus centeredSlides, and also through breakpoints that switch a swiper into that combination, since breakpoints go through the same function.

The fix makes slidesPerViewDynamic() fall back to 1 whenever the container size is unknown, which is the same result the non-auto branches already produce implicitly in that situation. No behavior change when the size is actually known.

Added tests/element-dom/loop-warning.test.mjs, a real DOM regression test using happy-dom (it never runs real layout, so it reproduces the unmeasured container condition on every run instead of needing a timing-dependent setup). It covers auto alone, centeredSlides alone, both together, a breakpoints switch into that combination, a control case confirming a genuine slide shortage still warns, and a slideToLoop check. Wired into npm test as loop-warning-test.

Ran against master before this change to confirm the two auto plus centeredSlides cases fail with the real warning text, then confirmed all six pass after the fix. Also ran type-check, build, and the existing contract, element-dom, ssr, and consumer-types suites with no regressions.

I did not touch the slideToLoop index bug some commenters described in the thread. I could not reproduce it against current master, with or without this fix, so I left it alone instead of guessing at a change for something I couldn't observe. I also did not add a way to suppress the warning directly. The calculation fix removes the false positive for the reported cases, and there is no existing debug/logger flag to attach a suppression option to.

Fixes #7586

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,8 +24,9 @@
     "consumer-types-test": "node tests/consumer-types/run.mjs",
     "ssr-test": "node tests/ssr/ssr.test.mjs",
     "element-dom-test": "node tests/element-dom/element-dom.test.mjs",
+    "loop-warning-test": "node tests/element-dom/loop-warning.test.mjs",
     "release": "npm run validate && node ./scripts/release",
-    "test": "npm run validate && npm run build:prod && npm run contract-test && npm run dist-types-test && npm run consumer-types-test && npm run ssr-test && npm run element-dom-test && npm run bundle-size",
+    "test": "npm run validate && npm run build:prod && npm run contract-test && npm run dist-types-test && npm run consumer-types-test && npm run ssr-test && npm run element-dom-test && npm run loop-warning-test && npm run bundle-size",
     "changelog": "npx conventional-changelog -p angular -i CHANGELOG.md -u -s",
     "build-sponsors": "node scripts/build-sponsors.js"
   },
--- a/src/core/core.ts
+++ b/src/core/core.ts
@@ -994,6 +994,16 @@ export class Swiper {
     let spv = 1;
     if (typeof params.slidesPerView === 'number') return params.slidesPerView;
 
+    // swiperSize is only known once updateSize() has measured a non-zero container (see
+    // updateSize.ts, which bails out and leaves `size` unset when clientWidth/clientHeight is 0 -
+    // e.g. the container is hidden or not yet laid out). Without this guard the centered branch
+    // below never finds a slide size that "overflows" an unknown size, so it keeps counting past
+    // the last slide and reports spv ~= slides.length. That inflated count then made loopFix()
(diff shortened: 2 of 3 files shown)


nolimits4web:
Merged into master as ec9a1709 (rebased, with the code comment in core.ts trimmed to a one-liner — your commit authorship is preserved). Will be released in the next patch. Thank you!

PR #8210, opened 2026-08-03
fix(core): make elementOffset viewport-relative again

### LINKED ISSUE #8209: Keyboard: onlyInViewport is broken for any slider below the fold in v14 (elementOffset changed to document-absolute, guard still compares against innerHeight)
### Reproduction

Self-contained — save as `repro.html` and open it. No build step, no dependencies beyond the CDN tags. (Inlined rather than linked because the whole thing is 30 lines and the root cause is pinned to a specific line below.)

```html
<!doctype html>
<html>
<head>
<meta charset="utf-8">
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/swiper@14.0.7/swiper-bundle.min.css">
<style>
  .spacer { height: 150vh; padding: 20px; font: 16px sans-serif; }
  .swiper { width: 600px; }
  .swiper-slide { height: 200px; display: flex; align-items: center;
                  justify-content: center; font: 700 48px sans-serif; color: #fff; }
  .a { background: #3a7bd5 } .b { background: #7048b6 } .c { background: #c1436d }
  #log { font: 14px monospace; padding: 12px 20px; }
</style>
</head>
<body>

<div class="spacer">Scroll down to the slider, click a slide, then press ArrowRight.</div>

<div class="swiper">
  <div class="swiper-wrapper">
    <div class="swiper-slide a" tabindex="0">1</div>
    <div class="swiper-slide b" tabindex="0">2</div>
    <div class="swiper-slide c" tabindex="0">3</div>
  </div>
</div>

<div id="log"></div>

<script src="https://cdn.jsdelivr.net/npm/swiper@14.0.7/swiper-bundle.min.js"></script>
<script>
  var swiper = new Swiper('.swiper', {
    keyboard: { enabled: true, onlyInViewport: true },
  });

  function report() {
    var box = swiper.el.getBoundingClientRect();
    var offsetTop = box.top + window.scrollY;   // what elementOffset() returns in v14
    document.getElementById('log').textContent =
      'activeIndex ' + swiper.activeIndex +
      ' | elementOffset().top ' + Math.round(offsetTop) +
      ' | window.innerHeight ' + window.innerHeight +
      ' | inView ' + (offsetTop >= 0 && offsetTop <= window.innerHeight);
  }
  report();
  addEventListener('scroll', report, { passive: true });
  swiper.on('slideChange', report);
</script>
</body>
</html>
```

Scroll down to the slider, focus a slide, press <kbd>ArrowRight</kbd>. Nothing happens.

Change the two CDN URLs from `14.0.7` to `12.2.0` and the same file works. That is the only edit needed to flip it.

I ran exactly this file headlessly against both versions at a 1280×720 viewport:

```
swiper 12.2.0: activeIndex 0 -> 1   MOVED
swiper 14.0.7: activeIndex 0 -> 0   STUCK
   elementOffset().top 1128 | window.innerHeight 720
```

### Bug description

With `keyboard.onlyInViewport: true` (the default), arrow keys stop working for **any slider positioned more than roughly one viewport height down the page**, regardless of where the page is actually scrolled to. A slider near the top of the document still works, which makes this easy to miss in a short demo page.

The cause is a semantic change to `elementOffset()` in the v14 rewrite that its callers were not updated for.

**`src/shared/utils.ts`** now returns a **document-absolute** top:

```ts
export function elementOffset(el: Element): { top: number; left: number } {
  const box = el.getBoundingClientRect();
  return {
    top: box.top + window.scrollY - (el.clientTop || 0),
    left: box.left + window.scrollX - (el.clientLeft || 0),
  };
}
```

In v12 it returned a **viewport-relative** top — the scroll term was `el.scrollTop`, which is `0` for a normal element:

```js
const scrollTop = el === window ? window.scrollY : el.scrollTop;
return { top: box.top + scrollTop - clientTop, ... };
```

**`src/modules/keyboard/keyboard.ts`** still compares that value against the viewport, unchanged:

```ts
const windowHeight = window.innerHeight;
const swiperOffset = elementOffset(el);
const swiperCoord: Array<[number, number]> = [
  [swiperOffset.left, swiperOffset.top],
  ...
];
for (...) {
  if (point[0] >= 0 && point[0] <= windowWidth && point[1] >= 0 && point[1] <= windowHeight) {
    inView = true;
  }
}
if (!inView) return undefined;
```

So a document-absolute coordinate is tested against `window.innerHeight`. Once a slider sits below one screenful, that comparison can never succeed again no matter how far the user scrolls, `inView` stays `false`, and the handler returns early.

I believe this was unintentional rather than a deliberate API change: the v14.0.0 changelog lists only the browser baseline, the Node build requirement and stricter types, and states of the last that "there are no runtime behavior changes". `onlyInViewport` is still documented as "will control sliders that are currently in viewport", default `true`. `PLAN_V14.md` item 7 lists `elementOffset` → `getBoundingClientRect()` among a batch of DOM-helper simplifications, which reads like the frame-of-reference change was not noticed.

### Two other callers look affected by the same change

I have not runtime-tested these — they are source reads — but both assume the old viewport-relative semantics:

**`src/modules/zoom/zoom.ts`** adds the scroll offset on top of a value that now already includes it, double-counting it once the page is scrolled:

```ts
offsetX = elementOffset(gesture.slideEl).left + window.scrollX;
offsetY = elementOffset(gesture.slideEl).top + window.scrollY;
```

These two lines are byte-identical between v12 and v14 — only the helper beneath them changed. In v12 they composed correctly to document-absolute.

**`src/modules/scrollbar/scrollbar.ts`** subtracts a now-absolute offset from a viewport-relative pointer coordinate (`getPointerPosition` returns `clientX`/`clientY`):

```ts
positionRatio =
  (getPointerPosition(e) -
    elementOffset(el)[swiper.isHorizontal() ? 'left' : 'top'] - ...
```

Horizontal sliders likely escape this in practice since `window.scrollX` is usually `0`; vertical ones would not.

### Expected Behavior

A slider visible in the viewport responds to arrow keys, wherever it sits in the document — as it did in v12 and as `onlyInViewport` is documented to behave.

### Actual Behavior

Any slider more than about one viewport height down the document ignores arrow keys permanently, including while fully visible on screen.

### Swiper version

14.0.7. Also present in 14.0.6, and unchanged on `master` at the time of writing — I checked both `src/shared/utils.ts` and `src/modules/keyboard/keyboard.ts`. 12.2.0 is unaffected.

macOS 15 (Darwin 25.5.0), Chromium 1280×720 via Playwright. Not browser-specific — it is arithmetic, and reproduces anywhere the page can scroll.

### Possible fix

Either compare in the same frame of reference the guard assumes:

```ts
const rect = el.getBoundingClientRect();   // viewport-relative
```

or subtract the scroll offset back out at the comparison. The other two call sites want the document-absolute value, so changing `elementOffset` itself back would need those updated in the opposite direction — auditing the three callers individually is probably the safer route.

Happy to open a PR if you would like one.

jmwerk:
Fixes #8209 (SWIPER-178).

The v14 rewrite changed `elementOffset` from returning a viewport-relative top/left to a document-absolute one, by folding `window.scrollY`/`scrollX` into it. None of its three callers were updated to match, and all three want the viewport-relative value:

| Caller | Expression | What it needs |
| --- | --- | --- |
| `modules/keyboard` | compares against `window.innerHeight` | viewport-relative |
| `modules/scrollbar` | `getPointerPosition(e) - elementOffset(el)[...]`, where the pointer position is `clientX`/`clientY` | viewport-relative |
| `modules/zoom` | `elementOffset(slideEl).top + window.scrollY`, matched against `e.pageY` | viewport-relative — it adds the scroll offset itself |

The user-visible symptom is the keyboard one: with `onlyInViewport: true` (the default), a document-absolute coordinate is tested against `window.innerHeight`, so any slider sitting more than roughly one viewport height down the page fails that test permanently and stops answering arrow keys — no matter how far the page is scrolled, or that the slider is plainly on screen. A slider near the top of the document still works, which is what makes it easy to miss.

Zoom is the subtler one. `offsetY = elementOffset(gesture.slideEl).top + window.scrollY` is byte-identical between v12 and v14; only the helper beneath it changed. In v12 that composed correctly to document-absolute to match `pageY`; in v14 it counts the scroll offset twice.

Restoring the original semantics fixes all three at once and leaves every call site untouched — this changes one function and nothing else.

`PLAN_V14.md` item 7 lists this helper among the DOM helpers to simplify down to `getBoundingClientRect()`, which is now exactly what it is. The scroll term looks like the accidental part of that change.

### Verification

**Behaviour.** I patched this function in the distributed 14.0.7 bundle and drove a below-the-fold slider at three scroll positions in Chromium at 1280×720:

```
scrollY=608   activeIndex 0 -> 1  MOVED    elementOffset().top 520 | innerHeight 720
scrollY=656   activeIndex 0 -> 1  MOVED    elementOffset().top 472 | innerHeight 720
scrollY=656   activeIndex 0 -> 1  MOVED    elementOffset().top 472 | innerHeight 720

control, slider genuinely off screen: 0 -> 0  correctly IGNORED
```

The control is the important half: `onlyInViewport` still suppresses the keypress when the slider really is off screen, so this restores the guard rather than defeating it. On unpatched 14.0.7 the same page is `STUCK` at every one of those positions; on 12.2.0 it moves.

**Against a real consumer.** I maintain a Swiper-based WordPress slider plugin with a Playwright suite that covers keyboard navigation specifically, including a sweep that drives every keyboard-enabled slider on a tall fixture page at real scroll offsets and names any that fail to move. That suite is what caught this in the first place — on unpatched 14.0.7 it reported **28 of 29 sliders deaf to arrow keys**, and 5 specs failing.

I swapped this branch's build into that plugin in place of 12.2.0 and reran it:

| Spec | 14.0.7 | this branch |
| --- | --- | --- |
| arrow keys on a slider below the fold | fail | pass |
| every keyboard-navigable slider answers an arrow key (29 sliders) | fail | pass |
| arrow key with focus inside a nested slider moves only it | fail | pass |
| outer slider answers arrow keys again once focus leaves | fail | pass |
| arrow key with focus outside any slider moves the outer one | fail | pass |

The **full suite is at parity with 12.2.0 — 103 passed**, covering RTL arrow direction, nested focus hand-off, responsive rebuilds, teardown and peek/overflow. So nothing else in the v14 rewrite regresses for that consumer once this is fixed.

Incidentally it confirms the v14 bundle-size goal holds: that plugin's frontend bundle drops from 34.85 kB to 34.19 kB gzipped on 14 versus 12.

**Suite.** `npm test` passes end to end on this branch — `check-format`, `tsc --noEmit`, the contract/dist-types/consumer-types/SSR/element-DOM/loop-warning suites (31 checks, 0 failures), exit 0.

**Size.** 30 bytes smaller minified (`shared/utils.min.mjs` 4195 → 4165 B, `swiper-bundle.min.js` 151615 → 151585 B). Note that `npm run bundle-size` reports large deltas on this branch for unrelated files — the committed baseline predates v14 (it still lists `shared/ssr-window.esm.min.mjs` as present), so those numbers are not from this change. I measured the delta by building the branch with and without the patch.

### Scope

I exercised the keyboard path directly. Zoom and scrollbar are reasoned from the source — the arithmetic is unambiguous and they were written against the old semantics, but I have not driven either, and it would be worth a maintainer eye on them before this lands.

Happy to add a regression test if you want one, though I could not find an existing home for a behavioural test of module logic — `tests/` is contract, types, SSR and element-DOM rather than a jsdom-style unit suite. A below-the-fold keyboard case is precisely the gap that let this ship, so if you would like me to stand one up, say where you would want it to live.

### DIFF
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -160,16 +160,8 @@ export function createElement(tag: string, classes: string | string[] = []): HTM
   return el;
 }
 
-/**
- * Offset of an element relative to the viewport.
- *
- * Viewport-relative, not document-absolute: every caller either compares the
- * result against viewport quantities (`window.innerHeight` in the keyboard
- * module, `clientX`/`clientY` in the scrollbar) or adds the scroll offset
- * itself (the zoom module, which matches it against `pageX`/`pageY`). Adding
- * `window.scrollX`/`scrollY` here breaks the first two and double-counts for
- * the third.
- */
+// Viewport-relative on purpose: every caller either compares against viewport quantities
+// or adds window.scrollX/Y itself, so folding the scroll offset in here double-counts.
 export function elementOffset(el: Element): { top: number; left: number } {
   const box = el.getBoundingClientRect();
   return {
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -160,11 +160,21 @@ export function createElement(tag: string, classes: string | string[] = []): HTM
   return el;
 }
 
+/**
+ * Offset of an element relative to the viewport.
+ *
+ * Viewport-relative, not document-absolute: every caller either compares the
+ * result against viewport quantities (`window.innerHeight` in the keyboard
+ * module, `clientX`/`clientY` in the scrollbar) or adds the scroll offset
+ * itself (the zoom module, which matches it against `pageX`/`pageY`). Adding
+ * `window.scrollX`/`scrollY` here breaks the first two and double-counts for
+ * the third.
+ */
 export function elementOffset(el: Element): { top: number; left: number } {
   const box = el.getBoundingClientRect();
   return {
-    top: box.top + window.scrollY - (el.clientTop || 0),
-    left: box.left + window.scrollX - (el.clientLeft || 0),
+    top: box.top - (el.clientTop || 0),
+    left: box.left - (el.clientLeft || 0),
   };
 }
 


nolimits4web:
Merged, thank you!

PR #8212, opened 2026-08-10
fix(core): recompute rtlTranslate in changeDirection (#8053)

### LINKED ISSUE #8053: slidePrev() slides to the very first slide instead of the previous one in certain circumstances

- [x] I confirm

### Reproduction link

https://codesandbox.io/p/sandbox/objective-platform-7tw8mt

### Bug description

After changing language direction to "rtl" then direction to "vertical", calling slidePrev() will goes to the very first slide, instead of the previous one.

### Reproduction

1. set initial direction to horizontal and language direction to ltr;
2. click button 'lang direction rtl' to set language direction to rtl;
3. click button 'direction vertical' to set direction to vertical;
4. click button 'next' several times to go to the last slide;
5. click button 'prev', it goes to the very first slide.

### Expected Behavior

In the circumstance described in the above steps, click button 'prev' (slidePrev()) should go to the previous slide, instead of the first slide.

This unexpected behavior is observed only when first changing language direction to rtl followed by changing direction to vertical.

### Actual Behavior

Calling slidePrev() goes to the very first slide.

### Swiper version

11.2.10

macOS Safari 18.5 Firefox 141.0 Chromium 138.0.7204.157

### Validations

Jaybhade:
Fixes #8053.

### The bug

`swiper.rtlTranslate` is the flag that decides the sign of every translate read:

```js
const translate = rtlTranslate ? swiper.translate : -swiper.translate;
```

It is computed in exactly two places — `mount()` and `changeLanguageDirection()` — both as
`params.direction === 'horizontal' && rtl`. `changeDirection()` sets `params.direction` but
never touches it, so the flag goes stale as soon as the axis changes:

1. horizontal + `dir="rtl"` → `rtlTranslate: true` (correct)
2. `changeDirection('vertical')` → `rtlTranslate` stays `true` (wrong — a vertical slider is
   never translated RTL)

From then on `slidePrev()` normalizes `+translate` against a snap grid built from
`-translate`, finds no match, and falls through to `prevIndex = 0`, which is the reported
symptom: prev jumps to the first slide. `slideToClosest()`, `updateActiveIndex()`,
`updateProgress()`, `onTouchMove`/`onTouchEnd` and `setTranslate()` read the same flag, so
dragging and progress are off in that state too.

The mirror case is broken as well: an RTL slider that starts vertical and is switched to
horizontal keeps `rtlTranslate: false` and behaves as if it were LTR.

### The fix

Recompute the flag next to `params.direction`, with the same expression
`changeLanguageDirection()` already uses.

### Verification

`npm test` is green (format, type-check, build, contract, dist-types, consumer-types, SSR,
element-dom, loop-warning, cross-realm).

The regression test added to `tests/element-dom/element-dom.test.mjs` asserts the flag across
both switches and the user-visible symptom (`slidePrev()` steps back one slide instead of
jumping to 0). It fails on `master` and passes with the fix:

```
# without the fix
  FAIL  changeDirection() recomputes rtlTranslate
        vertical is never translated RTL
  7 passed, 1 failed

# with the fix
  ok  changeDirection() recomputes rtlTranslate
  8 passed, 0 failed
```

`width`/`height` params are passed in the test so `swiper.size` is set without real layout,
which happy-dom does not perform; with a numeric `slidesPerView` the slide sizes derive from
that size.

I also ran the same scenario against the published `swiper@12.2.0` bundle to confirm this is
not a v14 regression — it reproduces identically there — and diffed 32 unrelated scenarios
(loop, centered, grid, pagination types, a11y, effects, virtual, manipulation, breakpoints,
free mode, scrollbar…) between `swiper@12.2.0` and this build: no behavioural differences,
so nothing else moves.

One note on `npm run bundle-size`: the checked-in baseline is already stale on a clean
`master` (a control run there reports `TOTAL -16676 B`). This branch reports `-16481 B`, i.e.
+195 B for the added line; I left `scripts/bundle-size-baseline.json` untouched since it looks like it is
refreshed at release time.

### DIFF
--- a/src/core/core.ts
+++ b/src/core/core.ts
@@ -1136,6 +1136,9 @@ export class Swiper {
     swiper.emitContainerClasses();
 
     swiper.params.direction = newDirection;
+    // RTL only inverts the translate axis while horizontal, so recompute the flag here —
+    // a stale `true` flips the sign of every translate read on a vertical slider.
+    swiper.rtlTranslate = newDirection === 'horizontal' && swiper.rtl;
 
     swiper.slides.forEach((slideEl) => {
       if (newDirection === 'vertical') {
--- a/tests/element-dom/element-dom.test.mjs
+++ b/tests/element-dom/element-dom.test.mjs
@@ -231,5 +231,51 @@ await check('core swiper resolves a string scrollbar.el selector', async () => {
   }
 });
 
+// Regression: https://github.com/nolimits4web/swiper/issues/8053 — `rtlTranslate` was only
+// recomputed by mount() and changeLanguageDirection(), never by changeDirection(). An RTL
+// slider switched to vertical therefore kept `rtlTranslate: true`, and every `rtlTranslate
+// ? translate : -translate` read (slidePrev, slideToClosest, updateActiveIndex, ...) got the
+// wrong sign — slidePrev() found no matching snap and fell back to slide 0.
+// `width`/`height` are passed so `swiper.size` is set without real layout, which happy-dom
+// doesn't do; with slidesPerView as a number the slide sizes are derived from it.
+await check('changeDirection() recomputes rtlTranslate', async () => {
+  const host = doc.createElement('div');
+  host.innerHTML = `
+    <div class="swiper">
+      <div class="swiper-wrapper">
+        <div class="swiper-slide">1</div>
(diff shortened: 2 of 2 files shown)


nolimits4web:
Merged, thank you!

Issue #8015, opened 2025-06-04
Pagination flicker with centeredSlides and loop


arnaudbrowet:
### Check that this is really a bug

- [x] I confirm

### Reproduction link

https://codesandbox.io/p/sandbox/swiper-pagination-forked-kkknw9

### Bug description

When using a small number of slides (the bug is less reproducible but still happens with a larger number of slides) the computation of the active slides occurs during sliding  and not after the slide changed. Moreover, the computation of the active slide is wrong and the pagination quickly changes between any elements (which i called flickering)

This is particularly evident when swiping a silde left and right without actually going to neither the next or the previous slide.

The issue is not observed in swiper 11.0 but occurs in 11.1 + 

https://github.com/user-attachments/assets/18246870-6b0c-4c2f-a391-e36dd5191e27

### Expected Behavior

The pagination should be computed after slide transition finished

### Swiper version

11.1 +

### Platform/Target and Browser Versions

chrome

### Validations

- [x] Follow our [Code of Conduct](https://github.com/nolimits4web/swiper/blob/master/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://swiperjs.com/swiper-api).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [x] Make sure this is a Swiper issue and not a framework-specific issue

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

hexanal:
I submitted a Pull Request that tries to address this issue (my first ever, so I might have done things the wrong the way!)

Essentially, I prevent some callbacks from firing too early. That way the pagination state correctly resolves once the slider has transitioned. Obviously, since it's my first foray into altering this source code, it'll need a thorough review process because I might have missed something important.

PR #8213, opened 2026-08-13
fix (core): use runCallbacks in updateActiveIndex

hexanal:
This fix aims to address the issue brought up here: https://github.com/nolimits4web/swiper/issues/8015

When using some combination of props like `centeredSlides`, `loop`, `slidesPerView` set to a small value, and with a small number of slides, the pagination bullets changed their active state in an erratic manner.

What the fix does: 

* Passes the `runCallbacks` prop to `updateActiveIndex`, to conditionally emit the callback events
* Makes `slideTo` use this new parameter on `updateActiveIndex`, set to `false` in order to prevent the pagination module from eagerly changing the index

Videos:

https://github.com/user-attachments/assets/bffce485-ac9b-4ca6-8686-6e591a47607c

https://github.com/user-attachments/assets/3d061738-0a48-4948-b871-bed3a1a0347b

https://github.com/user-attachments/assets/4c7bb085-2146-43df-9250-81a867a701d7

https://github.com/user-attachments/assets/e9888f02-e7b6-46f7-bca7-64c55c4c42ec

This is my first ever pull-request after many years of calling myself a developer. I'm happy to finally contribute!

### DIFF
--- a/src/core/core.ts
+++ b/src/core/core.ts
@@ -579,7 +579,7 @@ export interface Swiper {
    */
   updateSlidesClasses(): void;
 
-  updateActiveIndex(newActiveIndex?: number): void;
+  updateActiveIndex(newActiveIndex?: number, runCallbacks?: boolean): void;
   updateClickedSlide(el: HTMLElement, path?: EventTarget[]): void;
 
   // Breakpoints methods (prototype mixin: src/core/breakpoints)
--- a/src/core/slide/slideTo.ts
+++ b/src/core/slide/slideTo.ts
@@ -109,7 +109,7 @@ export default function slideTo(
     !isInitialVirtual &&
     ((rtl && -translate === swiper.translate) || (!rtl && translate === swiper.translate))
   ) {
-    swiper.updateActiveIndex(slideIndex);
+    swiper.updateActiveIndex(slideIndex, runCallbacks);
     // Update Height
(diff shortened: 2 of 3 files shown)


nolimits4web:
Hi @hexanal , thank you for this - and congrats on your first PR! You nailed the diagnosis: the erratic pagination comes from loopFix()'s compensating slideTo() teleports flowing through updateActiveIndex() and emitting index-change events for transient, wrong indexes. The videos made the problem very clear.

I couldn't merge it as-is, because tying the suppression to the public runCallbacks parameter changes its documented contract (it only silences transition events) and breaks other paths that rely on those events. Pagination bullets are updated via snapIndexChange, so with this patch slideTo(index, speed, false) leaves the bullets out of sync, addSlide()/removeSlide() (which end with a silent slideTo) leave pagination stale, and a drag that crosses a slide boundary can swallow a realIndexChange entirely.

I've landed a fix based on your finding in 9b78f71a: the suppression is scoped to loopFix() itself via an internal __loopFixInProgress__ flag (same pattern as __preventObserver__), so the teleports go silent while every other caller keeps emitting as before, plus a realIndexChange baseline correction so silent teleports can't swallow or duplicate it. It will be included in the next release.

Thanks again for digging into this one - closing in favor of the commit above.

PR #8217, opened 2026-08-26
fix(core): don't crash when calling updateSlides() after destroy()

arturovt:
So here's what was happening: if you call swiper.destroy() and then call swiper.updateSlides() by accident (or some other code in your app does), it would throw an error instead of just doing nothing.

Why? When destroy() runs, it wipes out all the properties on the swiper object (like swiper.slides) to help the browser clean up memory. But updateSlides() (and a few similar functions next to it) never checked if the swiper was already destroyed before trying to read swiper.slides.length. Since swiper.slides was gone, this blew up with "Cannot read properties of undefined (reading 'length')".

The rest of the codebase already has a simple check for this exact situation: "if the swiper is destroyed, just stop and do nothing." This change just adds that same check to updateSlides(), updateSlidesOffset(), updateSlidesProgress(), and updateSlidesClasses(), so they behave the same way as every other method in the library.

Added a test that destroys a swiper and then calls these four functions, making sure they no longer throw.

### DIFF
--- a/src/core/update/updateSlides.ts
+++ b/src/core/update/updateSlides.ts
@@ -8,6 +8,7 @@ import type { Swiper, SwiperSlideElement } from '../core';
 
 export default function updateSlides(this: Swiper): void {
   const swiper = this;
+  if (!swiper || swiper.destroyed) return;
 
   function getDirectionPropertyValue(node: CSSStyleDeclaration, label: string): number {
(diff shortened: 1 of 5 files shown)


nolimits4web:
Thanks, but calling internal update methods after destroy() isn't a supported flow - the public update() already guards, and destroy() clearing state is what surfaces these lifecycle bugs in apps. This also only covers 4 of the 10 update/* methods, so it wouldn't make things consistent.

PR #8218, opened 2026-08-26
feat: add swiperDevMode flag to strip loop-warning logs from prod builds

arturovt:
Right now, Swiper prints console.warn() messages when loop mode is set up in a way that might not work well (like not having enough slides). These warnings always run, even in your production build. That's annoying because error-tracking tools like Sentry or Rollbar often grab console.warn calls and turn them into noisy issues you have to dismiss.

This adds a new global flag called swiperDevMode, similar to how Angular has ngDevMode. By default it does nothing different - the warnings still show up exactly like they always have. But if you want, your bundler (webpack, esbuild, Rollup, or Terser) can be told to replace swiperDevMode with the literal value false at build time. Once that happens, your minifier can see the warning code will never run and deletes it completely, so your production bundle gets smaller and the console stays quiet.

Also added a small test that actually runs an esbuild build with the flag set to false and checks the warning text is really gone from the output, not just skipped at runtime.

No breaking changes here. This is fully opt-in - if you don't touch anything, Swiper works exactly the same as before. It's just a new tree-shaking option for people who want smaller, quieter prod bundles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,59 @@ $ npm run build:prod
 
 Production version will available in `dist/` folder.
 
+### Stripping Dev-Only Warnings
+
+Swiper emits a handful of `console.warn` diagnostics for loop mode misconfiguration
+(e.g. not enough slides, `slidesPerGroup` mismatch). These are gated behind a global
+`swiperDevMode` flag, following the same convention other frameworks use to let
+minifiers dead-code-eliminate dev-only code — Angular's `ngDevMode`, Vue's `__DEV__`,
+and similar:
+
+```js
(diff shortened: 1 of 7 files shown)


nolimits4web:
Thanks, but this isn't something anyone has asked for - these warnings only fire on actual loop misconfiguration, and the right fix is to correct the config. Adding a bare global flag as build-time API is a convention Swiper doesn't use, and it'd only cover 4 of the warnings anyway. If you need to suppress them in prod, filtering console.warn in your error tracker's beforeSend is the standard approach.

PR #8219, opened 2026-08-26
docs: document swiper/element and swiper/element/bundle usage

arturovt:
The README never mentioned the Web Component (custom elements) entry points at all, even though swiper/element/bundle is a real, published import. Added a short section explaining what register() does, the difference between swiper/element (bring your own modules) and swiper/element/bundle (everything pre-registered), and that register() is safe to import/call during SSR - it just no-ops on the server since customElements only exists in the browser, so it still needs to run from client-only code once the page hydrates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,63 @@ $ npm run build:prod
 
 Production version will available in `dist/` folder.
 
+### Web Components (`swiper/element/bundle`)
+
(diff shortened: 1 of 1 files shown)


nolimits4web:
Thanks - the README intentionally doesn't document usage/entry points; that lives on swiperjs.com (see /element and /get-started). Framework integration examples are also out of scope here.

PR #8220, opened 2026-08-27
chore: migrate package management from npm to pnpm

arturovt:
Swapped package-lock.json for pnpm-lock.yaml (generated via `pnpm import` so the resolved versions carry over unchanged), and pointed every internal `npm run` / `npx` call in package.json and scripts/release.js at pnpm instead.

Added a pnpm-workspace.yaml with onlyBuiltDependencies: [esbuild] - pnpm 10 blocks postinstall/build scripts by default for supply-chain safety, and esbuild (pulled in transitively by vite) needs its postinstall to fetch the real platform binary. Without this it silently falls back to a stub and anything using esbuild (vite dev servers, playground/demos scripts) would break.

Also added @vue/server-renderer as an explicit devDependency. It's only ever used transitively through vue, but tests/ssr/ssr.test.mjs imports it directly - npm's flat node_modules hid that gap by hoisting it, pnpm's strict node_modules does not, so the SSR test failed to resolve it until declared directly. This was a real phantom-dependency bug the migration surfaced, not something new the migration introduced.

Updated README's install/build instructions to pnpm. No packageManager field was added, so any pnpm version can be used to work on the repo.

Verified with a clean `pnpm install` + `pnpm run test` (validate, prod build, contract/dist-types/consumer-types/ssr/element-dom/loop-warning/ cross-realm-params tests, bundle-size) - all green.


nolimits4web:
Thanks, but Swiper is intentionally package-manager-agnostic and there's no workspace setup here that would benefit from pnpm; hardcoding pnpm into scripts/release breaks npm/yarn users. Happy to take the @vue/server-renderer devDependency fix as a separate one-line PR.

PR #8223, opened 2026-08-28
fix(core): remove lazyPreload load/error listeners so they can't leak

arturovt:
init()'s lazyPreload block was attaching a `load` listener straight on each incomplete `[loading="lazy"]` image, with no `{ once: true }` and no removal anywhere. The listener closure holds `swiper`, so:

- an image that never fires `load` (broken src / network failure fires `error` instead) keeps the listener - and swiper - attached forever
- even a normal image keeps swiper reachable through the DOM node for as long as that node exists, which can outlive the instance when a framework wrapper retains/reuses the slide DOM across re-renders

Fixed by listening for both `load` and `error` and removing both the moment either fires, instead of leaving whichever one didn't happen still attached. Every image shares a single onLazyImageSettled handler (defined once per Swiper instance, using e.currentTarget rather than closing over a specific image) instead of a new closure per image, and pending images are tracked in a Set - just enough to let destroy() call removeEventListener with the exact same handler reference for anything still waiting on an image that never settled before destroy() ran.

This direct per-image listener isn't fully redundant with the delegated capture-phase `swiper.onLoad` listener (attached on `el`, already cleaned up via detachEvents()): for swiper-element instances it also covers lazy images living in hostEl outside of el, which the delegated listener never sees. So the fix keeps this block rather than removing it in favor of the delegated one.

### DIFF
--- a/src/core/core.ts
+++ b/src/core/core.ts
@@ -639,6 +639,16 @@ export interface Swiper {
   onScroll: () => void;
   onLoad: (event: Event) => void;
 
+  /**
+   * !INTERNAL
+   * Images with a pending `load`/`error` listener registered by init()'s lazyPreload block,
+   * so destroy() can remove them even if the image never settles before then. Both events
+   * share the single onLazyImageSettled handler below, so a Set of images is enough - no
+   * need to store a handler per entry.
+   */
+  lazyPreloadListeners: Set<HTMLImageElement>;
+  onLazyImageSettled: (event: Event) => void;
+
   // Module-injected methods are contributed via `declare module
   // '../../core/core'` from each module's own .ts file.
 }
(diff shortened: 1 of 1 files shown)


nolimits4web:
Thanks for digging into this — the unremoved per-image listener was a real (if minor) leak, but the fix is heavier than it needs to be.

That per-image listener was redundant: `attachEvents()` already delegates `load` on `el` in the capture phase and `detachEvents()` removes it on destroy. It also covers the swiper-element case — a capture listener inside the shadow root does receive `load` events from slotted light-DOM images (verified in Chrome), so there's no need to keep a separate per-image path for `hostEl`.

So the fix is to drop the per-image listener rather than track and clean it up: landed in eb02f3ce, together with handling `error` in the same delegated listener (preloader cleared, no layout update). Closing in favor of that.

